Issue #20, opened 2026-06-25
Epic: Evolve langres from POC → usable, composable ER framework (M0–M6)


fxd24:
# Epic: Evolve langres from POC → usable, composable ER framework

**Goal:** take langres out of POC state into a framework that can **bootstrap labels, compose & benchmark ER methods, distill a cheap judge, and ship a versioned model artifact** — proven on the brainsquad use case (Person / Program / Geography), general by design.

**Direction doc:** [`docs/ROADMAP.md`](https://github.com/raisesquad/langres/blob/docs/roadmap-vision/docs/ROADMAP.md) (PR #19)
**Consumer / companion umbrella:** raisesquad/brainsquad#1051 (langres = the ER *brain*; brainsquad = the persistent *body*)

## The spine (see ROADMAP §2)

Entity = typed **feature bag** → **missing-aware comparison** (asymmetric sparse↔rich matching) → swappable **judge** (heuristic / embedding / LLM / DSPy-distilled / GLinker) → **clusterer** → serializable **Resolver** artifact. Labels gate everything, so the cold-start **bootstrapper** sits on the critical path.

## Critical-path insight

`Blocking (exists) → cold-start gold set → { benchmark · DSPy-distill judge · optimize } → Resolver artifact → brainsquad inference`

## Status & direction (updated 2026-07-02)

**Shipped:** M0 · M0.5 · M1 · M2 · M3 · **M4** (merged to `main`, PR #59). The seam works end-to-end and, on hard data, a **zero-shot cheap judge with a precision-tuned signature beats the frontier model** (0.757 vs 0.667 pair-F1, lower cost) — *without* any distillation. So the "does this even work?" question is answered: **yes.**

**Emerging direction (proposed — to be locked with David):** we are **not** in a "which method is best" phase — langres is the **seam** that lets *many* methods be composed. The next phase is **making it a usable ER tool**: a user hands langres two entities (or a list) and asks *"same? duplicate? new?"*, and langres makes it **easy to stand up and tune an ER architecture for their specific use case — including with _no labeled data_** (zero-shot inference as a first-class, default path). This pulls the KISS/DX work (three-verb **Retrieve → Judge → Resolve** core; collapse speculative `tasks.*`/`flows.*` into `Resolver` configs — see #55) *ahead* of the SOTA-method-breadth (M4.5 S1/S2), which stays **earned-by-need**, not speculative. Milestone detail lives in the living **`docs/ROADMAP.md`**; this epic tracks high-level progress.

[...]

PR #29, opened 2026-06-25
M0: Serializable Resolver + missing-aware Comparator (contract & spine)

fxd24:
## Milestone M0 — Contract & Spine

The foundation milestone from the [ROADMAP](docs/ROADMAP.md) / epic #20: a **serializable `Resolver`** (the `nn.Module` analogue for ER) plus the **entity feature-bag + missing-aware `Comparator`** that makes asymmetric sparse↔rich matching work. langres = the ER *training* brain; the integration contract with brainsquad (#1051) is a versioned, inspectable model artifact.

### Exit criterion — met ✅
An existing example runs through `Resolver.save`/`load` → **identical clusters**, and the external-adapter stub compiles.
- Round-trip identity holds (canonicalized both sides); BCubed **F1 = 1.0** on the company fixture.
- Artifact is human-readable + version-stamped (`resolver.json`: `artifact_version`, `langres_version`, per-slot `type_name`+`config`).
- **Fresh-process** subprocess reload reproduces identical clusters (catches registry/import side-effects).
- `GLinkerAdapter` satisfies both `Blocker` and `Module` ABCs under mypy strict, registers, and round-trips.

### The load-bearing idea
Every Blocker/Module was parameterized by Python **callables** (`schema_factory`, `text_field_extractor`, `field_extractors={"name": (lambda x: x.name, .7)}`). Lambdas don't serialize — so serialization and the feature-bag were **one task**: replace callables with declarative, data-only `FeatureSpec`s and the artifact becomes config-serializable (no pickle).

### What's in it
- **`feature.py`** — `FeatureSpec`, `ComparisonVector`, `combine_present` (the missing-aware weighted combine with the **evidence-floor** over-merge guard: a pair scores 0 unless ≥2 present features OR present-weight ≥ 0.5; missing→neutral, never a missing-vs-missing string compare).
- **`comparator.py`** — `StringComparator` + `Comparator.from_schema(Schema)` (auto-derives specs from Pydantic fields, excludes `id`, str-only).
- **`judges/weighted_average.py`** — `WeightedAverageJudge(Module)` (judge IS a Module — no split-brain; drops into the scoring slot alongside any other Module).
- **`resolver.py`** — composes blocker → optional comparator → module → clusterer; `resolve`/`predict`/`fit`(no-op)/`save`/`load`/`from_schema`; transparent index lifecycle (no manual `create_index`).
- **`registry.py` / `serialization.py`** — config-registry (`@register` + `from_config` + typed errors), `ComponentSpec`/`ArtifactManifest`, optional `SerializableState` for heavy assets.
- **FAISS/embedder `save_state`/`load_state`** (corpus reused on load, no re-embed; embedder persisted by `model_name`).
- **`adapters/glinker.py`** — contract-conformance stub (M3 benchmarks GLiNER behind langres interfaces).

### DX — the whole point
See **[docs/DX_RESOLVER.md](docs/DX_RESOLVER.md)** for before/after. The north-star path:
```python
resolver = Resolver.from_schema(CompanySchema, threshold=0.7)   # id-exclusion baked in
clusters = resolver.resolve(records)
resolver.save("artifacts/company_v0")
assert Resolver.load("artifacts/company_v0").resolve(records) == clusters
```
Runnable: `examples/resolver_company_dedup.py`.

### How it was built
Plan → gStack `/autoplan` (CEO+Eng+DX dual-voice review; ratified gate decisions) → 3 build waves, each a focused agent in an isolated worktree (Wave 1 froze contracts; Wave 2a–2d built leaves in parallel; Wave 3 integrated) → an adversarial pre-PR review whose findings are fixed in the last commit.

### Verification
- `ruff` ✓, `mypy src` strict ✓ (44 files), **842 passed** / 96.6% coverage.
- CI-mirror (`pytest -m "not integration"`, exactly what CI runs, slow tests included): **883 passed**, 98.1% coverage.
- Example executed for real; `resolver.json` inspected; fresh-process reload confirmed.

### Review verdict
Adversarial review → "conditional approve"; the one blocker was a **false positive** (Pydantic-model configs already serialize via `model_dump()`), but the warranted findings are fixed in `e720dc4`: GLinkerAdapter `type_name`, property-vs-method detection on the class, double-`load_state` guard, malformed-manifest error, + a new adapter round-trip regression test.

### Known follow-ups (out of M0 scope)
- `tests/test_integration_approach3.py` (3 tests) fail when run with `-m slow` because they call `VectorBlocker.stream()` without `create_index()` — **pre-existing** (test file unchanged, index-guard logic untouched) and `@pytest.mark.integration` so **excluded from CI**. The Resolver's transparent index lifecycle actually fixes this trap; worth updating those tests separately.
- `Clusterer.cluster` docstring could be more precise ("never adds isolated nodes" vs "drops singletons") — behavior is correct.
- Deferred by design: rich `ComparisonLevel` taxonomy (`mismatch` reserved), `is_anchor` logic, `fit` optimization, pickle escape hatch — all M1+/M3.

https://claude.ai/code/session_01KUj19Gwj4x3pvXywysezPp


fxd24:
### Update since open (review + deeper verification)

Two rounds of hardening after the initial push:

**1. Adversarial review fixes (`e720dc4`).** A pre-PR review flagged one "blocker" (false positive — Pydantic configs already serialize via `model_dump()`) and several warranted findings, all fixed: GLinkerAdapter `type_name` ClassVar; property-vs-method config detection on the *class* (not the resolved instance value); a `not accepts_state_dir` guard against double `load_state`; a malformed-manifest `ValueError` instead of bare `StopIteration`; + a new adapter round-trip regression test.

**2. Fuller verification surfaced a 4th failing test, now resolved (`ad82eb7`).** Running the *complete* suite (slow + integration) on the final commit — not just the fast lane — exposed `test_progressive_pipeline_building_runs`, which I'd missed. Investigation: the `progressive_pipeline_building.py` example was **broken before M0** (stale `langres.core.vector_index` import — that flat path never existed on `main`; the module is `langres.core.indexes.vector_index` — plus a no-embedder `FAISSIndex` call). Per the M0 plan's intent to replace it (superseded by `examples/resolver_company_dedup.py`), the example + its integration test were removed.

**Final verification state (commit `ad82eb7`):**
- ruff ✓, mypy strict ✓ (44 files).
- **CI-mirror** (`pytest -m "not integration"`, exactly CI's command, slow tests included): **884 passed**, 98.1% coverage.
- **Full suite** (slow + integration): **889 passed, 3 failed** — the 3 failures are all `tests/test_integration_approach3.py`, all `@pytest.mark.integration` (excluded from CI), pre-existing (test file untouched by M0, index-guard logic unchanged). They hit the same `create_index` trap the Resolver's transparent index lifecycle fixes — a worthwhile separate cleanup.
- Example run end-to-end by hand: 5 correct clusters, identical reload, **F1 = 1.0**; `resolver.json` inspected directly (version-stamped, full per-slot config).

Net: **CI will be green once the GitHub Actions billing block is cleared** (currently no jobs start — account-level, not code).

fxd24:
## Review findings addressed (commit `9c3620f`)

Thanks both — all warranted findings fixed, TDD with a regression test each. Verified on the exact CI selector (`pytest -m "not integration"`, slow FAISS path included): **892 passed, 98.15% cov, mypy --strict clean**, and the example still round-trips at F1 = 1.0.

| Finding | Resolution |
|---|---|
| **[MAJOR — both] `load()` slot identification fragile** (positional + hard-coded `type_name == "comparator"`) | `ComponentSpec` now carries a `slot`; `save()` records it, `load()` maps by slot — a registered Comparator subclass with a custom `type_name` now loads into the right slot. Legacy/no-`slot` manifests fall back to the old positional logic; a malformed manifest raises a clear error. Locked by `test_resolver_round_trips_comparator_subclass_with_custom_type_name`. |
| **[MAJOR — Claude] `WeightedAverageJudge` secretly required `StringComparator.feature_specs`** | `feature_specs` is now a concrete property on the `Comparator` ABC (default `[]`); `StringComparator` backs it with `_feature_specs`. The judge's `# type: ignore` is gone. Locked by `test_feature_specs_defaults_to_empty`. |
| **[P2 — Codex resolver.py:296] stale vector index on reuse** | `_ensure_index_built` rebuilds when the current corpus differs from the index's stored `_corpus_texts` (reuses on exact match, so a restored index + same records still skips re-embedding). Locked by `test_resolver_rebuilds_vector_index_on_new_corpus`. |
| **[P2 — Codex resolver.py:359] empty sidecar for unbuilt index** | `save()` removes an empty sidecar dir; `load()` only restores from a non-empty one (via a `_has_state` TypeGuard). Locked by `test_resolver_save_unbuilt_vector_index_writes_no_sidecar`. |
| **[MINOR — Claude] `_ensure_index_built` triple `type: ignore`** | Replaced with an `isinstance(self.blocker, VectorBlocker)` guard — all three suppressions removed. |
| **[MINOR — Claude] `GLinkerAdapter.config` instance attr** | Now `_config` + a `config -> dict` property, matching the other components. |
| **[NIT — Claude] 3.12 dropped from CI matrix** | Intentional (3.13-only by decision); `requires-python` left unchanged on purpose. |

**Confirmed non-issues** (you both noted): the `not left_val` "0-is-missing" path (string-only fields) and the GLinker config round-trip (works via the dual-convention bridge).

### Two follow-ups I'm tracking separately (own design review, not blocking M0)
1. **Unify the config convention** — components disagree on whether `config` returns a plain `dict` vs a Pydantic model, bridged by `Resolver._component_config_dict`. It works, but "two ways + a bridge" sits right on M0's serialization seam; worth unifying on one convention post-merge.
2. **Move `_component_spec`/`_rebuild_component`/`_component_config_dict`/`_state_owner` from `resolver.py` into `serialization.py`** — SRP (Resolver should orchestrate, not own marshalling) and it trims a 460-line file.

These will land as a small consistency PR after M0.

PR #31, opened 2026-06-26
M0.5: Seam consolidation — unified judge contract, vector-similarity fix, serializable LLMJudge, Person strong-path proof

fxd24:
## M0.5 — Seam consolidation (fundamentals before M1)

A focused **consolidation pass** that fixes the contract inconsistencies M1 would otherwise multiply, and ships the first **runnable embeddings + LLM strong-path proof**. Surgical, not a rewrite. Scope was deliberately **narrowed to M1-blockers** — the declarative `from_schema(blocker=,judge=)` builder, uniform component-instance API, config-convention unification, and typed feature-bag extension are explicitly **deferred to M2**.

### What changed

**1. Unified `Module`/judge contract.** Every judge now has one signature: `forward(candidates)`. `ERCandidate` carries an optional `comparison: ComparisonVector | None`; the `Resolver`'s comparator stage attaches it, and `WeightedAverageJudge` (which now owns its `feature_specs`/weights) reads it — raising a clear error if it's missing. Removes the forked `forward(..., *, comparator=)` kwarg and the `# type: ignore[call-arg]` bridge in `resolver.py`. Self-contained judges (`LLMJudge`) ignore `comparison` and read raw entities.

**2. Vector similarity correctness (bug fix).** Each `VectorIndex` now owns `to_similarities(distances)` using its known metric (FAISS L2 → `1/(1+√d)`, cosine → `(ip+1)/2`). This kills a real **L2 inversion bug**: the old `VectorBlocker` heuristic clipped raw squared-L2 distances, mapping the *least* similar pair to the *highest* score — which would have corrupted M1's hard-negative mining band. Golden regression test asserts near-dup > non-pair for both metrics.

**3. Serializable `LLMJudge`.** `@register("llm_judge")`, optional/lazy client (`from_env`), `config`/`from_config` that **never serialize the client or any secret**, honest cost via `litellm.completion_cost`, and a domain-neutral prompt (`entity_noun`) shared with `CascadeModule`. A Resolver with an LLM judge now saves/loads.

**4. Person strong-path proof.** New `examples/person_resolution.py` wires `VectorBlocker(FAISS cosine) → LLMJudge → Clusterer` on Person-shaped records, deterministically (fake LLM), and reports blocking Pair-Completeness, honest cost, and save/reload identity. Backed by `tests/examples/test_person_resolution.py` so it can't rot again. The previously-rotted `deduplication_example.py` (and 6 other examples) repaired to current import paths.

**Review fixes (from the gStack review pass):** artifact version bump `0→1` + a clean "older artifact" rejection (was a raw `KeyError` on cross-version load); removed an inverted `judges → llm_judge` dependency (moved `_inspect_scores_impl`/`_generate_recommendations_impl` to `reports.py`); `FakeHybridRerankingVectorIndex` protocol parity (latent `TypeError`); clear `TypeError` for unserializable components in a Resolver slot; honest docstrings on lossy score-clipping; and a mypy override that treats `litellm` as untyped so `# type: ignore` comments stop flipping needed/unused across environments.

### Verification (real numbers, not just green tests)
- **879+ tests pass**, ≥97% coverage; mypy strict clean (deterministic cold+warm cache); ruff clean.
- **rapidfuzz baseline regression:** clusters **byte-identical** to pre-change baseline, BCubed F1 = **1.000**, save/reload identical.
- **Person strong-path:** blocking **Pair-Completeness = 1.000** (target ≥0.95), BCubed/Pairwise F1 = **1.000**, save/reload identical, live OpenRouter smoke gates cleanly on key presence.

### Deferred (tracked, not forgotten)
M2: declarative vector `from_schema` builder, uniform component-instance API, config dict-vs-Pydantic unification, typed feature-bag (aliases/anchor-veto/MISMATCH). Follow-ups noted in review: parsing-helper dedup between LLMJudge/Cascade, `CascadeModule.llm_api_key` retained as a plain attribute (pre-existing secret-exposure smell), compact-JSON prompt tokens.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,3 +2,13 @@
 # Copy to .env (gitignored) before running pytest or pre-commit hooks.
 OMP_NUM_THREADS=1
 KMP_DUPLICATE_LIB_OK=1
+
+# --- LLM teacher / judge (OpenRouter via litellm) -----------------------------
+# OpenRouter is a unified gateway; litellm routes models prefixed "openrouter/".
+# Pull the value from brainsquad's .env (it already has OpenRouter creds):
+#   grep -i openrouter ../brainsquad/.env >> .env
+# Not needed for M0.5 (the example uses a fake/recorded LLM); required for the
+# M1 teacher run and the optional live smoke test.
+OPENROUTER_API_KEY=
+# litellm model id for the judge/teacher, e.g. openrouter/openai/gpt-4o-mini
+LANGRES_LLM_MODEL=openrouter/openai/gpt-4o-mini
--- a/docs/DX_RESOLVER.md
+++ b/docs/DX_RESOLVER.md
@@ -119,10 +119,11 @@ The four-slot constructor is unchanged for advanced use — swap in a
 scorer:
 
 ```python
+comparator = Comparator.from_schema(CompanySchema, weights={...})
 Resolver(
     blocker=AllPairsBlocker(schema=CompanySchema),
-    comparator=Comparator.from_schema(CompanySchema, weights={...}),
-    module=WeightedAverageJudge(),
+    comparator=comparator,
+    module=WeightedAverageJudge(feature_specs=comparator.feature_specs),
     clusterer=Clusterer(threshold=0.7),
 )
 ```
--- a/examples/blocking_evaluation_comprehensive_comparison.py
+++ b/examples/blocking_evaluation_comprehensive_comparison.py
@@ -53,10 +53,10 @@ from langres.core.embeddings import (
     FastEmbedSparseEmbedder,
     SentenceTransformerEmbedder,
 )
-from langres.core.hybrid_vector_index import QdrantHybridIndex
+from langres.core.indexes.hybrid_vector_index import QdrantHybridIndex
 from langres.core.metrics import evaluate_blocking_with_ranking, pairs_from_clusters
-from langres.core.reranking_vector_index import QdrantHybridRerankingIndex
-from langres.core.vector_index import FAISSIndex
+from langres.core.indexes.reranking_vector_index import QdrantHybridRerankingIndex
+from langres.core.indexes.vector_index import FAISSIndex
 from langres.data import load_labeled_dedup_data
 
 # Configure logging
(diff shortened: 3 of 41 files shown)


fxd24:
## Review response

Thanks both — triaged all findings. Pushed `65c874a`.

### Addressed (warranted)
- **[Codex P1] LLMJudge not registered on the fresh-process load path** — real and important: `core/__init__.py` imported `WeightedAverageJudge` (so it registered) but never `llm_judge`, so `Resolver.load()` of an LLMJudge artifact in a clean process raised `UnknownComponentType`. The existing test masked it by importing the module directly. Fixed by importing `LLMJudge` in `langres.core.__init__` (+ `__all__`), and added a **subprocess fresh-process reload test** that loads via `from langres.core import Resolver` only — fails without the fix, passes with it.
- **[claude-review] `_RateLimiter.acquire` deadlock** — confirmed: the whole body ran inside `async with self._lock`, awaiting `asyncio.sleep` while holding the non-reentrant lock → first rate-limit hit stalls all concurrent coroutines (M1 labels concurrently). Rewritten to compute-wait under lock, then sleep *outside* the critical section and retry. Added a concurrency regression test.
- **[claude-review] unused imports** `defaultdict` + `from openai import OpenAI` — removed.
- **[claude-review] `ERCandidate.comparison` docstring** — expanded to explain the two-phase pipeline (Comparator attaches, comparison-aware judge consumes, self-contained judge ignores).
- **[claude-review] `CascadeModule` ambiguity** — added an honest debt note to its docstring (not serializable; internal `SentenceTransformer`; duplicated parse/cost helpers — tracked for M2).

Earlier in the branch (from an internal review pass before the bots ran): artifact version `0→1` + older-artifact rejection, removed the `judges → llm_judge` dependency inversion, `FakeHybridRerankingVectorIndex` protocol parity, clear `TypeError` for unserializable components, example/docstring import-path rot, and a mypy override that makes `litellm` typing deterministic (stops `# type: ignore` flipping needed/unused across environments).

### Deferred / skipped (with reasons)
- **`gpt-5-mini` "is not a real model"** — *skipping*: it is a current OpenAI model (as of 2026) and the intentional default set back in M0; not flipping a product default on stale model knowledge. Happy to revisit if you want `None`-forces-explicit instead.
- **`BlockerEvaluationReport._gold_clusters` → proper Pydantic `Field`** — *deferred*: pre-existing report code, not part of M0.5's contract; tracked for a reports cleanup.
- **Deprecated `build()` coverage** (`vector_index.py`) — *deferred*: pre-existing deprecated path; remove-or-cover in the M2 cleanup.
- **`CascadeModule` full SRP/injection + helper dedup + `llm_api_key` attribute** — *deferred to M2*: Cascade isn't a serializable slot component this milestone; the debt is now documented in its docstring.
- **`FakeVectorIndex.search_all` nested-loop dedup (TODO)** — *deferred*: pre-existing, cosmetic.

Issue #6, opened 2026-02-20
Address code review findings from PR #5


fxd24:
## Context

The Claude code reviewer on PR #5 identified several issues worth addressing in follow-up work. These are tracked here for incremental fixes after the core foundation branch is merged.

## Bugs

- [ ] **BCubed precision inflated for unknown entities** (`src/langres/core/metrics.py:56-57`) — When entities are absent from `gold_clusters`, `gold_lookup.get(entity)` returns `None`, and `None == None` evaluates to `True`, falsely inflating precision. Needs a guard like `if gold_lookup.get(entity) is not None`.

- [ ] **`_RateLimiter.acquire()` holds the lock while sleeping** (`src/langres/core/modules/llm_judge.py:81-113`) — The `async with self._lock` is held across `await asyncio.sleep()`, serializing all concurrent requests and negating the async speedup. The sleep should happen outside the lock.

## Design Issues

- [ ] **`_index_is_built()` inspects private attributes of concrete types** (`src/langres/core/blockers/vector.py:187-205`) — Breaks dependency injection by checking `._index`, `._corpus_texts`, `._n_samples` on concrete implementations. Fix: add an `is_built: bool` property to the `VectorIndex` protocol.

- [ ] **Distance-to-similarity heuristic is unreliable** (`src/langres/core/blockers/vector.py:316-366`) — Metric type inference by checking monotonicity and `<= 2.0` is fragile. The `VectorIndex` protocol should expose the metric type, or similarity conversion should happen inside the index implementation.

- [ ] **`_format_entity` assumes Pydantic models** (`src/langres/core/modules/llm_judge.py:489`) — Calls `entity.model_dump_json()` but `SchemaT` is unbounded. Non-Pydantic schemas will get `AttributeError`. Needs a type bound or fallback.

- [ ] **Default temperature 1.0 is wrong for entity resolution** (`src/langres/core/modules/llm_judge.py:177`) — Temperature 1.0 introduces non-determinism. Should default to `0.0` for reproducibility.

- [ ] **`Clusterer.cluster()` silently drops unmatched entities** (`src/langres/core/clusterer.py:62-72`) — Entities only in sub-threshold judgements are absent from output instead of appearing as singletons. This is surprising behavior that should be documented or fixed.

## Minor

- [ ] **`create_llm_client()` returns `Any`** (`src/langres/clients/llm.py:15`) — Bypasses type checking. A typed protocol would be better.

[...]

Issue #3, opened 2025-11-15
Blocker evaluation requires similarity scores - not compatible with all blocker types


fxd24:
## Problem

The new blocker evaluation API (`Blocker.evaluate()` and `evaluate_blocker_detailed()`) requires all candidates to have `similarity_score` populated to compute ranking metrics (MAP, MRR, NDCG, recall@k curves, rank distributions).

This creates a **compatibility limitation**:
- ✅ **Works with**: `VectorBlocker` (automatically populates similarity scores based on embedding distance)
- ❌ **Fails with**: `AllPairsBlocker` (doesn't compute similarity scores)
- ❌ **Fails with**: Custom blockers that don't compute similarity (e.g., rule-based, blocking key approaches)

## Current Behavior

When calling `evaluate()` on a blocker that doesn't populate similarity scores:

```python
# AllPairsBlocker doesn't populate similarity_score
blocker = AllPairsBlocker(schema_factory=company_factory)
candidates = list(blocker.stream(entities))

# This raises ValueError
report = blocker.evaluate(candidates, gold_clusters)
```

**Error:**
```
ValueError: evaluate_blocking_with_ranking requires similarity_score to be populated 
in all candidates. VectorBlocker.stream() automatically populates this field.
```

See: `src/langres/core/metrics.py:418-424`

## Why This Matters

1. **Current main use case is VectorBlocker** - This is the primary production blocker and works fine
2. **But it's a design limitation** - The evaluation API assumes similarity-based ranking, which not all blockers provide
3. **Examples broke** - `blocker_evaluation_comprehensive.py` originally used `AllPairsBlocker` and failed (now fixed by switching to `VectorBlocker`)

### Option 1: Split evaluation methods (recommended for V1)
```python
# Basic metrics - works with any blocker
basic_report = blocker.evaluate_candidates(candidates, gold_clusters)
# Contains: recall, precision, total_pairs, missed_matches, etc.

# Full metrics - requires similarity_score
full_report = blocker.evaluate_full(candidates, gold_clusters, k_values=[...])
# Contains: everything + MAP, MRR, NDCG, rank distributions, recall curves
```

### Option 2: Graceful degradation
Make ranking metrics optional - skip them when `similarity_score` is None:
```python
report = blocker.evaluate(candidates, gold_clusters)
# report.candidates.*  -> always populated
# report.ranking.*     -> None if no similarity scores
# report.scores.*      -> None if no similarity scores
# report.ranks.*       -> None if no similarity scores
[...]

Issue #2, opened 2025-11-13
Remove variant_ids/variant_names backward compatibility from data loader


fxd24:
## Problem

Backward compatibility for `variant_ids`/`variant_names` was supposed to be removed during code cleanup (confirmed by user), but it has crept back into the codebase because example data files still use the old format.

## Current State

**Backward compatibility code in `src/langres/data/loaders.py` (lines 109-113):**
```python
# Support both "entity_ids"/"entity_names" and "variant_ids"/"variant_names"
raw_ids = raw_group.get("entity_ids") or raw_group.get("variant_ids", [])
group_entity_ids = [str(id_val) for id_val in raw_ids]
group_entity_names = raw_group.get("entity_names") or raw_group.get("variant_names", [])
```

**Data file using old format:**
- `examples/data/funder_name_deduplicated_groups.json` uses `variant_ids`/`variant_names`

**Affected example scripts:**
- `examples/blocking_evaluation_faiss_vs_qdrant.py`
- `examples/blocking_evaluation_with_reranking.py`

## What Needs to Be Done

1. **Update data file to new format:**
   - File: `examples/data/funder_name_deduplicated_groups.json`
   - Change: Replace `variant_ids` → `entity_ids`, `variant_names` → `entity_names`

2. **Remove backward compatibility code:**
   - File: `src/langres/data/loaders.py`
   - Remove lines 109-113 (the fallback logic)
   - Simplify to direct field access: `raw_group["entity_ids"]` and `raw_group["entity_names"]`

3. **Verify examples still work:**
   - Test both affected example scripts
   - Run `uv run pytest tests/data/test_loaders.py`

## Impact

- ✅ No tests affected (no test files use old format)
- ✅ Only 2 example scripts affected
- ✅ Cleaner, simpler loader code (-5 lines)
- ✅ Enforces standard schema

## Context

This cleanup was part of the aggressive data refactoring to make the library clean and lightweight. The standard schema uses `entity_ids`/`entity_names`, and we should not maintain backward compatibility at this early POC stage.

PR #34, opened 2026-06-28
M1 Wave 2: Miner + Labeler interfaces, HardNegativeMiner, budget-capped TeacherLabeler

fxd24:
## Wave 2 — mining interface + labelers (M1 cold-start gold-set bootstrapping)

Builds on Wave 1 (`langres.bootstrap.models`, `langres.data.er_benchmarks`). Adds the mining/labeling layer that turns raw blocker candidates into a labeled `GoldSet`.

### What's here
- **`bootstrap/base.py`** — `Miner` and `Labeler` ABCs. One honest signature each; not `@register` Resolver slots (bootstrapping is offline data-prep, W2).
- **`bootstrap/miners.py`** — `HardNegativeMiner`: three-stratum sampling over `similarity_score` (high ≥85th pct / mid 40–85th / low <40th), proportions weighted toward mid; deterministic seeded RNG; `max_pairs` cap preserves proportions with shortfall redistribution (W4).
- **`bootstrap/labelers.py`** — `GroundTruthLabeler` (zero-spend, deterministic) + `TeacherLabeler` (budget-capped LLM teacher) + `BlindCostError`.
- **`bootstrap/__init__.py`** — exports all of the above alongside `GoldPair`/`GoldSet`.
- **`tests/bootstrap/`** — `test_miners.py`, `test_labelers.py` (fake judge, no network/key).

### How the design-review items are satisfied
- **B1 (cap mechanism).** `TeacherLabeler` uses only the **synchronous** `LLMJudge.forward` (never `forward_async` — a dispatched `gather` can't be capped). Three layers bound spend: (1) **pre-flight hard cap** truncates input to `floor(budget_soft_usd / worst_case_per_pair_cost)`; (2) **running tally** from actual token counts × pinned prices, cross-checked against the judge's `cost_usd` (max of the two), with a **budget stop** before each batch that returns the partial result; (3) **per-call resilience** — each pair judged in its own `forward` call wrapped in try/except, so one failure skips that pair and the loop continues. If a judgement reports neither tokens nor cost, spend can't be tracked → `BlindCostError` abort.
- **B4 (no-Langfuse client).** `TeacherLabeler.from_env` builds the judge via `create_llm_client(Settings(), enable_langfuse=False)`, so the teacher never depends on Langfuse creds (the usual `LLMJudge.from_env` enables Langfuse by default and raises without `LANGFUSE_*`). Fully constructible/testable with no key.
- **W1.** No speculative `Sampler` ABC.
- **W4.** Three-stratum sampling (not just a mid band); cross-source filtering is the Wave-4 orchestrator's job *before* `mine` (documented in the miner docstring).

### Gates
- `uv run mypy src/` — clean (strict, 50 files).
- `uv run ruff check` + `format --check` — clean.
- `uv run pytest -m "not integration"` — **1000 passed**, overall coverage **98.76%**, **100%** on all new bootstrap files (branch coverage included).

### Deliberate deferral
No `docs/` changes: the M1 bootstrap public-contract docs land when M1 integrates (later waves touch the same surface); adding partial docs now would conflict.

### DIFF
--- a/src/langres/bootstrap/__init__.py
+++ b/src/langres/bootstrap/__init__.py
@@ -1,8 +1,27 @@
 """Cold-start gold-set bootstrapping (M1).
 
-Public data contract for labeled record pairs used to seed entity resolution.
+Public surface for turning raw blocker candidates into a labeled
+:class:`GoldSet`:
+
+- Data contract: :class:`GoldPair`, :class:`GoldSet`.
+- Interfaces: :class:`Miner`, :class:`Labeler`.
+- Miners: :class:`HardNegativeMiner` (stratified sampling).
+- Labelers: :class:`GroundTruthLabeler` (zero-spend, deterministic),
+  :class:`TeacherLabeler` (budget-capped LLM teacher).
 """
 
+from langres.bootstrap.base import Labeler, Miner
+from langres.bootstrap.labelers import BlindCostError, GroundTruthLabeler, TeacherLabeler
+from langres.bootstrap.miners import HardNegativeMiner
 from langres.bootstrap.models import GoldPair, GoldSet
 
-__all__ = ["GoldPair", "GoldSet"]
+__all__ = [
+    "BlindCostError",
+    "GoldPair",
+    "GoldSet",
+    "GroundTruthLabeler",
+    "HardNegativeMiner",
+    "Labeler",
+    "Miner",
+    "TeacherLabeler",
+]
--- a/src/langres/bootstrap/_pairs.py
+++ b/src/langres/bootstrap/_pairs.py
@@ -1,0 +1,23 @@
+"""Shared record-pair identity helper for bootstrapping.
+
+Both the miner (deduplication) and the ground-truth labeler need an
+order-independent key for a record pair so that ``(a, b)`` and ``(b, a)`` are
+treated as the same pair. Keeping that logic in one place prevents the two
+call sites from drifting apart.
+"""
+
+
+def canonical_pair_key(left_id: str, right_id: str) -> tuple[str, str]:
+    """Return an order-independent key for a record pair.
+
+    ``(a, b)`` and ``(b, a)`` both map to the same tuple, so the key identifies
+    a pair regardless of which side each record was placed on.
+
+    Args:
+        left_id: One record id.
+        right_id: The other record id.
+
+    Returns:
+        The two ids as a sorted 2-tuple.
+    """
+    return (left_id, right_id) if left_id <= right_id else (right_id, left_id)
--- a/src/langres/bootstrap/base.py
+++ b/src/langres/bootstrap/base.py
@@ -1,0 +1,68 @@
+"""Abstract interfaces for cold-start gold-set bootstrapping (M1).
+
+Two tools cooperate to turn raw blocker candidates into a labeled
+:class:`~langres.bootstrap.models.GoldSet`:
+
+- A :class:`Miner` selects *which* candidate pairs are worth labeling (e.g.
+  hard-negative / stratified sampling) — it neither labels nor spends.
+- A :class:`Labeler` assigns a match / non-match label to candidate pairs,
+  emitting one :class:`~langres.bootstrap.models.GoldPair` per candidate. The
+  label may come from a teacher LLM, benchmark ground truth, or a human.
+
+These are plain ABCs, NOT Resolver slot components: they are not registered via
+``@register`` and carry no serializable ``config`` — bootstrapping is an offline
+data-preparation step, not a slot in the resolve pipeline.
+
+Each ABC declares exactly one honest signature. Subclasses must implement it
+without widening (Liskov): a caller holding a ``Miner``/``Labeler`` reference
+must be able to rely on the contract here regardless of the concrete class.
+"""
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+from langres.bootstrap.models import GoldPair
+from langres.core.models import ERCandidate
+
+
+class Miner(ABC):
+    """Selects candidate pairs worth labeling from a larger candidate pool."""
+
+    @abstractmethod
+    def mine(
+        self,
+        candidates: list[ERCandidate[Any]],
+        *,
+        max_pairs: int | None = None,
+    ) -> list[ERCandidate[Any]]:
+        """Return the subset of ``candidates`` worth labeling.
+
+        Args:
+            candidates: The candidate pairs to mine from (e.g. blocker output).
+            max_pairs: Optional cap on the number of returned pairs. ``None``
+                means "no cap" — return every selected pair.
+
+        Returns:
+            A deduplicated list of selected candidate pairs.
+        """
+        raise NotImplementedError
+
+
+class Labeler(ABC):
+    """Assigns match / non-match labels to candidate pairs."""
+
+    @abstractmethod
+    def label(self, candidates: list[ERCandidate[Any]]) -> list[GoldPair]:
+        """Label each candidate pair, returning one :class:`GoldPair` per input.
+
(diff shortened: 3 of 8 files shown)


fxd24:
Thanks both — addressed in `2482d71`.

**Fixed (warranted):**
- **[P1 / MAJOR] Intra-batch budget overshoot** (codex + claude): the budget gate is now **per-pair**, not per-batch. A pair is only started if `total_spent + worst_case_per_pair_cost <= budget_usd`, so a large `batch_size` can no longer overshoot when the pinned prices under-estimate actual cost. Residual overshoot is bounded to the single in-flight pair's actual cost beyond the estimate (unavoidable — an LLM's cost is known only post-call); documented in the class docstring. New regression test `test_budget_stop_holds_within_a_large_batch` (batch_size=50) proves the cap holds where the old per-batch check would have spent $50.
- **[CRITICAL] Test inspected `_judge.client` internals**: removed; the test now asserts only public config + that construction succeeds without keys.
- **[MAJOR] Duplicated pair-key logic**: extracted `canonical_pair_key` to `bootstrap/_pairs.py`, used by both miners and labelers.
- **[MAJOR] Dead `raise NotImplementedError`** in abstractmethods: removed.
- **[MAJOR] `label()` per-call reset**: documented (class + method docstrings) and covered by `test_stats_reset_between_calls`.
- **[minor]** Added empty-input tests for both labelers; `test_preflight_cap_truncates_input` now uses `budget_soft_usd < budget_usd`.

**Intentionally skipped (rationale):**
- **`model="gpt-5-mini"` default**: matches the existing `LLMJudge` default (llm_judge.py) — kept for consistency rather than diverging from the established codebase convention.
- **`positive_pairs` tuple-length validation**: the param is typed `set[tuple[str, str]]` and the main constructor `from_clusters` always yields valid 2-tuples — runtime validation here would be speculative error-handling the repo's "simplicity-first" rule discourages.
- **`# type: ignore[arg-type]` in `_stratify`**: stable (not env-flipping) and guarded by the `_dedup` non-None invariant; left as-is.
- **`judge` as a structural `Protocol` (note)**: good idea, but a NOTE rather than required — deferring to keep this change surgical.

Open Q3 (CI without keys): confirmed — `test_from_env_builds_teacher_without_langfuse` passed in the green `test (3.13)` run with no `OPENAI_API_KEY` / `LANGFUSE_*` set (`enable_langfuse=False`).

fxd24:
Round-2 review addressed in `e4ef61d`:

- **BlindCostError discards paid pairs**: went with Option B — `BlindCostError` now carries the already-labeled pairs on `.partial`, set by `label()` before re-raising, so a caller can recover paid work. New assertion in `test_blind_cost_aborts_after_recording_prior_spend`.
- **`batch_size` docstring**: corrected the `__init__` Args to say it's iteration/log chunking only; the gate is per-pair.
- **Redundant sort in `from_clusters`**: switched to `itertools.combinations(sorted(cluster), 2)`.
- **Mutable price attrs nit**: added a comment that the price/token attrs are read-only after `__init__` (the worst-case cost is cached there).

mypy strict / ruff clean; bootstrap files remain at 100% line+branch coverage.

fxd24:
Round-3 addressed in `737fa30`:

- **`_stratify` type-ignores → `cast`**: replaced both `# type: ignore[arg-type]` with `typing.cast(float, ...)`, guarded by the `_dedup` non-None invariant (documented). Resolves the recurring question.
- **`_apply_preflight_cap` copy**: returns the input list directly when no cap fires (no allocation); `_label_one` never mutates it.
- **`BlindCostError.partial`**: added an inline comment that it's populated by the catcher (`label()`), not at raise time.

**Not a bug (kept):** `from_env(model="gpt-5-mini")` — this is the repo's established default, matching `LLMJudge.__init__`'s `model="gpt-5-mini"` and its docstrings/examples. It's a real model in this project's provider config; the flag appears to stem from a pre-gpt-5 knowledge cutoff. Diverging here would make the teacher inconsistent with the judge it wraps.

fxd24:
Round-4 addressed in `a98a9a0`:

- **[MAJOR #1] Stale-cache footgun**: `_worst_case_per_pair_cost` is now a computed `@property` (live from the price/token attrs), so mutating a price after construction can never leave a stale value. Dropped the read-only comment; added `test_worst_case_cost_reflects_live_attrs`.
- **[MAJOR #2] Coverage intent**: added `tests/bootstrap/test_pairs.py` exercising `canonical_pair_key` both branches directly (was 100% only transitively).
- **[MINOR #3] Double canonicalization**: `from_clusters` now adds the `combinations(sorted(...))` pairs directly; `__init__` canonicalizes the set.
- **[MINOR #5] Tie-break**: documented the `-_STRATA.index` bias (favors high→mid→low on exact frac ties; only affects the 1-2 leftover units, and mid is already weighted heaviest).
- **[MINOR #6] `_percentile` boundaries**: added a `pct=0` / `pct=100` test.
- **[N2/N3]** Added clarifying comments to `FakeJudge` empty path and `BlindCostError.partial`.

**Skipped (rationale):**
- **[MINOR #4]** Keeping `test_from_env_builds_teacher_without_langfuse` as a unit test is deliberate: its whole purpose (design-review B4) is to prove `from_env` constructs with no keys and no network in unit CI. `create_llm_client(..., enable_langfuse=False)` makes no network call; marking it `integration` or mocking the factory would stop verifying the exact wiring B4 is about.
- **[N1]** `model="gpt-5-mini"` matches the repo's established `LLMJudge` default (bare name, not `azure/`-prefixed) — kept for consistency with the judge the teacher wraps.

mypy strict / ruff clean; bootstrap files at 100% line+branch (65 tests).

fxd24:
Round-5 addressed in `cdf4332`:

- **[B2] `max_pairs=-1` opaque crash**: `mine()` now raises `ValueError("max_pairs must be non-negative")` before the fast-path; added `test_negative_max_pairs_raises`.

**Skipped (rationale):**
- **[B1] Blind-cost on free calls — false positive for this wiring.** `LLMJudge.forward` *always* writes `prompt_tokens`/`completion_tokens`/`cost_usd`, defaulting them to `0` exactly when `response.usage` is missing (llm_judge.py). So the proposed `prov.get("prompt_tokens") is None` check would **never** be true with the real judge — it would silently disable blind detection. "Both values 0" is the correct signal (and what the design-review spec mandated): a real call always spends prompt tokens, so all-zero reliably means "usage absent." Added a code comment documenting this so it isn't "fixed" later.
- **[B3] `_allocate` precondition**: already documented in the method docstring ("`total` is assumed `<` `sum(sizes)`... so the leftover loop always terminates"). The sole caller guarantees it; an `assert` would be stripped under `-O`, so the docstring is the durable guard.
- **[D1] `gpt-5-mini`**: matches the repo's established `LLMJudge` default; kept for consistency with the wrapped judge (covered in earlier replies).

mypy strict / ruff clean; bootstrap at 100% line+branch (66 tests).

fxd24:
Round-6 (doc-only) addressed in `57d96fe`:
- Pre-flight cap docstring now states it (and the budget stop) are enforced **per `label()` call** — multi-round callers budget across calls (concern 1).
- `worst_case_tokens_per_pair` docstring notes the cap conservatively prices all tokens at the higher rate (concern 2).
- Fixed the misleading no-cap return comment (concern 3) and clarified `from_clusters` shares `__init__` with the direct constructor.

Kept: `gpt-5-mini` default (matches `LLMJudge`); the direct `_worst_case_per_pair_cost` test (it pins the round-4 no-cache fix); `from_env` keyless unit test (confirmed green in CI with no `OPENAI_API_KEY`/`LANGFUSE_*`).

The substantive review items (per-pair budget cap, shared pair-key, recoverable BlindCostError, negative-max_pairs guard) are all resolved and the remaining notes are cosmetic. CI merge gates (lint, test 3.13, claude-review) are green.

PR #35, opened 2026-06-28
M1 Wave 4: Bootstrapper orchestrator + deterministic Fodors-Zagat e2e

fxd24:
## What

The **integration wave** of M1: wires the merged Waves 1-3 pieces into one cold-start orchestrator and proves the loop end-to-end on Fodors-Zagat with **real embeddings, zero spend, full determinism**.

This is composition, not new machinery — it reuses Wave 1's data adapter, Wave 2's Miner/Labeler/HardNegativeMiner/TeacherLabeler, and Wave 3's BootstrapReport unchanged.

## The wiring

`Bootstrapper.build(corpus, *, candidate_filter=None, gold_clusters=None) -> (GoldSet, BootstrapReport)`:

1. Build the blocker's vector index from the corpus (using the blocker's own `schema_factory` + `text_field_extractor`, so the orchestrator is entity-type-agnostic).
2. `candidates = list(blocker.stream(corpus))` — materialize the single-pass iterator (the miner needs the full list for percentile strata).
3. Apply the caller's `candidate_filter` (e.g. cross-source).
4. `miner.mine(filtered)` → `labeler.label(mined)`.
5. Assemble `GoldSet` (honest counts + cost) and a `BootstrapReport` (pair-completeness on the filtered candidates + teacher-vs-truth agreement/calibration).

## Design-review compliance

- **W1** — `Bootstrapper` is a plain class with exactly 3 injected deps (`blocker`, `miner`, `labeler`); describable without "and".
- **W3** — the orchestrator builds the index itself (the private `Resolver._ensure_index_built` is unavailable) and materializes the candidate iterator before mining.
- **B2** — entity-type-agnostic: the Fodors-Zagat cross-source rule (`lambda c: c.left.source != c.right.source`) is passed IN as `candidate_filter`, not baked in.
- **B3** — the e2e example uses real `SentenceTransformerEmbedder("all-MiniLM-L6-v2")` + FAISS at `DEFAULT_BLOCKING_K=5`.

## FakeLabeler (CI stand-in for the teacher)

`GroundTruthLabeler` makes teacher-vs-truth agreement trivially 1.0, so it can't exercise the calibration report. `FakeLabeler` (new, in `labelers.py`, exported) labels by a similarity threshold and emits a deterministic, deliberately miscalibrated `confidence` (over-confident on the ambiguous near-threshold band) so agreement < 1 and Brier/ECE are non-trivial. `source="fake"`, no spend, fully deterministic. `GoldPairSource` is extended with `"fake"` (additive, backward-compatible).

## Measured e2e (deterministic, $0)

| Metric | Value |
|---|---|
| Blocking Pair-Completeness (cross-source) | **0.9911** (gate ≥ 0.95 ✓) |
| Teacher(fake)-vs-truth | acc=0.911, F1=0.915, kappa=MCC=0.821 (n=213) |
| Calibration | Brier=0.106, ECE=0.181 |
| Labels | 127 match / 1254 non-match |
| Cost | $0.0000 |
| GoldSet save→load round-trip | identical ✓ |

## Gated real-GLM branch (Wave 5)

`maybe_real_teacher()` behind `if os.getenv("OPENROUTER_API_KEY")` constructs a budget-capped `TeacherLabeler.from_env(...)` and notes that Wave 5 pins the GLM model/prices and runs it for real. It **only constructs** (never calls `.label()`), so it never spends or hits the network, and CI's pytest never calls `main()`.

## Tests / gates

- `tests/bootstrap/test_bootstrapper.py` (fast, no network): fake blocker/miner/labeler assert build() wires steps in order, builds the index, materializes the iterator, applies/omits the filter, handles gold_clusters present/absent, and records honest counts/cost. **100%** on `bootstrapper.py`.
- `tests/bootstrap/test_labelers.py`: FakeLabeler determinism, threshold correctness, confidence range, `similarity_score is None` guard. **100%** on the new code.
- `tests/examples/test_m1_bootstrap_fodors_zagat.py` (`@pytest.mark.slow`, real embeddings, runs in CI): asserts Pair-Completeness ≥ 0.95, a non-empty GoldSet + non-degenerate agreement/calibration, and the GoldSet round-trips.
- `uv run mypy src/` strict clean; `ruff check`/`format` clean.
- Full combined Waves 1-4 suite: **1078 passed, 0 failed**, overall coverage **98.87%** (gate 95%), **100%** on both new files.

## Integration notes (composing Waves 1-3)

- No source changes needed in Waves 1-3 — they composed cleanly. The only contract touch is the additive `GoldPairSource += "fake"`.
- Observation (not a bug): importing `langres.clients` auto-loads a repo `.env` into `os.environ`, so the example's `os.getenv("OPENROUTER_API_KEY")` gate sees a key even when the shell doesn't. Harmless here because the gated branch only constructs (no spend/call) and tests never invoke it.

### DIFF
--- a/examples/m1_bootstrap_fodors_zagat.py
+++ b/examples/m1_bootstrap_fodors_zagat.py
@@ -1,0 +1,220 @@
+"""M1 cold-start bootstrapping, end-to-end on Fodors-Zagat (deterministic, free).
+
+This is the runnable proof that the M1 Wave 1-4 pieces compose into one pass:
+real semantic blocking (sentence-transformer embeddings + FAISS) -> cross-source
+filter -> stratified hard-negative mining -> labeling -> a labeled
+:class:`~langres.bootstrap.models.GoldSet` plus a
+:class:`~langres.bootstrap.report.BootstrapReport` (blocking pair-completeness,
+teacher-vs-truth agreement, calibration), saved to disk and round-tripped.
+
+The default run is **deterministic and costs $0**: it labels with
+:class:`~langres.bootstrap.labelers.FakeLabeler`, a similarity-threshold stand-in
+for the real teacher whose over-confident confidence keeps the calibration report
+non-trivial. Embeddings are real (so blocking pair-completeness is honest) but
+involve no network and no API key.
+
+A gated real-GLM branch (``maybe_real_teacher``) shows how Wave 5 swaps in a
+budget-capped :class:`~langres.bootstrap.labelers.TeacherLabeler`; it only runs
+when ``OPENROUTER_API_KEY`` is set and never costs anything otherwise.
+
+Run it::
+
+    OMP_NUM_THREADS=1 KMP_DUPLICATE_LIB_OK=1 uv run python examples/m1_bootstrap_fodors_zagat.py
+
+``print`` is allowed in examples (this is demonstration, not library code).
+"""
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+from typing import Any
+
+from langres.bootstrap import (
+    BootstrapReport,
+    Bootstrapper,
+    FakeLabeler,
+    GoldSet,
+    HardNegativeMiner,
+)
+from langres.core.blockers.vector import VectorBlocker
+from langres.core.embeddings import SentenceTransformerEmbedder
+from langres.core.indexes.vector_index import FAISSIndex
+from langres.data.er_benchmarks import (
+    DEFAULT_BLOCKING_K,
+    RestaurantSchema,
+    load_fodors_zagat,
+)
+
+
+def _cross_source(candidate: Any) -> bool:
+    """Keep only candidate pairs whose two records come from different sources.
+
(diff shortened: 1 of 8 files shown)


fxd24:
Thanks — addressed in 3afb38c:

- **Double index build (Medium):** verified `stream()` never calls `create_index` (it raises if the index isn't built), so there's no double build. Added an exact-once assertion (`create_calls == 1`) to document and guard it.
- **FakeLabeler confidence range (Low):** docstring now states the real range `[0.7, 0.99]` (0.7 floor at the threshold).
- **Private `teacher._judge` in example (Low):** removed; the gated branch now prints only the budget.
- **Empty-corpus + all-filtered (Nit):** both added as tests.
- **`_cross_source` typing (Nit):** now `ERCandidate[RestaurantSchema]`.

Skipped: `VectorBlocker[Any]` → Protocol to drop the test `# type: ignore`s — a broader design change better suited to a later wave; noted for follow-up.

PR #46, opened 2026-06-29
M3 W3: 5 methods behind one resolver-factory + EmbeddingScoreJudge

fxd24:
## M3 Wave 3 — competing methods behind one resolver-factory + the missing judge

Wires the **5 competing methods** behind ONE uniform `resolver_factory` interface so the W1 benchmark harness (`run_method`) can race them on any dataset, and builds the one missing judge. **Zero real LLM spend in this wave** — LLM/cascade are tested with a mocked client; the paid run is W4.

### Deliverables

**1. `EmbeddingScoreJudge`** (`src/langres/core/judges/embedding_score.py`) — registry-serializable embedding-cosine scorer. Reads `ERCandidate.similarity_score` (set by `VectorBlocker`), emits a `PairwiseJudgement` whose `score` is that cosine; raises a clear `ValueError` pointing at `VectorBlocker` when `similarity_score is None`. `@register("embedding_score_judge")` + `config`/`from_config`, imported in `core/__init__` + `judges/__init__`, with a fresh-process subprocess reload test. **100% coverage.**

**2. `langres.methods`** — the method registry. `make_resolver_factory(method, benchmark, *, llm_client=...)` returns `Callable[[float], Resolver]`. Blocking is held constant per dataset (`benchmark.build_blocker(benchmark.blocking_k)`), so the race compares **judges, not blockers**:

| method | wiring |
|---|---|
| `rapidfuzz` | VectorBlocker → `RapidfuzzModule` (field extractors derived from the schema's comparable fields, same set as the comparator) → Clusterer |
| `weighted_average` | VectorBlocker → `Comparator.from_schema` → `WeightedAverageJudge` → Clusterer (generalizes `build_restaurant_resolver` to any schema) |
| `embedding_cosine` | VectorBlocker → `EmbeddingScoreJudge` → Clusterer (no Comparator) |
| `llm_judge` | VectorBlocker → `LLMJudge(client=injected)` → Clusterer (mock in tests, real GLM/frontier client in W4; never a live-key requirement at build) |
| `cascade` | VectorBlocker → `CascadeModule` (placeholder key + injected `_llm_client`) → Clusterer |

A dataset opts in by conforming to the new `BlockingBenchmark` protocol (`schema` + `blocking_k` + `build_blocker`). `FodorsZagatBenchmark` gets those three (minimal). `cascade_cost_track` surfaces escalation-rate + llm-calls-per-candidate (the generic `_cost_track` leaves them `None`).

**3. Tests** (deterministic, zero spend): judge unit + reload; each factory builds a valid `Resolver` and runs end-to-end on a tiny synthetic corpus (LLM/cascade with a mock client); `BudgetedModuleRunner` integration (pre-flight cap, `BlindCostError` on $0, per-call isolation, cascade `llm_cost_usd` flowing through the runner's fallback); a fast `run_method` race on a `FakeVectorIndex` benchmark + a slow real-embedding race on `FodorsZagatBenchmark`, each asserting both pair-level and pipeline tracks populate; real-network glue behind an `OPENROUTER_API_KEY` skipif.

### Verification
- Full suite (incl. slow, excl. integration): **1181 passed**, **98.97%** total coverage. **100%** on `methods.py`, `embedding_score.py`, and `er_benchmarks.py`.
- mypy strict clean; ruff clean; no new deps; no public Resolver/Blocker/Module/judge API changes.
- Did not touch `data/amazon_google.py` or the `amazon_google/` dataset (W2).
- Self-reviewed (langres-code-reviewer); fixed warranted findings in this PR (single-source `CASCADE_LLM_DECISION_STEP`, call-shape-correct mocks).

### Note for W4
The implemented `CascadeModule` is **embedding → LLM** (no rapidfuzz stage); the task brief's "rapidfuzz→embedding→LLM" doesn't match the existing class, which was left unchanged (out of scope). W4 swaps the mock client for the real one via `make_resolver_factory(..., llm_client=<real>)`.

https://claude.ai/code/session_01Qik2ZVTbXtYt6rx2uLkYSd

### DIFF
--- a/src/langres/core/__init__.py
+++ b/src/langres/core/__init__.py
@@ -35,6 +35,7 @@ from langres.core.indexes import (
     QdrantHybridIndex,
     VectorIndex,
 )
+from langres.core.judges.embedding_score import EmbeddingScoreJudge
 from langres.core.judges.weighted_average import WeightedAverageJudge
 from langres.core.models import (
     CompanySchema,
@@ -80,6 +81,7 @@ __all__ = [
     "ComparisonVector",
     "ComponentSpec",
     "EmbeddingProvider",
+    "EmbeddingScoreJudge",
     "EntityProtocol",
     "ERCandidate",
     "ErrorExample",
--- a/src/langres/core/judges/__init__.py
+++ b/src/langres/core/judges/__init__.py
@@ -2,9 +2,14 @@
 
 A judge is a :class:`~langres.core.module.Module` that turns candidate pairs
 into :class:`~langres.core.models.PairwiseJudgement` scores. Wave 2a ships the
-:class:`~langres.core.judges.weighted_average.WeightedAverageJudge`.
+:class:`~langres.core.judges.weighted_average.WeightedAverageJudge`; M3 adds the
+zero-spend :class:`~langres.core.judges.embedding_score.EmbeddingScoreJudge`.
+
+Importing both here fires their ``@register`` decorators on package import, so a
+fresh process can ``Resolver.load`` an artifact using either judge.
 """
 
+from langres.core.judges.embedding_score import EmbeddingScoreJudge
 from langres.core.judges.weighted_average import WeightedAverageJudge
 
-__all__ = ["WeightedAverageJudge"]
+__all__ = ["EmbeddingScoreJudge", "WeightedAverageJudge"]
--- a/src/langres/core/judges/embedding_score.py
+++ b/src/langres/core/judges/embedding_score.py
@@ -1,0 +1,111 @@
+"""EmbeddingScoreJudge: the zero-spend embedding-cosine scorer Module.
+
+This judge is the scorer slot for the ``embedding_cosine`` method. Unlike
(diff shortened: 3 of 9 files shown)


fxd24:
Thanks @codex — both findings addressed (doc/comment only, no behavior change), commit 42e0937:

- **P2 rapidfuzz empty-vs-empty match:** confirmed `fuzz.ratio("","")==100`. Kept the `-> ""` convention (it's RapidfuzzModule's documented pattern) and documented that this missing-field asymmetry vs the missing-aware `weighted_average` is an *intrinsic* property of the classical baseline the race is meant to surface — not a wiring bug. Making rapidfuzz artificially missing-aware would require changing RapidfuzzModule (out of scope) and would misrepresent the baseline.
- **P2 cascade client shape:** real W4 footgun — documented in `make_resolver_factory` + `_build_cascade_module` that `cascade` needs an OpenAI-shaped client (`chat.completions.create`) distinct from `llm_judge`'s LiteLLM `.completion()` client, so W4 must not reuse one client for both. (Tests already inject the call-shape-correct mock per method.)

PR #65, opened 2026-07-02
W1.1: SelectJudge (ComEM-style set-wise, call-count cost lever)

fxd24:
## Summary

- `SelectJudge` (`src/langres/core/modules/select_judge.py`): a `GroupwiseModule` (W1.0 contract) that makes **ONE LLM call per anchor group** ("select the single candidate, if any, that matches the anchor") instead of K pairwise calls, decomposed into per-member `PairwiseJudgement`s (`score_type="prob_group_llm"`). DummyLM-driven dev throughout — $0 spend.
- Group-call cost convention (E5) reused verbatim via `stamp_group_cost`: full call cost on the first judgement, $0 on siblings, `provenance["group_id"]` on all — with a dedicated test asserting `sum(cost over group) == one call's cost`.
- `select_error` handling (CEO #12), never raises mid-stream: a malformed LLM response (`AdapterParseError`), a selection referencing a candidate outside the group, or a selection of more than one candidate all map to whole-group "no match" judgements carrying `provenance["select_error"]`. Each sub-case has a dedicated test, plus a "continues to next group" test proving one group's error doesn't kill the stream.
- Design grounding (verified via arxiv/COLING, not assumed): Wang et al., "Match, Compare, or Select? An Investigation of Large Language Models for Entity Matching" (COLING 2025) describes ComEM's selecting strategy as choosing **"the"** (single) most-likely match — this is what makes "selected multiple when one is expected" a well-defined error rather than an arbitrary design choice.
- Wired into `langres.methods._make_module_builder` + `LLM_METHODS` tuple as `"select_judge"` — selectable by name in the benchmark harness, mirroring the `dspy_judge` branch exactly (lazy import, DSPy-LM injection, honest per-call price from the OpenRouter table). **Not** wired into `Resolver.from_schema(judge=...)` or the verbs' `judge="auto"` dispatch (out of this branch's scope — it is a benchmark-harness method, not yet part of the zero-label default path).
- `docs/TECHNICAL_OVERVIEW.md` updated: the W1.0 SelectJudge skeleton now points at the shipped implementation and the results doc.

## Call-count + honest-cost reduction table (Amazon-Google, seed 0, k=50)

| quantity | naive pairwise judge | SelectJudge | reduction |
| --- | --- | --- | --- |
| test records | 1,374 | 1,374 | — |
| candidate pairs in scope | 48,854 | 48,854 | — |
| LLM calls | 48,854 (1/pair) | 1,374 (1/group) | **35.56x** |
| cost at $P/call | 48,854 × $P | 1,374 × $P | **35.56x** |

Group-size distribution (E3 — rules out a skewed-distribution artifact): min 2, max 50 (the `k_neighbors` cap), mean 35.56, **zero** groups at the trivial size-1 floor, 86% of groups at size ≥ 20. Full histogram in `data/benchmarks/w1/W1_RESULTS.md`.

`select_judge` also runs end-to-end through `run_method` on the real `AmazonGoogleBenchmark` (harness plumbing proof — `pair_f1=0.000` is expected since the injected DummyLM always answers "no match"; `bcubed_f1=0.8545`, `usd_total=$0`). The `stream_groups()` ≡ `stream()` pairs-equivalence property (CEO #14) is re-verified inline with a real `GroupwiseModule` consumer attached — no dupes, no losses across the 48,854 pairs.

Quality replication vs ComEM's published gains is explicitly **deferred to W3** (the $3 paid verification gate) — documented as a note in the results doc, not measured here (per the plan's zero-spend-except-final-smoke rule).

Reproduce: `uv run python examples/w1_select_judge_benchmark.py` (deterministic, $0, writes `data/benchmarks/w1/select_judge_amazon_google.json`).

## Coverage / quality gate

```
uv run ruff check src tests          -> All checks passed!
uv run ruff format --check src tests -> 170 files already formatted
uv run mypy src                      -> Success: no issues found in 66 source files
uv run pytest -m "not slow and not integration" --cov=src/langres --cov-report=term-missing
  -> 1419 passed, 98.13% total coverage
  -> src/langres/core/modules/select_judge.py: 100%
  -> src/langres/methods.py: 100%
```

## Commits (TDD, tests-first per commit)

1. `test:` SelectJudge unit tests (20 tests) — verified failing (`ModuleNotFoundError`) before implementation.
2. `feat:` SelectJudge implementation — all tests green, 100% cov, mypy/ruff clean.
3. `test:` methods.py registry wiring tests — verified failing (`unknown method 'select_judge'`) before wiring.
4. `feat:` methods.py wiring — all tests green.
5. `docs+bench:` benchmark script + results doc + TECHNICAL_OVERVIEW.md update.

## Test plan

- [x] All 20 `SelectJudge` unit tests pass (happy paths across 2 schemas, cost convention, 3 `select_error` sub-cases, empty-group skip, `inspect_scores`, lazy-LM construction)
- [x] All 5 `methods.py` wiring tests pass (factory build, price wiring known/unknown model, membership tuple, `run_method` harness race)
- [x] Full non-slow/non-integration suite green (1419 passed), 98.13% total coverage, 100% on new/changed files
- [x] mypy strict clean, ruff clean
- [x] Benchmark script re-run twice, deterministic, $0, no stray parse-errors
- [x] `stream_groups()` ≡ `stream()` property re-verified with SelectJudge in the loop

https://claude.ai/code/session_0115MvWbHaCKT1A39PKDWqey


fxd24:
Addressed the Codex P2 finding: **SelectJudge had no `type_name`/`config`/`from_config`/registry entry, so `Resolver.save()` raised `TypeError` for any resolver holding a `SelectJudge` in its module slot.**

Fix (TDD, serialization-only scope, mirrors `DSPyJudge`):
- `type_name = "select_judge"` + `@register("select_judge")` + lazy entry in `registry._LAZY_COMPONENT_MODULES` (so `import langres.core` still never imports `dspy` — verified by the existing `test_import_langres_core_does_not_import_dspy` slow test, still green).
- `config`/`from_config` (model/temperature/entity_noun, never the LM) — pure, JSON-able, matching the no-pickle config-registry contract.
- No `save_state`/`load_state`: unlike `DSPyJudge`, `SelectJudge` has no `compile()`/out-of-band program state in this scope, so `from_config` alone rebuilds an equivalent judge — the same shape as any other stateless judge (`LLMJudge`, `WeightedAverageJudge`).

5 regression tests added first (commit `3af0507`), verified failing against the pre-fix code (`UnknownComponentType`, `AttributeError`, `TypeError: ... not serializable`), then the fix (commit `5a67f16`). The fresh-process test (`@pytest.mark.slow`, subprocess) goes beyond a type check — it re-injects a `DummyLM` into the reloaded judge and calls `forward_groups()` on a real group, confirming the round-tripped judge can actually score, not just deserialize.

Gate: `ruff check/format`, `mypy strict`, and `pytest -m "not slow and not integration" --cov=src/langres` all green (1423 passed, 98.14% total coverage); the slow fresh-process + `test_import_langres_core_does_not_import_dspy` tests also verified green in an explicit follow-up run.

`select_judge` is still **not** wired into `Resolver.from_schema(judge=...)` or the verbs' `judge="auto"` dispatch — that remains the tracked, larger follow-up (needs the group-blocking path plumbed through `Resolver`), unchanged from the original PR description.

PR #71, opened 2026-07-03
W2.2: incremental single-record assign (AnchorStore + resolver.assign -> ClusterDelta)

fxd24:
## W2.2 — Incremental single-record assignment (`AnchorStore` + `resolver.assign`)

Implements M5's incremental-linking exit (S6), per the gated W2.2 design memo (decisions D1–D6).

A batch `resolve()` answers "which of these records group together?" but returns only id-sets (positional, no stable ids, **singletons dropped**). This adds the *incremental* answer: given an anchor set built from a prior batch, `resolver.assign(new_record)` returns a `ClusterDelta` — `link` to a **stable** existing entity id, or `new`.

### Public surface
- **`AnchorStore`** (`langres.core`) — a serializable, composable unit *around* a `Resolver`:
  - `AnchorStore.build(resolver, records) -> AnchorStore` — dedicated build pass; mints a stable, monotonic entity id for **every** record, including the singletons `resolve()` drops (clusterer-agnostic — it enumerates the record-id universe itself). (**D2, D3**)
  - `store.assign(record) -> ClusterDelta` — reuses the vector index's single-record kNN (`index.search`) for candidates when the resolver blocks on a vector index, else all-pairs; runs the **same** Comparator + Module judge (no new matching logic). Append-only allocator ⇒ idempotent per record id; ids never renumber. (**D4, D5**)
  - `store.save(path)` / `AnchorStore.load(path)` — the existing config-registry artifact seam (no pickle); delegates the pipeline to `Resolver.save`/`load` (incl. a built FAISS index's sidecar), plus a small `anchor_store.json`.
  - `entity_id_of(record_id)`, `entity_ids`.
- **`ClusterDelta`** (`langres.core`) — `type` (`new` | `link`; `merge`/`split`/`reject` **reserved** in the enum so the contract is stable — **D5**), `record_id`, `entity_id`, `matched_anchor_ids`, `score`, `reasoning`.
- **`Resolver.build_anchor_store(records)`** + **`Resolver.assign(record)`** — thin sugar over `AnchorStore`. The reserved cross-source `Resolver.link()` / `stream_against()` `NotImplementedError` stubs are **untouched** (three meanings of "link" kept distinct — **D1**).

### Exit criteria — all met
- Sparse record matching an existing entity → `link` (stable id); matching nothing → `new`. Tested on committed **Fodors-Zagat** data (offline string pipeline): held-out gold partner `z219` links to `f534`'s entity; a novel record → `new`.
- **Singleton coverage** tested: a record equal to a known unique anchor links (not spurious `new`).
- **Stable ids** tested: same record twice → same id; interleaved assigns never renumber prior ids.
- **Fresh-process save/load round-trip** (the M2 lesson): committed-data test loads the saved store in a clean subprocess and assigns identically. Two-process transcript:

```
===== process 1 (build + save) =====
[PROC 1 build+assign] z219 -> link e0  (f534 entity=e0)
[PROC 1] saved store to tmp/pr_roundtrip_store
===== process 2 (separate interpreter: load + assign) =====
[PROC 2 fresh load+assign] z219 -> link e0  matched=['f534', 'f543', 'f545']
```

- **100% coverage** on new `anchor_store.py` (129 stmts / 26 branches, 0 missed); `mypy --strict` clean (75 files); `ruff` clean. Full non-slow suite: **1671 passed**, 98.44% total coverage.

### Design-for-reuse
`AnchorStore` is a standalone serializable component (not wired into Resolver's guts — it uses only the public blocker/comparator/module/clusterer slots), so W2.3 golden records / W2.4 flywheel / M6 build on it directly. `ClusterDelta`'s type enum reserves the maintenance operations (merge/split/reject) up front so the contract won't reshape.

https://claude.ai/code/session_0115MvWbHaCKT1A39PKDWqey

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,26 @@
 - Implemented core primitives (`Module`, `Blocker`, `Clusterer`) with Pydantic data contracts and 100% test coverage
 - Completed Approach 1 (classical baseline): `AllPairsBlocker` + `RapidfuzzModule` end-to-end pipeline
 
+### M5 (W2.2): incremental single-record assignment — `AnchorStore` + `Resolver.assign`
+
+The incremental-linking exit (S6): after a batch `resolve()`, answer "here is one NEW
+record — which existing entity, or new?" with a **stable** entity id.
+
+- **`AnchorStore`** (`src/langres/core/anchor_store.py`) — a serializable, composable unit
+  around a `Resolver`. `AnchorStore.build(resolver, records)` runs a dedicated pass that
+  mints a stable, monotonic entity id for **every** record, including the singletons
+  `resolve()` drops (clusterer-agnostic). `save`/`load` via the config-registry artifact
+  seam (no pickle), delegating the pipeline to `Resolver.save`/`load`.
+- **`Resolver.build_anchor_store(records)` + `Resolver.assign(record) -> ClusterDelta`** —
+  thin sugar; the reserved cross-source `link`/`stream_against` stubs stay untouched.
+  `assign` reuses the vector index single-record kNN (with `similarity_score` + `query_prompt`,
+  so `EmbeddingScoreJudge` works incrementally) or all-pairs, and the same Comparator + Module
+  judge. Append-only allocator (idempotent per record id); `CompositeBlocker` supported.
+- **`ClusterDelta`** — `new` / `link`, with `merge`/`split`/`reject` reserved in the enum so
+  the contract stays stable for W2.4/M6.
+- Committed-data (Fodors-Zagat) + fresh-subprocess save/load round-trip tests; 100% coverage.
+  See `examples/incremental_assign.py`.
+
 ### M4: langres is the seam — DSPy experimentation foundation + first paid signal
 
 Reframed from a distillation-metric chase to **building the composable scorer seam we're
--- a/docs/TECHNICAL_OVERVIEW.md
+++ b/docs/TECHNICAL_OVERVIEW.md
@@ -438,6 +438,17 @@ The rich data object passed out of a Flow. This is the auditable log of a decisi
 - `reasoning: Optional[str]`: The LLM's natural language explanation.
 - `provenance: Dict[str, Any]`: A full audit trail (e.g., `{"model": "e5-small", "rapidfuzz_score": 0.85}`).
 
+### ClusterDelta (`langres.core`, M5/W2.2)
+
+The result of one incremental `Resolver.assign(record)` / `AnchorStore.assign(record)` — the outcome of assigning a single new record against a prior batch's anchor set.
+
+- `type: Literal["new", "link", "merge", "split", "reject"]`: `new` (a fresh entity was minted) or `link` (attached to an existing entity). `merge`/`split`/`reject` are **reserved** for the wider entity-maintenance surface (W2.4/M6) so the contract stays stable; W2.2 only ever emits `new`/`link`.
+- `record_id: str`: The assigned record's id.
+- `entity_id: str`: The **stable** entity id the record now belongs to (freshly minted for `new`, existing for `link`). Never changes on later assigns (append-only allocator).
+- `matched_anchor_ids: list[str]`: Anchor record ids that cleared the threshold (evidence for a `link`; empty for `new` and for the idempotent already-assigned-id `link`).
+- `score: Optional[float]`: Best matching score across judged candidates (observability); `None` when there were no candidates or on the idempotent path.
+- `reasoning: Optional[str]`: Human-readable note about the decision.
+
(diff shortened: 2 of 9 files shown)


fxd24:
Thanks — addressed in `2f98308`:

- **1 (CompositeBlocker crash):** added `_schema_factory(blocker)` — `build`/`assign`/`_candidate` now raise a clear `NotImplementedError` naming the unsupported blocker instead of a cryptic `AttributeError`, and document that composite/hybrid-blocked resolvers are a later wave. Tested (`test_unsupported_blocker_without_schema_factory_raises`).
- **2 (records not re-indexed across assigns):** taking your "document as a known limitation deferred" option — this matches the intended W2.2 scope (assign against the build-time anchor set). Documented explicitly on `assign()`/the class, and **locked with a test** (`test_new_records_are_not_added_to_the_searchable_set`) so the boundary is now signalled, not silent. Growing the searchable set is reserved for a later wave alongside the reserved `merge`/`split` delta types.
- **3 (dropped -1 guard breaks Qdrant):** good catch — restored the `index >= 0` skip with a comment on why (fusion indexes pad with -1 even when `k <= ntotal`). Covered by `test_vector_search_skips_minus_one_padding` (a fake padding index; asserts no wrap to the last anchor).
- **4 (example private access):** added a public read-only `AnchorStore.assignments` property; the example uses it now.

100% coverage on `anchor_store.py` holds; full non-slow suite green (1678 passed).

fxd24:
## Round-3 review fixes (claude-review round-2) — both warranted, both addressed

Pushed as `e52bb5b`. CI green on 3.13 (test, lint, security, claude-review all pass). Full `pytest -m "not integration"` (1764 passed, incl. slow) run locally before push.

### Fix 1 — recurse into nested `CompositeBlocker`s
Both the schema-factory resolution and the vector-candidate-source lookup previously handled only a single level of `blocker.children`, while `CompositeBlocker` documents composite-of-composites as supported — the same recursion-gap class fixed for the index build in #66.

- **`_schema_factory`**: extracted a depth-first `_find_schema_factory` helper that walks `children` at arbitrary depth; a `CompositeBlocker` wrapping another `CompositeBlocker` now surfaces a nested child's factory. Unreachable case still raises the same actionable `NotImplementedError`.
- **Vector-source lookup (`_candidate_anchors`)**: now **reuses the resolver's own `_iter_vector_blockers`** — the *same* recursive walk `resolve()` builds the index with — instead of a one-level `type_name == "vector_blocker"` check. This isn't just mirroring #66's pattern: reusing the canonical walk guarantees `assign()` searches *exactly* the index that was built for that blocker, so the two can never drift. A `VectorBlocker` nested at any depth is now searched against its real index rather than silently falling back to the all-anchors (`similarity=None`) path. Removed the now-unused `# type: ignore[attr-defined]` comments (the var is `VectorBlocker`-typed after the walk).
- **Test**: `test_nested_composite_blocker_supports_build_and_vector_assign` — a depth-2 `outer(inner(VectorBlocker, KeyBlocker), KeyBlocker)`, asserting (a) `build` derives every anchor id (schema-factory recursion), (b) `_candidate_anchors` returns **non-None** similarities (proving the walk reached the depth-2 `VectorBlocker`, not the fallback), and (c) end-to-end `assign` links through the nesting.

### Fix 2 — stop unbounded `_records` growth on `assign()`
`assign()` did `_records.setdefault(record_id, record)` for every newly-assigned record, but nothing ever reads a non-anchor record back: `_candidate()` only reads `_records[anchor_id]` for `anchor_id in _anchor_ids`, and the idempotent path keys on `_assignments`. So the raw payloads were dead weight that grew in-memory **and** in `anchor_store.json` for exactly the long-running incremental-ingestion case this targets.

- Removed the `setdefault`; `assign()` now grows only the `record_id -> entity_id` map. Verified (repo-wide grep) that the only `_records[...]` read is `_candidate()`, always keyed by an anchor id. `_records` = anchors only, matching the `AnchorStoreManifest.records` docstring.
- Docstring updated (assign W2.2-boundary note + the register comment) to state the payload is intentionally not retained.
- **Test**: `test_assign_does_not_store_non_anchor_records` — after 5 non-matching assigns, `_records` (and the persisted manifest's `records`) is unchanged (anchors only), while `assignments` grew (idempotency preserved).

### Nit — `build()` re-normalization
Added a one-line docstring note explaining the intentional, build-only duplicate `factory(record)` pass (keeps id derivation self-contained; `assign()` doesn't pay it). Not deduped — the cost is one local pass, no network.

100% coverage on `anchor_store.py` (155 stmts / 40 branches), mypy strict + ruff clean.

### Codex inline comments
The 3 open Codex inline comments are the already-addressed round-1 P1/P2s (similarity_score, query_prompt, composite support) — they just don't auto-resolve.

PR #79, opened 2026-07-07
feat: close the flywheel loop — fail-fast auto, review selector, CLI, cascade, GETTING_STARTED (0.2.0, BREAKING)

fxd24:
## Close the flywheel loop (0.2.0)

langres closes a loop most ER tools leave open: an expensive judge bootstraps **silver** labels → every judgement is logged → a human reviews only the **uncertain margin** → those labels train a **cheap student** → a **cascade** runs the student everywhere and escalates only the still-uncertain pairs back to the expensive judge → disagreement seeds the next review round. This PR ships every missing piece of that loop plus one entry doc that tells the lifecycle end to end.

## ⚠️ BREAKING (lead with this)

- **`judge="auto"` (the default for `link`/`dedupe`) now RAISES `NoJudgeAvailableError` when no LLM API key is set**, instead of silently falling back to fuzzy string matching. Unsupervised string matching over-merges on unlabeled data — in the motivating demo it collapsed five distinct entities into one cluster with no error. The library refuses rather than hand back a confidently-wrong answer. The unpinned-model-price branch raises the same error.
- **`fallback_reason` removed** from `DedupeResult` / `LinkVerdict` / `ResolvedModule` / `ResolvedJudge` (no path could set it after fail-fast; always-`None` is anti-self-describing).

### Migration
- Keyless callers: pass `judge="string"` explicitly to opt into offline fuzzy matching (lower quality; pair with `derive_threshold` on labeled data).
- Keyed default path: install the `[llm]` extra (`uv sync --extra llm` / `pip install 'langres[llm]'`) **and** export `OPENROUTER_API_KEY`; runs spend-capped at `$1` by default (`budget_usd=`).
- Catch `NoJudgeAvailableError` (now root-exported from `langres`, alongside `BudgetExceeded`).
- Replace `result.fallback_reason` reads with `result.judge_used` / `result.score_type` + the auto-path selection notice.

## What's included

| Piece | Summary |
|---|---|
| **Fail-fast `auto`** | `NoJudgeAvailableError`; dspy `ImportError` wrapped with install guidance; `model=` honored on the auto path; `len(records)<2` short-circuit |
| **`select_for_review` + `ReviewQueue`** (`langres.core.review`) | uncertainty / disagreement / first-class **audit** strategies; snapshot queue; root-exported |
| **`langres` CLI** (`langres.cli`) | `review` (resumable terminal labeler), `export-csv`/`import-csv` (spreadsheet round-trip = primary path), `--version`; formula-injection + terminal-control-char hardened; stream-injectable |
| **`CascadeJudge`** (`cascade_judge`) | cheap student + escalation-in-band; provenance preserves `cost_usd`/`model`; serializes a fitted student through `Resolver.save`/`load`; old `CascadeModule` deprecated |
| **Silver-only guard** | `derive_threshold_from_pairs` warns when calibrating on a judge's own verdicts (circular) |
| **`examples/flywheel_closed_loop.py`** | the whole loop at **$0** on committed Fodors-Zagat fixtures, data-derived escalation band, honest "plumbing not economics" report |
| **`docs/GETTING_STARTED.md`** | the lifecycle at altitude, two lanes (keyless/keyed), inline runnable snippets, competitive positioning (dedupe/Zingg/Splink), snippet-rot acceptance test |

## Tests & quality

- **Full `not integration` suite green: 1980 passed**, total coverage **99.36%** (gate 95%).
- New code at **100%**: `cli.py`, `core/review.py`, `core/modules/cascade_judge.py`.
- `mypy src/` strict clean; ruff clean.
- Two findings from an independent integration review were addressed: a spend-cap-placement footgun (documented — caps belong outside a cascade) and an unvalidated `audit_fraction` (now range-checked). The `select_for_review` disagreement sort had a dead sort key + a weak test — both fixed with a hardened regression test.

## Paid validation (gate-authorized ≤$12; actual spend tiny)

Committed, re-runnable, SpendMonitor-capped scripts under `examples/research/` (`$0`-verified with a proven cap-trip test); results in `data/benchmarks/flywheel/`:

- **T8(a) Fodors-Zagat wiring smoke** — real `gpt-4o-mini`, **$0.066 spent** (cap $2). teacher F1 0.944 / student 1.000 / cascade 0.944; escalation 31.7%, frontier-call reduction 68.3%, escalated-pair accuracy 89.5%. FZ is easy, so the student already saturates — this proves the wiring on a live model, not the economics.
- **T8(b) Amazon-Google economics** (hard dataset, real teacher/student gap; cap $8) — real `gpt-4o-mini` teacher, **$1.05 spent**. Bounded 2000-pair AG subset (seed 7): escalation 22.5%, **frontier-call reduction 77.5%**, escalated-verdict accuracy 84.9%; pairwise F1 teacher 0.557 / student 0.302 / **cascade 0.381**.
  - **Honest hard-data reading** (the plan's "economics live in T8's real results"): the **cost lever is unambiguous** — 77.5% fewer frontier calls for $1.05. The **cascade beats the cheap student** (F1 0.302 → 0.381) but sits **below the teacher** (0.557) — because AG is hard, so the *uncompiled* frontier teacher is itself low-precision (0.407), its silver labels are noisy, and a student trained on only 30 corrections + that noisy silver is weak. A cascade can't exceed the signal it distils; **more human labels at the margin** is the lever that closes the gap — which is exactly what the loop accumulates. This is the mechanism validated on real data, honestly measured — not an over-claim that a lightly-trained student already matches the frontier. Full reading + caveats in `data/benchmarks/flywheel/flywheel_amazon_google_results.md`.

**Total paid validation spend: $1.11** (FZ $0.066 + AG $1.05), well under the $12 gate authorization.

## Deferred (recorded in `TODOS.md`)

Blocker-recall diagnostics (P2); `methods.py`→`CascadeJudge` migration, Label-Studio export, stratified-audit knob, `langres select` subcommand (P3).

---


fxd24:
Thanks — all five items triaged. None blocking; addressed the two warranted-for-clarity ones + both cosmetic nits, kept the two by-design ones with a note:

- **`registry.py` `cascade_judge` lazy entry** (the one you'd want addressed): kept, with a clarifying comment. It isn't dead — it's the same *saved-artifact safety net* its pure-core peers rely on. I verified all six checked entries (`cascade_judge`, `rf_judge`, `correlation_clusterer`, `key_blocker`, `composite_blocker`, `fellegi_sunter_judge`) are eager-imported today, so the "eager ⇒ dead" logic would condemn all of them equally — but they're deliberately mapped so a saved `Resolver` artifact keeps resolving if a future eager-import trim (like W0.4's) drops them from `core/__init__.py`. A `CascadeJudge` wrapping a fitted student is exactly what lands in such an artifact, so removing it would make it the *lone unprotected* core judge. Added a comment paragraph naming it explicitly so it isn't re-flagged. (`f8a7c80`)
- **CSV id-column formula-injection**: added the residual-risk doc note you suggested — ids are assumed internal/trusted; a `-`/`=`-leading id from an *untrusted* source is a documented tradeoff for the lossless round-trip, not a bug. (`93e6552`)
- **Cosmetic test nits**: both fixed — `tmp_path: Path` (dropped the `object` downcast + `Path(str(...))` re-wrap); `test_flywheel_paid.py` docstring now says "three things" and enumerates the preflight-refusal category that was already in the body. (`f8a7c80`)
- **`_pair_key` duplication (rule of three)**: left as-is — deliberate anti-coupling, already commented at each of the three call sites; agreed it's a flag, not a fix, until a fourth copy appears.

CI is green; the Amazon-Google economics run is finishing and its numbers will land in the PR body shortly.

fxd24:
Round-2 review triaged — thanks, these were sharp. No blockers; addressed the three doc-syncs, and the three robustness items are by-design (explained) with refinements tracked. Details:

**Doc-sync (fixed, `4f30715`):**
1. **CLAUDE.md Project Structure** — added `cli.py`, `core/review.py`, `core/registry.py`, `core/harvest.py` and named `CascadeJudge` in the `modules/` line. (Minor: `harvest.py`/`registry.py` pre-exist on `main` rather than being new here, but I added them for accuracy since the tree omitted them.)
2. **component-design.md "Real components"** — now lists `CascadeJudge`, the review seam (`select_for_review`/`ReviewQueue`), and `core.harvest`.
6. **pyproject `[project.scripts]` nit** — the hand-edit exception comment already lived at `pyproject.toml:51`; I completed the doc-sync by carving the exception into `python-style.md`'s "never hand-edit pyproject.toml" rule so it doesn't read as a violation.

**Robustness — by-design, refinements tracked in `TODOS.md` (P3):**
3. **`import-csv` duplicate `Correction`s on re-run** — append-always is *intentional*: `harvest_labeled_pairs` is last-write-wins by pair, so a re-import is how you **update** a label. A naive `answered`-set skip (like `_review`'s) would silently drop legitimate label updates. Non-corrupting; the audit-trail bloat on *identical* re-runs is the only cost. Tracked a smarter "skip-identical-keep-changed" guard.
4. **`ReviewQueue.write` truncate-in-place vs atomic** — deliberate: the queue is a *derived, regeneratable snapshot* (regenerate from the log; never hand-edit), so a half-written queue is recoverable by re-selecting. The durability care (append-immediately) correctly lives on the append-only source-of-truth logs, not the snapshot. Atomic temp-then-`os.replace` tracked as a nicety.
5. **No packaged-entry-point test** — agreed it's a gap; `main()` is fully covered in-process. Tracked a console-script resolve/subprocess test.

**Questions:**
- **Committed `data/benchmarks/flywheel/*.json|md`** — intentional **permanent research record**, following the repo's existing `data/benchmarks/` convention. No secrets (verified); each doc labels its mode (`simulated: false`, REAL/paid) and is re-runnable via the committed `examples/research/` scripts.
- **`harvest`/`Correction`/`CorrectionLog` out of the root `langres` export** — intentional **tiering**. The review/selection surface (`select_for_review`, `ReviewQueue`) is the front-door flywheel API → root-exported; `harvest` is eval/calibration-tier → stays in `langres.core.harvest`, consistent with CLAUDE.md's dependency-tier language (harvest/calibration are not on the `link()`/`dedupe()` runtime path).

CI is green on the prior head; this commit is docs-only. Re-verifying green now.

fxd24:
Round-3 triaged and fixed (pushed in `fe8432f`). This round found real issues — thanks. 9 addressed, 2 skipped with rationale:

**Correctness (fixed):**
- **`import-csv` now reads `utf-8-sig`** — the important one. Excel/Sheets "CSV UTF-8" export prepends a BOM that left `left_id` reading as `﻿left_id`, tripping "missing required column" on the *advertised* spreadsheet round-trip. Fixed + BOM-CSV test.
- **`_review` checks the queue file exists** — a typo'd `--queue` now returns a clear "not found" (exit 1) instead of a misleading "queue is empty". + test.
- **Negative `limit` guarded** in `select_for_review` — uniform upfront `limit >= 0` check so `strategy="audit"` fails with a clear error instead of a raw `random.sample` ValueError. + test.
- **`ReviewItem.score` constrained to `[0, 1]`** (`Field(ge=0, le=1)`) — an out-of-range score in a hand-edited queue now surfaces as the documented "corrupt queue line N" error (integrates with `read()`'s existing ValidationError wrap). + test.

**Hardening (fixed):**
- **ids sanitized in the terminal render** — `_render_item` now runs `left_id`/`right_id` through the same control-char/ANSI stripper as record content (display-only; CSV id columns stay byte-exact). + ANSI-id test.
- **`--reviewer` added to `import-csv`** — CSV-sourced corrections were always `reviewer=None`, losing attribution on the CSV-primary path. + test.

**By-design (documented + tested):**
- **Non-`BudgetExceeded` escalation failure** — the partial-preservation is deliberate (only `BudgetExceeded` carries the `partial_judgements` field the log slice needs; a network/API error is a hard failure). Documented in `forward`'s `Raises:` + a regression test asserting a generic exception propagates unchanged.

**Docs:**
- Reworded the dedup docstring to "last-in-input-sequence-wins (not by `timestamp`)" with a note to sort concatenated logs first; added `judgement_log.py`/`reports.py` to the CLAUDE.md tree.

**Skipped (with rationale):**
- **Upfront API-key validation** — would require a live network probe on every keyed call; standard practice is to fail on the first real call with the upstream error. Not worth the added latency/cost.
- **Narrowing the dspy `except ImportError`** — low-probability masking; the message already points at the right fix (`install [llm]`) in ~all real cases.

Verified: full suite **1999 passed** (0 failures), `mypy src/` clean, ruff clean, **100% coverage** on all three changed source files (`cli.py`, `review.py`, `cascade_judge.py`).

Also in this push: the automated review now triggers only on PR **open/reopen**, not every push (`synchronize` dropped) — so iterating no longer stacks a full review per commit.

Issue #86, opened 2026-07-07
Reusable hard-case / informative-pair mining seam (data prep for training + DSPy tuning)


fxd24:
**Related:** #85 (experiments epic — esp. #83 AnyMatch) · #79 (flywheel — already ships the seed) · #55 (research backlog) · #20 (epic)

## The insight

AnyMatch's headline result — a **124M** model beating GPT-4 on entity matching — comes overwhelmingly from **data preparation**, not model size: mining the **hard / informative pairs** (the AutoGluon-misclassified positives) + attribute augmentation + label balancing. The lesson generalizes: **the data recipe is the lever.** If langres exposes a *reusable* way to mine hard cases and feed them into training and/or prompt optimization, that lever applies to **many new use cases** — a path to "beat these numbers" broadly, on a swappable seam, not just on AG/Abt-Buy.

## What langres already has — generalize, don't rebuild

The flywheel (shipped 0.2.0, #79) already contains the seed: **`select_for_review`** picks the **uncertainty margin** (|score − threshold|) and **judge-disagreement** pairs, and the **audit** slice catches confident false merges. **That *is* hard-case mining** — it's just currently aimed only at the human-review queue. This issue **generalizes that seam** so the same mined hard cases can *also* feed **training** and **DSPy prompt optimization**, with **pluggable mining strategies**. (This is the flywheel's hard-case selection turned toward the *model*, not only the human.)

## The reusable abstraction

A **hard-case / informative-pair mining seam** — `mine(candidates | judgement_log, strategy, …) -> ranked informative pairs` — with **pluggable strategies**:
- **uncertainty / margin** — already in `select_for_review`
- **judge-disagreement** — already in `select_for_review`
- **classifier-misclassification / AutoML** — AnyMatch-style: a cheap classifier's errors are the hard cases
- **hard-negative mining** — near-miss non-matches surfaced by the blocker (langres blockers already emit near-duplicate candidates)
- **(more from Task 0's survey)**

…whose output feeds **three consumers on one seam**:
1. **Fine-tuning** — curated hard-case set → the trained judge (`SupervisedFitMixin`), incl. AnyMatch-style balancing/augmentation.
2. **DSPy prompt optimization** — hard cases as the bootstrap/few-shot demo pool for `DSPyJudge` (MIPROv2 selects demos; feeding it *hard* demos is the same lever).
3. **Active-learning review** — the existing CLI labeling loop (unchanged — it's just one consumer of the seam).

[...]

Issue #85, opened 2026-07-07
Epic: ER method-replication & small-LM fine-tuning experiments


fxd24:
**Parent:** #20 · **Research source:** #55, `docs/research/20260701_er_seam_audit.md` · **Context:** seam-capability + SOTA review, 2026-07-07 (see #81 Task 0)

## Thesis

langres claims to be a **composable ER seam** — every matcher swappable behind one interface, plus a serializable artifact. The way to *prove* that (and to build genuinely competitive judges) is to **replicate the field's methods and fine-tune modern small models through the seam**, under an honest eval protocol, on license-clean data, on hardware we own. This epic groups those experiments so they share one set of foundations instead of each reinventing them.

## Shared foundations (apply to every experiment under this epic)

- **Honest eval protocol:** full standard test split (no subsampling) + threshold **pinned from train/valid** (no argmax-on-test). Always state the protocol; cross-paper F1 numbers are *not* apples-to-apples (Ditto-on-AG is reported as 0.66 / 0.756 / 0.80 across three papers).
- **Hardware:** QLoRA on an **8GB RTX 3070** via **Unsloth**; model ladder **Qwen3-0.6B** (fast-iteration/smoke) → **Qwen3-1.7B** (primary) → **Qwen3-4B** (stretch). Newer Apache-2.0 challengers (**Qwen3.5**, **Gemma-4-E4B**) added *after* a baseline works — **each verified empirically, never assumed** (incl. QLoRA tooling maturity and on-disk 4-bit size).
- **Two arms per model:** **DSPy-prompting** (`DSPyJudge`, MIPROv2) first, then **QLoRA fine-tune** — measure the lift each buys. (Breunig's Overture pipeline shows DSPy alone lifts Qwen3-0.6B 60.7%→82%.)
- **License hygiene:** Apache-2.0 / MIT bases; only permissive data (CC-BY-4.0 or better) for anything shippable/vendorable. OpenSanctions Pairs = **CC-BY-NC → research-only, not shippable**. Person-schema stays on FEBRL4.
- **Verification discipline:** every performance/tooling claim is **measured or sourced before it is stated** (CLAUDE.md · `.claude/rules/expert-knowledge.md`).

## Experiments

- [ ] **#80** — Replicate a **Ditto-class cross-encoder** (prove-the-seam SOTA baseline; AG ≈ 0.756 / Abt-Buy ≈ 0.893).
[...]

Issue #83, opened 2026-07-07
Experiment: replicate AnyMatch — a tiny-LM zero-shot entity matcher via hard-pair mining


fxd24:
**Epic:** #85 · **Sibling experiments:** #80, #81 · **Source:** AnyMatch (arXiv:2409.04073, AAAI-2025 GOOD-DATA workshop)

## Why replicate AnyMatch

AnyMatch shows a **124M-param GPT-2**, fine-tuned via smart *data curation*, lands within **~4% of GPT-4 MatchGPT** on entity matching at **~3,900× lower cost** — and generalizes **zero-shot to unseen EM datasets** (mean ~82 F1 across 9 sets). The lesson isn't the model, it's the **training-data recipe**: hard-pair mining + attribute augmentation + label balancing. That recipe plugs directly into langres's harvest/curation seam, and a zero-shot generalist matcher is exactly what a "drop-in cheap judge" should be. Replicating it proves langres can *curate + train* a competitive tiny matcher, not just host one.

## What's actually available (verified 2026-07-07)

- **Code:** [github.com/Jantory/anymatch](https://github.com/Jantory/anymatch) — but **no LICENSE file → all-rights-reserved by default**. We **reimplement the recipe from the paper**, we do not fork the code.
- **Curated dataset:** **not a download** — regenerate it by running their `automl_filter` (`utils/data_utils.py`) on the raw public Magellan/WDC datasets.
- **Weights:** not released.

## The recipe to reimplement

- **Hard-pair mining:** train an **AutoGluon** (AutoML) classifier on the source pool; keep the **positive pairs it *misclassifies*** as the "difficult" training set (if a decent classifier gets it wrong, it's genuinely informative).
- **Attribute-level augmentation:** add single-matching-attribute pairs (record stripped to one attribute) as auxiliary examples.
- **Label balancing:** final mix ~**2:1 negative:positive**.

## Plan

- [ ] Reimplement hard-pair-mining curation as a langres utility (fits the harvest/curation seam; complements the flywheel's own label curation).
- [ ] Fine-tune a small model (Qwen3 ladder, QLoRA on the 8GB RTX 3070) on the curated multi-dataset pool.
- [ ] Evaluate **zero-shot on held-out EM datasets** (train on N-1, test on the held-out one) under the **honest protocol** (full test split + train/valid-pinned threshold; inherit #80/#81).
[...]

PR #88, opened 2026-07-07
docs(research): data-prep, ER architecture & decoder-embedding survey (#86 Task 0)

fxd24:
## What

A research-only synthesis doc: `docs/research/20260707_data_prep_hard_case_mining_survey.md`. It is **Task 0 of #86** (survey data-prep / hard-case-mining methods *before* designing the reusable mining seam), plus two follow-up research threads the seam design depends on. **No code changes.**

Built via a multi-agent research fan-out; every method claim verified against primary sources (papers + code + model cards), with unverified points flagged explicitly.

## Sections
- **§1–10 — Data preparation & hard-case mining for EM.** AnyMatch's exact recipe (verified from its `automl_filter` code) + its lineage; a taxonomy of mining/selection methods organized by signal type; the label-noise trap and its named fix (Confident Learning/Cleanlab + Dataset Cartography); a map of what `select_for_review` already covers vs. new strategy candidates; a cost/value quadrant.
- **§11 — How SOTA divides recall (blocking) vs precision (matching).** The 4-stage consensus pipeline and why the split is structural (recall ceiling); a spectrum from crisp two-stage → no split (collective/relational ER); production reality check (Splink/Zingg/Dedupe/JedAI).
- **§12 — One model for both stages: decoder-LLM embeddings.** Decoder-embedding techniques (Echo, LLM2Vec, NV-Embed); GritLM's one-model precedent; current best embedding models by family (Qwen3-Embedding, EmbeddingGemma, Harrier-OSS); the ER counter-signal (UniBlocker deliberately chose encoder-only for structured records).
- **§13 — Future work.** The one-model-both-stages bets tiered by risk and framed around the **cross-role transfer question**; DSPy/GEPA prompt-tuning of the embedding instruction for recall (Bet C).
- **§14 — References.** ~100 primary sources, all with resolvable links.

## Key verified findings
- **AnyMatch is hard-*positive* mining + *random* negatives** (not hard-negative mining) — the seam should treat "mine hard positives" and "mine hard negatives" as two distinct pluggable strategies.
- `select_for_review` already implements the active-learning core (uncertainty margin, judge-disagreement, confident-merge audit); the new value is model-error mining, blocking-derived hard negatives, difficulty scoring, coverage/dedup, and label-noise filtering.
- The recall/precision split is **structural** — blocking is the only stage allowed to produce false negatives, so a match it drops is unrecoverable (recall ceiling).
- Best decoder embedder for langres's one-model idea: **Qwen3-Embedding-0.6B/4B** (Apache-2.0, same family as the matching ladder). GritLM shows one model can do both without degradation at 7B — but that's untested at 0.6–4B/QLoRA scale.

## Scope
Research + langres seam **mapping** only — no seam interface/API/function design (deferred), no code. Verify-before-assert throughout; unverified claims flagged in §10 and inline.

Task 0 of #86 · feeds #83 (AnyMatch) · epic #85


fxd24:
Addressed in `6db5c19` — thanks to both reviewers for the catch (independently flagged by claude-review and Codex).

Verified against source before fixing: `DSPyJudge` (`@register("dspy_judge")`) already exposes `compile(optimizer="bootstrap"|"mipro")` wrapping `dspy.BootstrapFewShot` / `dspy.MIPROv2` (`dspy_judge.py:297-312`), wired via `methods.py:219`. So the DSPy/MIPRO consumer is **not** prospective.

Corrections:
- **§8** — replaced the false "no DSPyJudge/MIPRO wiring in-repo yet (that consumer is prospective)" with the actual implementation pointer, noting §7's hard-demo pre-filtering is *testable against the existing seam*, not a from-scratch build.
- **§7** and **§13 Bet C** — de-framed from "prospective / the DSPyJudge plan" to "testable today against existing code."

This was a research-agent assertion I propagated without checking against the repo — exactly the class of thing the "verify before asserting" rule targets. No other code-quality/security/coverage findings apply (docs-only PR); the Codex review otherwise raised only this one P2.

PR #99, opened 2026-07-08
feat(tracking): experiment tracking & LLM observability (run-store + MLflow/W&B adapters + trace correlation)

fxd24:
## What & why

langres benchmark runs are **ephemeral**: `run_method`/`run_methods` produce rich Pydantic
results (`MethodResult`, `CostTrack`, the `reports.py` family) that are `.model_dump()`'d or
rendered to markdown, then lost — no run id, no config snapshot, no dataset/split identity, no
persistence, no cross-run comparison, no lineage. As the team reproduces published ER research
(Ditto/Jellyfish/AnyMatch, epic #85) *as clients of langres*, increasingly via agents, a result
must be **reproducible** months later and **comparable** across sessions.

Five-workstream research (`docs/research/20260708_experiment_tracking_observability_analysis.md`)
concluded: **don't build a tracker — add a thin pluggable adapter that wraps langres's existing
seams and reuses its existing Pydantic result models.** langres already has the hard parts (auto
per-judge capture via `LoggingModule`, honest cost via `SpendMonitor` + pinned prices, rich
serializable reports, a `"v":1` JSONL idiom). This PR adds the missing pieces: **run identity +
config/dataset/split capture + backend persistence + LLM-trace correlation + lineage.**

Scope = **Phase 0 (run-store + identity) + Phase 1 (adapter, BOTH MLflow & W&B) + Phase 3
(Langfuse correlation-id).** The `benchmark.py` harness wrap (Stream C) is **deferred** until
eval-readiness lands its `slice_fn` seam — so this PR is **disjoint from eval-readiness code**
(`benchmark.py` untouched) and targets `main` directly.

## The centerpiece — what gets tracked

Two frozen Pydantic models in a **dep-free** `core/runs.py`:

- **`RunContext`** — the recipe. `recipe_id = sha256(canonical_json(RECIPE_FIELDS))` over
  **inputs only** (experiment, resolver config, model, dataset fingerprint, split, seeds) —
  **excluding** git_sha/git_dirty/lockfile/timing, so an idempotent replay on a dirty tree still
  matches. `attempt_id = f"{recipe_id}-{started_at}"` is the record PK.
- **`RunRecord`** — `context` + outcomes (metrics as a reused-model dict, `metric_definition`,
  native `spend_usd`, status/failure, artifacts, timing), one JSONL line. `status="running"` is
  written **first**, so a crashed/torn-down run leaves a *visible* gap.

`RunStore` (JSONL, `fcntl.flock` per append, last-wins-by-attempt_id) is written
**unconditionally** by `capture_run` — trackers are the optional live mirror. Even with zero
trackers you get a replayable record.

## The layer

- **`ExperimentTracker` Protocol** (🤗 Accelerate `GeneralTracker` shape) + `NoOpTracker` (null
  default, zero overhead) + `MultiTracker` (fan-out → MLflow **and** W&B at once) +
  `resolve_tracker(spec)` mirroring langres's `judge="auto"|"string"|MyJudge(...)` idiom.
- **`MlflowTracker`** / **`WandbTracker`** — lazy adapters behind `[project.optional-dependencies]`
  `mlflow` / `wandb`; a missing extra raises a helpful `ImportError` naming the extra.
- **`Resolver.config_dict()`** — public config snapshot (factored from `save()`, byte-identical).
- **LLM correlation** — `LLMJudge.forward` stamps `metadata={langres_attempt_id,left_id,right_id,
  decision_step}` onto the litellm `completion()` **only** when a run is active; `JudgementLog`
  rows gain a nullable `run_id` — the exact three-way join (RunRecord ↔ JudgementLog ↔ Langfuse span).
- **DSPy lineage** — `DSPyJudge.compile()` optionally opens a `capture_run` (an optimization run
  that can parent the eval runs using its compiled program).

**Additive & byte-identical when unused:** omit store/tracker and behavior is unchanged — no files
written, same `MethodResult`. **Import budget preserved:** `import langres` pulls no
mlflow/wandb/ranx/litellm/torch (verified in `tests/test_import_budget.py`, 0.42s).

## Verification (beyond tests)

- ✅ **Import budget** — bare `import langres` leaks no heavy module; result-model refs are
  `TYPE_CHECKING`-only.
- ✅ **Real offline dual-backend e2e** — a zero-spend run mirrored to `["mlflow","wandb"]` +
  `RunStore`; inspected the JSONL `RunRecord` (recipe_id, git dirty flag, resolver_config, seeds
  union, dataset_fingerprint, metrics, spend_usd, `"running"`→terminal transition, `"v":1`).
- ✅ **Idempotent replay** — identical config → identical `recipe_id`; dirty tree flips `git_dirty`
  without changing the id.
- ✅ **No-tracker no-op** — omitting store/tracker is byte-identical.
- ✅ **mypy strict** clean; tracking modules at the `core/**` 95–100% tier (runs 100%, trackers
  100%, mlflow 98%, wandb 100%); full fast suite green.
- ✅ Ships a runnable offline demo (`examples/research/experiment_tracking_demo.py`) + a
  "Persisting & comparing runs" section in `docs/EXPERIMENTS.md`.

## Known fast-follows (non-blocking, tracked)

- **Stream C** (benchmark.py wrap + default-store-on + `run_methods` example) — deferred until
  eval-readiness merges its `slice_fn` seam.
- `DSPyJudge` async `acompletion` path not yet instrumented (sync `completion()` only).
- `git_sha`/`dataset_fingerprint` reachable via `langres.core.runs`, not re-exported from
  `langres.core` (polish).
- If a live tracker's `start_run` throws, the `"running"` line can dangle (the default `NoOpTracker`
  never triggers this).

Plan/contract: `docs/plans/20260708_tracking_observability_plan.md`.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main, feat/m4-foundation, feat/m5-foundation]
+    branches: [main, feat/m4-foundation, feat/m5-foundation, feat/tracking-observability]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ tmp/
 uv.lock
 .claude/*.bak
 wandb/
+mlflow.db
+mlruns/
 
 # macOS
 .DS_Store
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -53,6 +53,43 @@ Docs (`docs/GETTING_STARTED.md` + the doc-ladder rewire) are detailed under
 - Implemented core primitives (`Module`, `Blocker`, `Clusterer`) with Pydantic data contracts and 100% test coverage
 - Completed Approach 1 (classical baseline): `AllPairsBlocker` + `RapidfuzzModule` end-to-end pipeline
 
+### Experiment tracking & observability — run store, `ExperimentTracker`, LLM trace correlation
+
+The missing spine under langres's otherwise-**ephemeral** benchmark runs (rich results
+were printed, then lost — no run id, no config/data snapshot, no cross-run compare):
+content-addressed run identity, JSONL persistence, a pluggable tracker seam, and
+end-to-end trace correlation — **dependency-free** on the core path (`import langres`
+still pulls no `mlflow`/`wandb`).
+
+- **Run store + identity** (`langres.core.runs`, root-exported) — `RunContext` (the
+  recipe) + `RunRecord` (recipe + outcomes) with a content-addressed `recipe_id`
+  (`sha256` over config/data/seeds, *excluding* code/env provenance so a dirty tree or
+  `uv.lock` bump keeps the id) and an `attempt_id` PK; an `fcntl.flock`-guarded,
+  append-only JSONL `RunStore` (`read()` collapses `running`+terminal lines
+  last-wins-by-`attempt_id`); and `capture_run(context, *, store=None,
+  tracker=NoOpTracker())` — writes a `running` line at start and a terminal line on
+  exit, and sets the `current_run` contextvar. **`store=None` writes nothing.**
+- **`ExperimentTracker` Protocol + adapters** (`langres.core.trackers`) — an
+  Accelerate-style seam (`NoOpTracker` null default, `MultiTracker` fan-out to run
+  MLflow *and* W&B at once, `resolve_tracker` dispatch) with lazy **MLflow** and **W&B**
+  adapters behind the `[mlflow]` / `[wandb]` extras (a missing extra raises a helpful
+  `pip install 'langres[<backend>]'` `ImportError`). MLflow defaults to a local file
+  store out of the box; W&B supports keyless `offline`/`disabled` runs for CI/no-key use.
+- **LLM trace correlation** — `capture_run` sets `current_run`; `JudgementLog` records
+  the active `run_id`, and `LLMJudge` injects litellm `metadata` (`langres_attempt_id`
+  + pair ids + decision step) on **both** the sync (`forward`) and async
+  (`forward_async`) paths, so a Langfuse/OTel trace joins the `RunRecord` and
+  `JudgementLog`. Off a run (or a non-litellm client) the calls stay byte-identical
+  (no `metadata`).
+- **DSPy compile lineage** — `DSPyJudge.compile(...)` records the compilation as a
+  first-class optimization run via `capture_run` and stamps `_compile_run_id`, so a
+  later eval run threads it into `parent_run_id` (compile → eval lineage).
+- **`Settings`** — `RUN_STORE_PATH`, `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT` (the
+  MLflow ones consumed by `MlflowTracker`; a zero-config default `store` from
+  `RUN_STORE_PATH` is deferred to the benchmark wrap — pass `store=` explicitly today).
+  Docs: `docs/EXPERIMENTS.md`; runnable zero-spend
+  `examples/research/experiment_tracking_demo.py`.
+
 ### Flywheel closed loop — `docs/GETTING_STARTED.md` + doc-ladder rewire
 
 The one entry doc for the closed flywheel (fail-fast `auto`, `select_for_review`
--- a/docs/EXPERIMENTS.md
+++ b/docs/EXPERIMENTS.md
@@ -241,6 +241,19 @@ agents can write one file safely. Pass `tracker=` (an `ExperimentTracker`) to *a
 mirror params/metrics into MLflow or W&B; omit it for the JSONL-only path.
 (`git_sha()` and `dataset_fingerprint()` live in `langres.core.runs`.)
 
+**How the tracking `Settings` take effect (today).** `Settings` reads
+`RUN_STORE_PATH`, `MLFLOW_TRACKING_URI`, and `MLFLOW_EXPERIMENT`, but they are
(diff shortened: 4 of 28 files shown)


fxd24:
Thanks both — good catches. Addressed the warranted findings in three focused commits (`035a2da` runs, `cfb42f8` trackers, `b716358` correlation/docs), each with regression tests. Summary:

### Fixed (correctness — Claude bot)
- **H1 — `RunStore.append` flock released before the write reaches the OS.** Now `flush()` + `os.fsync()` *before* `LOCK_UN`, so bytes land under the lock. `read()` also hardened to tolerate a torn trailing line (skip at DEBUG) while WARNING-and-skipping a malformed *complete* line instead of raising. Tests spy the `LOCK_EX → fsync → LOCK_UN` syscall order + the torn/malformed read cases.
- **H2 — `tracker.start_run()` outside the try/finally.** Moved inside: a `start_run` failure now records `error_type`/`error_message`, resets the `current_run` contextvar, writes the terminal `status="failed"` record, and re-raises — no more orphaned `"running"` line. Regression test asserts the failed record + contextvar reset.
- **M3 — `MultiTracker` per-child isolation.** All six fan-out methods now guard each child: a raising backend is logged (via `logging`) and the remaining children still get the call, so one backend can't abort the others or mask the user's exception in `finish()`.
- **M4 — MLflow nested runs.** `start_run` now passes `nested=mlflow.active_run() is not None`; a real nested `capture_run` (e.g. DSPy `compile()` inside a benchmark run) no longer raises "Run already active". Tested against a real local file store.
- **M5 — WandbTracker global state.** `log_metrics`/`finish` now route through `self._run` (not module-global `wandb.log`/`wandb.finish`), so a child can't end the parent's run.
- **M6 — async correlation.** Extracted a shared gate (`current_run` set **and** client is litellm) used by *both* `forward()` and the async `acompletion` path — async judging inside `capture_run()` now stamps the same `metadata`. Byte-identical off a run. +3 async regression tests.

### Fixed (Codex)
- **P2 — MLflow bracketed param keys.** Flattening now emits dotted indices (`cascade_band.0`) and every key is sanitized to MLflow's charset. Verified end-to-end: a `cascade_band=(lo,hi)` context logs bracket-free params and `start_run` succeeds.
- **P2 — top-level `import fcntl` breaks Windows.** `runs.py` is on the bare `import langres` path, so `fcntl` is now imported lazily/guarded; on a platform without it, append degrades to no-lock with a one-time warning rather than failing the import.

### Fixed (nits)
- **M7** — settings docstrings + `docs/EXPERIMENTS.md` now state honestly how each setting takes effect (`MLFLOW_*`/`WANDB_*` via `resolve_tracker`; `RUN_STORE_PATH` auto-default deferred to the benchmark wrap — `capture_run` needs an explicit `store` today).
- **L2** — seed-label disambiguation now traverses in canonical (sorted) order, so key-permuted-but-identical configs get a stable `recipe_id`.
- **L3** — added a CHANGELOG entry for the subsystem. **L4** — `DSPyJudge.compile`'s `tracker` default is now `None` + `resolve_tracker(...)`.

### Not changed (with reason)
- **L1 (`_build_manifest` ordering)** — the reviewer notes it's harmless (each component's config is a pure function of its constructor args, not of `save_state`-mutated state), and `test_config_dict_writes_nothing_to_disk` already pins the no-I/O behavior. Left as-is to stay surgical.
- The two remaining known limitations (default-store-from-settings; a live tracker whose `start_run` fails inside a `MultiTracker`) fold into the deferred Stream C / are now mitigated by M3 isolation — noted in the PR body.

### Verification (beyond tests)
Merged tip: `ruff format --check` + `ruff check` clean; `mypy src` clean (86 files, `--all-extras`); full fast suite **2134 passed / 4 skipped**; and a real dual-backend e2e (`MultiTracker([mlflow, wandb])` on a local file store + offline wandb) exercising every fixed path — bracket-free MLflow params, a nested MLflow run that doesn't raise, both backends really writing, running→completed lifecycle, and **0** MultiTracker-swallowed child errors.

fxd24:
@codex review — re-review after the fix commits (035a2da runs, cfb42f8 trackers, b716358 correlation/docs) addressing all prior High/Medium/P2 findings.

fxd24:
Second Codex pass (on `b716358`) raised 3 more P2s — dispositioned as follows:

**Fixed (recipe-hash correctness):**
- **`config_dict()` leaked version into the recipe hash** → `4d35986`. `config_dict()` now returns only the reproducible `{"components": ...}` payload, dropping `artifact_version`/`langres_version` (those stay on `RunContext` as unhashed provenance, and `save()` still writes them to `resolver.json`). Regression test pins the invariant: same logical config → same `recipe_id` across a `langres.__version__`/`ARTIFACT_VERSION` bump; a second test asserts the on-disk manifest still carries both versions.
- **DSPy `compile()` didn't fingerprint its trainset** → `15c3d21`. `dataset_fingerprint` is now a canonical content hash of the trainset, so distinct trainsets get distinct `recipe_id`s (identical trainset still dedups). Regression test covers different-vs-identical trainsets.

**Deferred by decision — cost:**
- **Paid DSPy compile records no spend** (`spend_usd` stays `0.0`) — this is part of the broader cost-tracking story, and rather than bake in a quick fix we're designing cost properly (LLM + embedding + infra, decomposition, honest auto-capture) in **#100**. A `# see #100` marker sits where the compile-spend recording will go. Verified context: the LLM-cost seam is honest (`SpendMonitor`, pinned prices) but `record_cost` isn't auto-wired in production yet (that folds into the deferred Stream C benchmark wrap); embeddings are local/not-$-metered; infra is only `duration_seconds`.

Merged-tip verification: `ruff format --check`/`ruff check` clean, `mypy src` clean (86 files, `--all-extras`), full fast suite **2137 passed / 4 skipped**.

Issue #103, opened 2026-07-09
LLMJudge is unreachable by name from link()/dedupe()/Resolver.from_schema() — the verbs cannot express "use this prompt"


fxd24:
Found while replicating Peeters et al. on langres (#102).

## The gap

`judge="auto"` and `judge="zero_shot_llm"` both construct a **`DSPyJudge`** (all three judge-dispatch sites: `core/presets.py`, `methods.py`, `core/registry.py`). `LLMJudge` — the *only* judge class exposing a prompt-template seam (`system_prompt`, `response_parser`, `record_serializer`, `prompt_template`) — has **no preset name**.

Consequence: a user who wants to run a specific prompt (a published paper's, or their own domain prompt) cannot say so through `link()`, `dedupe()`, or `Resolver.from_schema()`. They must drop to `langres.core` and wire a `Module` by hand.

This is the single largest DX finding from the replication: the pipeline layer was reusable as-is; the judge layer was not reachable.

## Why it matters for the vision

The ROADMAP frames langres as the composable ER seam — *"compose, benchmark, and tune."* Zero-shot LLM matching with a **user-supplied prompt** is the most common thing a practitioner wants to try first, and it is the entry point for every LLM-ER paper published in the last two years. Today the declarative surface has no vocabulary for it, so such a matcher cannot be saved, loaded, or benchmarked via `Resolver` — only hand-assembled.

## Sketch (not a decision)

Add a preset that resolves to `LLMJudge` and lets the prompt through, e.g.

```python
langres.link(a, b, judge="prompt_llm", judge_kwargs={"prompt_template": ..., "response_parser": ...})
```

Open questions worth settling before coding:

- Does this become a new preset name, or does `zero_shot_llm` switch backing class (a behaviour change for existing callers)?
- `LLMJudge` is serializable via the config-registry — does the prompt template belong *in* the `Resolver` artifact? This interacts with the artifact-provenance question raised in the #100 design note (#101): an artifact that says what it *is* but not what it *achieves and costs* isn't choosable, and one that omits the prompt isn't reproducible either.
- All three judge-dispatch sites must stay in sync (see `.claude/rules/component-design.md`).

## Context

- Judge seams (`system_prompt`, `response_parser`, `record_serializer`, `on_parse_error`, and the `temperature=0.0` default) landed in #102.
- Related: #100 / #101 (cost + metric provenance on the artifact).

https://claude.ai/code/session_01KGKsfNuVYS826hQXpTuebu

PR #104, opened 2026-07-10
feat(eval): spend cap by default, honest best_threshold, ROC-AUC, public seams

fxd24:
## What this is

**PR-1 of three.** Groundwork for the judgement-contract change, split out because
it contains the only item that bills real money if wrong: **`evaluate()` had no spend cap.**

Nothing here touches `PairwiseJudgement`'s schema. `decision`/abstain (PR-2) and the
`EvalReport` tearsheet (PR-3) follow, with a $0.0158 logprob probe gating whether a
permanent `confidence` field is earned at all.

Motivating audit: `docs/research/20260710_dx_audit_examples_we_wish_we_had.md` (carried in
this branch, with a Status table marking G1/G2/G3/G4/G6 addressed and G5 open as #103).

## The headline

`evaluate()` now builds a `BudgetedModuleRunner` internally and caps spend **by default**
(`DEFAULT_BUDGET_USD = $1.00`). Previously a paid judge over a large candidate set billed
until it finished.

The cap is enforced **between calls** — a single in-flight call can overrun it by that
call's own cost. When that happens the run stops before starting another call, sets
`JudgePairEval.budget_exceeded`, and warns. It does **not** raise: the run completed, the
metrics are valid, and the money is already spent. Raising would only discard paid work.

## Bugs the review found in a fully-green branch

The branch passed 2508 tests, ruff, mypy, 98.79% coverage, and reproduced the pinned
Peeters F1 exactly — and still shipped a spend cap that could not detect being breached.

1. **The spend cap could not detect being breached.** The per-pair stop compared
   `total_spent + worst_case_price` *before* the next call, against the placeholder
   `_NEGLIGIBLE_WORST_CASE_PRICE = 1e-9` that `evaluate()` passes. Real cost was added
   *after* the call and never compared to the budget. Reproduced: one `$10.00` call under
   `budget_usd=1.00` returned `usd_total=10.0, truncated=False, truncation_reason="none"`,
   `n_judged=1/1`, **zero warnings**. Post-fix a `$2`/pair judge under a `$1` cap makes
   exactly one call and reports `budget_exceeded=True`.
   **Found by the Codex adversarial pass. The security specialist and the Claude
   adversarial subagent both read the same code and concluded the cap was "sound and
   bounded."**
2. **`cost_basis` disagreed with `usd_total` about whether money was spent.**
   `_judgement_cost()` sums `provenance["llm_cost_usd"]` (what `CascadeModule` writes), but
   the basis classifier only recognized `cost_usd` — so a real cascade run reported
   `usd_total=$0.004` alongside `cost_basis="none"`, `cost_is_real=False`.
3. **`make_token_cost_track` never set `cost_basis`/`usage`.** The second `CostTrack`
   producer (4 example call sites) returned real dollars labelled `"none"` with an all-zero
   token vector.
4. **`roc_auc_score` accepted non-finite scores** — a `NaN` returned `0.75` or `0.5` for the
   same multiset depending on input order, because `NaN` breaks both `sorted()` and the
   equality-based tie grouping.

Two more I found by running deliberately broken doubles rather than reading:

5. A judge failing on **100% of pairs** returned `f1=0.0, precision=0.0, recall=0.0` with only
   a warning — indistinguishable from a healthy judge that matched nothing. Now raises.
6. An **empty candidate list** silently returned `f1=0.0, truncated=False`. Now raises `ValueError`.

## Also in here

- **`evaluate()` warns that `best_threshold` is fitted to the gold it reports on.** The
  default still sweeps `DEFAULT_PAIR_GRID` and the number is unchanged — it just says once
  that it's an optimistically-biased argmax, not a held-out estimate. `threshold=<float>`
  grades honestly at a fixed cut. The default was deliberately *not* flipped to `0.5`: a
  global cut collapses an embedding judge from F1 1.000 → 0.667, because cosine non-matches
  sit at 0.70–0.80.
- **`CostTrack.cost_is_real` becomes a derived property**; the stored field is
  `cost_basis: Literal["real","estimated","mixed","untracked","none"]` — a bool cannot say
  "mixed". `CostTrack.usage` carries the summed token vector. *Tokens are the fact; dollars
  are derived.*
- **`DSPyJudge` abstained to the opposite verdict from `LLMJudge`** (`score=0.5`, no
  `parse_error` flag) — so at any threshold ≤ 0.5 a DSPy abstention was predicted a **match**,
  and `n_parse_errors` couldn't see it. Both judges now abstain at `0.0` with `parse_error=True`.
- `roc_auc_score` / `average_precision_score` — **pure Python**, tie-aware midranks (Mann-Whitney-U),
  zero numpy, zero sklearn. Cross-checked against sklearn in tests only, behind `importorskip`.
- `Resolver.candidates()` and `langres.eval.candidates_for()` — the seam the selling example needs.
- `src/langres/testing.py` with a public `ScriptedJudge`. Not `@register`-ed, not imported by
  `__init__.py`, locked by an import-budget subprocess test.
- **Docs de-fictionalized:** `langres.tasks`/`flows`/`ui`, `core.Optimizer`, `core.Evaluator`,
  `blockers.EmbedBlocker`, `data.SyntheticGenerator`, `pytrec_eval` — none of these exist.
  Removed from `TECHNICAL_OVERVIEW.md`. Also deleted README's false "`import langres` is heavy,
  eagerly pulls torch/litellm" bullet: measured 223ms with zero heavy modules in `sys.modules`.

## Corrections to my own claims

Two things I wrote and the maintainability specialist caught:

- The CHANGELOG claimed "no sklearn in `src/`". False — `calibration.py:20` and
  `random_forest_judge.py:30` import it behind `[trained]`. Reworded to "adding no numpy or
  sklearn dependency".
- The CHANGELOG and `testing.py` docstring claimed `ScriptedJudge` "replaces six hand-rolled
  copies across three test files." It replaced **zero**. `DummyBlocker` subclasses `Blocker`, so
  a `Module` double structurally cannot replace it. Fixed by actually migrating the one genuine
  duplicate (`test_cascade_judge.py`) and rewriting both claims to be checkable.

## Verification

- `pytest -m "not slow and not integration"` → **2508 passed, 4 skipped**, 98.79% coverage.
- `ruff format --check` (305 files), `ruff check`, `mypy src` (98 files) — all clean.
- `examples/quickstart_verbs.py` runs offline.
- `import langres` → 223ms, `heavy=[]`, `langres.testing` not in `sys.modules`.
- **$0 end-to-end regression:** `peeters_llm_em_replication.py --report-only --compare-archived`
  still reproduces F1 **92.09** (gpt-4o-mini) / **90.71** (gpt-4o) and 99.25% per-pair agreement,
  reading the committed v1 rows. Zero API calls.
- Every bug above was reproduced with a scripted double before the fix and re-run after.

## Not in this PR (deliberate)

`classify_pairs` was **not** taught to filter `parse_error` rows. `DEFAULT_PAIR_GRID` starts at
0.05 and an abstention scores 0.0, so the exclusion is a provable no-op — and it would teach the
generic pair primitive an LLM-specific provenance dialect. The semantic fix needs `decision=None`
and lands in PR-2.

Also out of scope: wiring a new judge name into the three dispatch sites (#103), promoting
Fellegi-Sunter/RF/DSPy to `confidence_source="calibrated"`, and the dead `[viz]` extra (flagged
with a TODO, not fixed).

https://claude.ai/code/session_01KGKsfNuVYS826hQXpTuebu


fxd24:
## Review response

Both reviewers found real bugs. Two fixed, one declined with evidence.

### ✅ Fixed — `evaluate(threshold=0.0)` grades an abstention as a match (claude, #1)

Confirmed by running it before touching anything:

```
threshold=0.0     tp=1 fp=0 fn=0   abstention predicted a match: True
threshold=-0.1    tp=1 fp=0 fn=0   abstention predicted a match: True
threshold=0.05    tp=0 fp=0 fn=1   abstention predicted a match: False
```

`classify_pairs` predicts a match iff `score >= cut`, and **both** `LLMJudge` and (after
this PR) `DSPyJudge` abstain at `score=0.0`. So a cut of `0.0` turns "I could not parse the
model's answer" into a confident YES and inflates recall off parse failures. `evaluate()`
did no validation of `threshold` at all — a parameter this PR itself introduced.

The reviewer suggested "reject `threshold <= 0` or document the caveat". I rejected, and
went one further: a cut above `1.0` is unreachable for a `ge=0.0, le=1.0` score, so F1 is a
structural `0.0` rather than a measurement. Both are degenerate cuts that yield a
plausible-looking number. `threshold` and every `grid` point must now lie in `(0.0, 1.0]`.
Guarding `threshold` but not `grid` would have been arbitrary — same predicate, same
invariant. Nothing in the repo passes a grid containing `0.0`.

The regression test asserts the **consequence** (`classify_pairs([abstain], gold, 0.0).tp == 1`),
not just that a `ValueError` fires — an assertion of the latter kind would have passed
against the broken code.

`caad9c1`, `f03d85a`

### ✅ Fixed — `candidates_for()` silently grades the train split on a typo (codex, P2)

Real. The expression was `(test_records, test_clusters) if split == "test" else (train_records, ...)`,
so `split="valid"` / `"validation"` / `"Test"` all fell through to **train** — a report that
looks valid while scoring the wrong partition. `Literal` only protects type-checked callers;
a CLI flag or a dict lookup reaches it untyped. Unknown splits now raise.

The test pins the consequence too: a companion assertion that `train_gold != test_gold`, so a
future silent fallback is a *detectable* wrong answer rather than a coincidence.

`f03d85a`

### ❌ Declined — `candidates_for()` comparator divergence (claude, #2)

> if a caller evaluates a comparator-reading judge (e.g. `WeightedAverageJudge`) via
> `candidates_for()`, the attached comparison vectors won't match what `run_methods`/the
> method registry would have produced for that benchmark (different weights/excludes).

I checked this rather than adding the suggested caveat, and **the divergence cannot occur**:

1. `candidates_for` builds `Resolver.from_schema(bench.schema)`, whose comparator is
   `Comparator.from_schema(schema)` (`resolver.py:394`).
2. The method registry derives its comparator the *same* way — `Comparator.from_schema(schema)`
   at `methods.py:147`, `:203`, `:257`, `:260`. No weights, no excludes, at any site.
3. I compared `feature_specs` for **every** registered benchmark. None exposes a comparator
   of its own at all:

   ```
   abt_buy / amazon_google / dblp_acm / dblp_scholar / febrl_person /
   fodors_zagat / tiny_fixture / walmart_amazon / wdc_computers
       -> no own comparator; candidates_for's derived one is the only one
   opensanctions -> external-only, not loadable
   ```

4. The one class that *does* carry a custom comparator, `FixedSplitPairBenchmark` (stored as
   `_comparator`, `exclude=`-derivable), has **no** `load`, `split`, `schema`, `build_blocker`,
   or `blocking_k` — it cannot satisfy `candidates_for`'s `Benchmark` + `BlockingBenchmark`
   contract and cannot be passed to it.

Adding a docstring caveat about a divergence that cannot happen would put fiction back into
the docs, which is the thing this PR exists to remove. If `FixedSplitPairBenchmark` ever grows
the blocking protocol, the caveat becomes true and earns its line then.

### Unverified claim I did not rely on

The `claude` review says "each with a reproduction and a test that fails red before the fix."
That is true of the six bugs in the description, but I want to be precise: I did not re-run
every one of those tests against the pre-fix commit for *this* response. The ones I did
re-verify from scratch here are the two above.

### Also in this push

While verifying the PR body's own claims against the tree, I found `core/presets.py` telling
users on the `judge="auto"` path that LLM spend is **"hard-capped"** — the exact overstatement
this PR corrects in `benchmark.py`. The `BudgetedModuleRunner` stops *between* calls, so one
in-flight call can overrun the cap by its own cost. Both user-facing messages now say what the
code actually does. No test asserted either string. `caad9c1`

### Verification after the fixes

- `2518 passed, 4 skipped`, coverage **98.79%**; `ruff format`/`ruff check`/`mypy src` clean.
- The `$0` end-to-end regression still reproduces the pinned numbers exactly, reading the
  committed v1 rows: F1 **92.09** (gpt-4o-mini) / **90.71** (gpt-4o), 99.25% per-pair agreement.
- A 38-assertion end-to-end run against the real bundled `fodors_zagat` benchmark (1023 blocked
  candidates, 33 gold pairs, zero API calls): `candidates_for` → `evaluate` → cap → AUC. The
  pure-Python `roc_auc_score`/`average_precision_score` match sklearn **exactly** (`0.825130` /
  `0.086651`) on those 1023 real scores, ties included. A `$2`/pair judge under a `$1` cap makes
  exactly **one** call and reports `budget_exceeded=True, truncation_reason="budget_stop"`.

fxd24:
### Follow-up: my own fix had a regression, caught by testing the guard against an input it didn't anticipate

The `grid` validation I added above (`for point in grid: _validate_cut(...)`) **consumed a
generator**. `grid` is typed `Sequence[float]`, but an untyped caller can pass a generator —
and before my guard, that worked. After it, the sweep saw an empty grid and died deep in the
harness with `max() iterable argument is empty`.

`evaluate()` now materialises `grid` once, validates the list, and passes the list down. An
empty grid is its own clear `ValueError` instead of a `max()` crash. This is precisely the
hazard `Resolver.candidates()` exists to close, one function over — I closed it there and
promptly reopened it here.

Two tests added: a generator grid behaves exactly like a list, and an empty grid raises.

`c32dca3` · suite now **2520 passed, 4 skipped**, coverage 98.79%, ruff + mypy clean.

fxd24:
@codex review

Please re-review the four commits pushed since your first pass — they add input validation that did not exist when you reviewed:

- `caad9c1` — presets.py wording (no logic change)
- `f03d85a` — `_validate_cut()` in benchmark.py; `split` validation in eval.py
- `c32dca3` — materialise `grid` before validating it

Specifically worth attacking: does `_validate_cut` admit any value that still lets an abstention (`score=0.0`) be graded a match, and does materialising `grid` change behaviour for any existing caller?

fxd24:
### Self-attack on the two new guards (so the answer doesn't depend on a reviewer noticing)

I asked Codex two questions above. I also answered them myself, by running the code rather
than reading it.

**Q1 — does any cut `_validate_cut` admits still let an abstention (`score=0.0`) be graded a match?**
No. Swept the admitted range at 1e-4 resolution plus the boundaries and denormals
(`5e-324`, `2.2e-308`, `nextafter(1.0, 0.0)`, `1.0`):

```
10005 cuts admitted; abstention graded a match by 0
every degenerate cut rejected: True      # 0.0, -0.0, -1e-300, -0.5, 1.0000001, 1.5, ±inf, nan
DEFAULT_PAIR_GRID (19 pts, min=0.05) passes its own validator
```

`-0.0` is worth calling out: it compares equal to `0.0`, so `0.0 < -0.0` is `False` and it is
rejected — the guard does not have a signed-zero hole.

**Q2 — does materialising `grid` change behaviour for any existing caller?**
No. Same `(best_threshold, f1, pr_curve_points)` for every sequence kind, and the curve is not
truncated by the second downstream pass:

```
list  -> (0.3, 1.0, 3)
tuple -> (0.3, 1.0, 3)
gen   -> (0.3, 1.0, 3)
```

Before `c32dca3` the generator row was a `max() iterable argument is empty` crash.

fxd24:
### Codex's second finding is right, and it exposed a false claim of mine

> When callers use `evaluate_judge_on_candidates` directly … `grid` is still passed to
> `pair_pr_curve` without `_validate_cut`.

Correct. That is the path documented for paid/compiled judges and the one the research
examples call — the invariant `evaluate()` enforced did not hold for the API that actually
spends money. Fixed in `3ed526a`, and validated **before** the judge runs, so a bad grid never
costs an API call. The test asserts that with a `ScriptedJudge` spy: `judge.seen == []`.

**Correction to my earlier comment and commit message.** I wrote *"Nothing in the repo passes a
grid containing 0.0."* That is **false**. `tests/data/test_wdc_computers.py` computes

```python
grid = tuple(round(0.05 * i, 2) for i in range(21))   # 0.00 .. 1.00
```

My grep was for `grid=[0.0` and could not see a computed grid. The full suite caught it — the
ban I'd added broke that test.

### The claim was wrong because the rule was wrong

Being forced to look at a real caller showed the guard was over-broad. A **fixed grading
threshold** and a **swept curve point** are different cuts and deserve different rules:

| cut | what it is | rule |
|---|---|---|
| `threshold` | graded at, reported, carried into production | `(0.0, 1.0]` — `0.0` calls every pair a match, abstentions included |
| `grid` point | a coordinate on a PR curve | `[0.0, 1.0]` — `0.0` is the curve's legitimate predict-all anchor (recall `1.0`, precision = prevalence) |

Banning `0.0` from the sweep would outlaw an honest ranking-judge PR curve in order to defend
against an *abstaining* judge's convention. That is the same layering mistake the plan's **E1**
rejected when it refused to teach `classify_pairs` an LLM-specific `parse_error` dialect — and I
walked straight into it one function over.

So `_validate_cut` (strict) is now split from `_validate_grid_point` (loose), and the real signal
is surfaced where it belongs: **`evaluate()` warns when the argmax lands on `0.0`.**

```
evaluate(): the best-F1 threshold is 0.0 -- the predict-all point, where every
candidate (including an abstention at score=0.0) counts as a match and recall is
trivially 1.0. This judge does not beat predicting every pair a match. …
```

That catches the case you were actually pointing at (`grid=(0.0,)` producing a flattering
number) without lying about what a PR curve is.

Also in `3ed526a`: an empty grid raises instead of dying in `max() iterable argument is empty`,
and `run_method()` holds a dataset-supplied `threshold_grid` to the same rule.

### Verification
`2525 passed, 4 skipped`, coverage 98.79%; ruff + mypy clean. The `$0` Peeters regression is
unchanged: F1 **92.09** / **90.71**, 99.25% per-pair agreement. The 38-assertion end-to-end run
on `fodors_zagat` still passes.

PR #106, opened 2026-07-10
PR-2: the judgement contract — a judge decides, optionally says how sure

fxd24:
## The judgement contract: a judge says what it *decided*, optionally how *sure* it is

`PairwiseJudgement.score: float` conflated two different things — the **decision**
("is this a match") and the **confidence** ("how sure am I"). A binary LLM judge has
no score; it says Yes or No. To satisfy the old contract it fabricated `0.0`/`1.0`
tagged `score_type="prob_llm"` — a decision wearing a probability's clothes. Three
real bugs followed, all fixed here:

1. **The flywheel silently no-op'd.** `select_for_review(strategy="uncertainty")`
   ranks by `|score − threshold|`. For a binary judge that distance is *always* `0.5`,
   so the uncertainty band was always empty and the loop reported **exhausted** having
   never started.
2. **An abstention was graded as a confident NO.** A parse failure emitted `score=0.0`,
   which `classify_pairs` counted as a non-match, silently depressing recall.
3. **Two judges abstained to opposite verdicts.** `LLMJudge` abstained at `0.0` (a NO);
   `DSPyJudge` abstained at `0.5` — predicted a **MATCH** at threshold 0.5.

## The split

`PairwiseJudgement` now carries, all additive/defaulted:

- `decision: bool | None` — a decider's verdict (`None` == abstain)
- `score: float | None` — **widened** from a required `float`; a ranker's score
- `confidence: float | None` + `confidence_source: Literal["none","unrequested","logprob","calibrated","heuristic"]`

One seam decides everything: **`predicted_match(judgement, threshold) -> bool | None`**
— a decider's `decision` wins; else `score >= threshold`; neither = abstain (`None`,
*not* a "no"). Every consumer routes through it. `classify_pairs` excludes an abstain
from the predicted set — so an abstained gold pair is still a false negative (recall is
**not** flattered; the win is honesty, not a metric bump).

## What each judge now does

- A **binary `LLMJudge`** emits `decision=True/False, score=None`.
- With `confidence="logprob"` it promotes `score=p_yes` (an honest `prob_llm` ranking
  signal), `confidence=max(p_yes, 1−p_yes)` (credence in its *own* answer),
  `confidence_source="logprob"`. `confidence` is now in `config`/`from_config`, so
  `Resolver.save`/`load` no longer silently drops it (PR #105 review note).
- An **abstain** emits `decision=None, score=None` and routes `on_parse_error`.
  `DSPyJudge` now abstains the same way (`provenance["parse_error"]=True`), so
  `n_abstained` sees it and it is never a silent match.

## Confidence is *earned*, on measured evidence

The permanent `confidence` field was gated on the logprob probe (merged PR #105,
`$0.0158`, user-authorized): on all 1206 Abt-Buy pairs, `roc_auc(correct, credence)
= 0.9500` — reviewing the least-confident 5% catches 60% of the judge's errors (12.1×
lift). Had it come back ≈ 0.5, this PR would have shipped `decision` + abstain alone.

**Honest scope:** confidence via logprobs is free *in output tokens, at `explain=False`,
on a logprob-returning model* — never "free" unqualified. Logprobs are an OpenAI-family
feature; **unverified** for the GLM/DeepSeek/Qwen judges our own paid experiments use —
which is exactly why `confidence_source` separates `"none"` (structurally none) from
`"unrequested"` (could, you didn't ask).

## The flywheel now runs

- `select_for_review("uncertainty")` ranks by `confidence` when present, and **raises**
  (naming `strategy="disagreement"` and `LLMJudge(confidence="logprob")`) when no
  rankable signal exists — instead of returning `[]`, which it documents as "exhausted".
- `ReviewItem` gained `reasoning`/`confidence`/`confidence_source`; the CLI review loop
  renders a decider's credence instead of crashing on its `None` score.
- `JudgementLog` is **v3**: rows carry `decision`/`confidence`/`confidence_source`;
  `read()` backfills `decision` from `verdict` for v1/v2 rows. Also fixes a real cost
  bug — `append()` read only `cost_usd`, so cascade rows (which write `llm_cost_usd`)
  persisted `0.0`.
- `link()` raises `JudgeAbstainedError` (a `RuntimeError` subclass, root-exported) when
  the judge abstains — a `match=None` verdict would be read as "no match" by the obvious
  `if verdict.match:`, recreating the very bug the split removes.

## Verification

- **2621 tests pass**, coverage 98.73%; `ruff`, `ruff format`, `mypy --strict` all clean.
- **$0 Peeters replication still pins exactly** after the contract change: F1 **92.09**
  (gpt-4o-mini) / **90.71** (gpt-4o), 99.25% per-pair agreement — reading committed v1
  rows through the new contract. `_metrics_from_rows` grades off the `verdict` column
  (→ decision), never the now-`None` score.
- **The logprob analysis still reproduces** `roc_auc(correct, credence) = 0.9500` at $0.
- A new **cross-slice integration test** (`tests/test_flywheel_integration.py`) proves
  the loop the four independently-built slices compose into: decision → `JudgementLog`
  v3 → `read()` → `select_for_review`, ranked by credence; the no-signal raise; the
  disagreement fallback; the cost-key fix.

## Not in scope (deliberately)

- The `EvalReport` tearsheet / `to_html()` — **PR-3**.
- Promoting Fellegi-Sunter/RF/DSPy to `confidence_source="calibrated"` — left `"none"`
  (the honest default).
- Wiring a new judge name into the three dispatch sites (#103).

🤖 Built as Wave 1 (contract + `score` widening) → Wave 2 (four parallel agents:
llm_judge, judgement_log/harvest/labelers, review, benchmark/dspy), each in an isolated
worktree, integrated and gated one slice at a time.

https://claude.ai/code/session_01KGKsfNuVYS826hQXpTuebu

### DIFF
--- a/.agents/skills/prompting-claude-4/SKILL.md
+++ b/.agents/skills/prompting-claude-4/SKILL.md
@@ -1,0 +1,202 @@
+---
+name: prompting-Codex-4
+description: Expert guidance for prompting Codex 4.x models (Opus, Sonnet, Haiku tiers). Use when writing system prompts, optimizing model behavior, fixing behavioral issues like overtriggering or over-engineering, or migrating from older Codex models. Includes XML patterns and behavioral fixes.
+---
+
+# Prompting Codex 4.x Models
+
+This skill provides expert guidance for writing effective prompts for Codex 4.x
+models across the Opus, Sonnet, and Haiku tiers. It stays version-agnostic: the
+tiers trade intelligence for latency and cost — pick per task, and consult the
+live Anthropic model docs for current IDs and benchmarks.
+
+## Core Principles
+
+### 1. Be Explicit with Instructions
+
+Codex 4.x models are trained for **precise instruction following**. They take instructions literally and do exactly what you ask—nothing more.
+
+**Less effective:**
+
+```
+Create an analytics dashboard
+```
+
+**More effective:**
+
+```
+Create an analytics dashboard. Include as many relevant features and interactions as possible. Go beyond the basics to create a fully-featured implementation.
+```
+
+### 2. Add Context for Why
+
+Explaining motivation helps Codex understand your goals:
+
+**Less effective:**
+
+```
+NEVER use ellipses
+```
+
+**More effective:**
+
+```
+Your response will be read aloud by a text-to-speech engine, so never use ellipses since the text-to-speech engine will not know how to pronounce them.
+```
+
+### 3. Tell Codex What TO Do (Not What NOT to Do)
+
+Format guidance is more effective when positive:
+
+- Instead of: "Do not use markdown in your response"
+- Try: "Write in smoothly flowing prose paragraphs."
+
+### 4. Use XML Tags for Structure
+
+Codex is trained to recognize XML tags for organizing prompts:
+
+```xml
+<behavior_instructions>
+Your tone should be professional and direct.
+</behavior_instructions>
+
+<coding_guidelines>
+Follow existing patterns in the codebase.
+Always read files before editing.
+</coding_guidelines>
+```
+
+## Behavioral Tendencies to Address
+
+Codex 4.x models share tendencies worth steering with explicit prompt guidance.
+See [prompt-snippets.md](prompt-snippets.md) for copy-paste fixes.
+
+- **Over-engineering**: add explicit guidance to keep solutions minimal.
+- **Tool overtriggering**: use moderate language instead of aggressive directives.
+- **Conservative code exploration**: explicitly instruct to read files first.
+- **"Think" sensitivity**: when extended thinking is disabled, replace "think"
+  with "consider/evaluate/believe".
+- **Aggressive parallel tool calling**: powerful, but can bottleneck a system —
+  scope it when concurrency is costly (see the parallel-tool-calls pattern below).
+- **Concise-by-default communication**: models may skip post-tool summaries; ask
+  explicitly when you want them (see Communication Style Adjustments below).
+
+Higher tiers (Opus) also expose finer output-effort control and preserve thinking
+blocks across turns; consult the live model docs for the current per-model feature
+matrix rather than pinning it here.
+
+## Essential Prompt Patterns
+
+### Parallel Tool Calling
+
+Boost parallel tool usage to ~100%:
+
+```xml
+<use_parallel_tool_calls>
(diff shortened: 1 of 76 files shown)


fxd24:
## Addressed the bot reviews (claude-review + Codex) — commit `66aca59`

Cross-referenced both automated reviews against the current source. Four of the findings they raised were **already fixed** by the earlier adversarial-review pass (`83edb36`, which landed after the bots reviewed `8fb734e`): the CascadeJudge escalate-every-decider bug, the CLI `verdict=None → NO-MATCH` render, the empty-score `inspect_scores`/`np.mean` crash (Codex `reports.py:1014`), and the `_map_verdict` score-clobber. Verified each guard is present at HEAD.

Three were genuinely open; all now fixed with tests:

| # | Finding | Fix |
|---|---------|-----|
| Codex **P1** `harvest.py:225` | `harvest_labeled_pairs` coerced an abstention row (`verdict=None`) to a `False` non-match — the label-side twin of coercing a null score to `0.0`, the exact "abstain read as NO" bug this contract exists to kill, relocated to silver-label harvesting. | Skip the row (no usable label) unless a human correction supplies one. +2 tests (skip; correction-rescues). |
| claude **bug 3** `review.py:466` | `_select_uncertainty` returned **only** confidence-bearing rows the instant any row had a confidence, silently dropping score-only uncertain rows in a mixed CascadeJudge log. | New `_uncertainty_by_score` helper folds the score band back in (credence-ranked first, then score-ranked). Flagship raise on a pure no-signal decider log preserved. +1 mixed-log test. |
| claude **bug 4** `verbs.py dedupe()` | The abstain asymmetry with `link()` was undocumented. | Docstring note: `link()` raises `JudgeAbstainedError` (one caller needs a verdict); `dedupe()` leaves an abstained pair **unmerged** (conservative) rather than aborting a batch on one unparseable judgement. +1 behavior test. |

**Declined — Codex P2** (`review.py:480`, "uncertainty should exhaust, not raise, on an all-0/1 score log"). This conflicts with a deliberate, tested flagship decision: an all-`0/1` score log is the binary-decider fingerprint, and `test_uncertainty_on_logged_binary_judge_raises_end_to_end` asserts it **raises**. A continuous scoring judge is indistinguishable from a fabricated-`0/1` decider at the row level, so raising loud beats risking the silent no-op the whole PR exists to remove.

The bots' narrower test-coverage notes (base `Clusterer.cluster` abstain path, `LoggingModule` verdict=None) are noted; the two substantive new fixes are covered.

Local verification: **2631 pass**, coverage **98.71%**, ruff/format/mypy clean, and the `$0` Peeters oracle still pins **F1 92.09 / 90.71 at 99.25% agreement** off the committed v1 rows.

Issue #145, opened 2026-07-15
Epic: Autoresearch loop for entity resolution — prove the loop closes on quick wins


fxd24:
# Epic: Autoresearch loop for entity resolution — prove the loop closes on quick wins

## Thesis

Apply the Karpathy "autoresearch" loop — an agent that iterates **propose → run →
evaluate → keep-if-better-else-revert** against a fixed objective — to entity
resolution, with **langres as both the substrate and the measuring stick**.

The magic in that loop is not the agent; it's the **objective**. Validation loss
works because it is cheap, trustworthy, sensitive, low-variance, and cheaply
reversible — you watch it move in small steps and *know* the direction is right.
ER hands us the objective (55 years of blocking/matching/clustering metrics), but
those metrics **saturate and are lumpy** — at the 98–99% F1 we already hit, F1
chases label noise, not real gains. So the core move is to steer on a **loss-like**
objective (continuous: AUC-PR, log-loss, recall@budget) and let thresholded F1 be
the report card, not the steering wheel.

**We have not proven anything yet.** This epic's job is to prove the loop closes on
the cheapest, most deterministic experiments with real headroom — **before** we
commit to any expensive or combined training.

## Why now / relationship to the other epics

**This is a standalone epic.** It is about *running the autoresearch loop* — the
generic harness. **#85 is about replicating the research itself** — producing
credible ER numbers vs Ditto/Jellyfish/AnyMatch. The two connect (a working loop can
eventually *drive* those replications), **but they are not the same, and this loop's
building blocks must exist first.** So we build the harness here, independently, and
only point it at #85's payloads once it's proven.

- **#125 (training surface) shipped** (PR #140, 2026-07-15): langres is now
  *trainable* (`fit(method=QLoRA|MIPRO|GEPA|Platt)`, `finetune()`, `Calibrator`,
  `align_pairs`), not just measurable. The loop finally has trainable knobs to turn.
- **#85 (ER method-replication & small-LM fine-tuning)** — the *payloads* this loop
  will drive later; reused, not re-derived.
- **#55 (seam-audit deltas)** — the standing method backlog; several C-tier items
  are cheap objectives this loop can optimize.

## Where we sit on the programmatic↔agentic spectrum

Every real autoresearch system (Karpathy's AutoResearch, AIDE, Optuna) keeps
**Run / Evaluate / Decide / Persist programmatic and the Objective immutable** —
Karpathy's scorer (`prepare.py`) is off-limits so the agent physically cannot game
[...]

PR #155, opened 2026-07-15
feat(autoresearch): declarative SearchSpace + config→blocker factory

fxd24:
## What

The **proposal substrate** for the autoresearch loop (`propose → run → evaluate → keep-if-better`, epic #145 M1, **PR P-B**): a declarative `SearchSpace` the loop enumerates, and a pure `factory` that turns one config dict into a runnable blocker. Builds on the sibling `Objective` (#154); the loop + `langres.optimize` facade land in a later PR.

## `core/autoresearch/search_space.py` — import-light

Frozen dataclass; each field is the candidate-value tuple for one parameter:

```python
SearchSpace(
    blocker=("vector",),
    embedding_model=("all-MiniLM-L6-v2",),
    metric=("cosine",),
    text_field=("name",),
    k_neighbors=(5, 10, 20),
)
```

- `configs() -> Iterator[dict[str, Any]]` yields the Cartesian product as config dicts.
- **Ordering contract (load-bearing):** `k_neighbors` is the **innermost** varying dimension, so consecutive configs hold `(blocker, embedding_model, metric, text_field)` fixed while `k` sweeps its full range. This lets the downstream loop build **one** vector index per `(model, metric, text_field)` and reuse it across every `k` (`k` lives on the blocker, not the index) instead of re-embedding per `k`.
- `__len__` sizes the grid; `__post_init__` fails loud on an empty axis.
- **Pure stdlib + typing** (itertools/dataclasses) — no faiss/torch/sentence-transformers, so it can sit on the public API surface (users build a `SearchSpace` to call `langres.optimize`) without breaking the `import langres` budget. It never imports `factory`.

## `core/autoresearch/factory.py` — lazy leaf

Two pure functions, no global mutable state (index reuse is the caller's job via the k-innermost ordering + passing `index=`, not a cache here):

- `build_index(embedding_model, metric, texts, *, embedder=None) -> VectorIndex` — mirrors the **canonical** VectorBlocker build path (`SentenceTransformerEmbedder` → `FAISSIndex(embedder=, metric=)` → `create_index`) used in `core/presets.py:_build_vector_blocker` and `data/_benchmark_utils.py:sweep_blocking_k`. The optional `embedder=` seam lets tests inject `FakeEmbedder` (no model download).
- `build_blocker_from_config(config, *, schema, index=None) -> Blocker` — dispatches on `config["blocker"]`: `"vector"` requires a prebuilt `index` (clear `ValueError` if `None`) and builds `VectorBlocker(vector_index=index, schema=schema, text_field=…, k_neighbors=…)`; `"all_pairs"` builds `AllPairsBlocker` (index unused). Named `build_blocker_from_config` to avoid colliding with `Benchmark.build_blocker`.
- **Not import-light** by design — imports the [semantic] stack at module top, so it must only ever be imported lazily (never reached by a bare `import langres`, since the package `__init__` exports nothing).

## Scope

Adds only `search_space.py`, `factory.py`, and their tests. No `__init__.py` / `pyproject.toml` / `test_import_budget.py` edits.

## Verification

`ruff check .`, `ruff format --check .`, `mypy src` (clean), 21 fast + 1 `@slow` new tests pass, `tests/test_import_budget.py` still green. Both new modules are covered 100% by construction (fast tests + the slow real-embedder smoke covers the non-injected branch; the coverage gate runs on the Linux `test-full` job).

## Summary by Sourcery

Introduce a declarative SearchSpace for autoresearch configurations and a factory to build blockers and vector indexes from those configs.

New Features:
- Add a frozen, import-light SearchSpace dataclass that enumerates Cartesian products of blocking configurations with a k-neighbors-innermost ordering contract.
- Provide a factory module with build_index and build_blocker_from_config helpers to construct FAISS-based vector indexes and corresponding Blocker instances from config dicts.

Tests:
- Add high-coverage tests validating SearchSpace defaults, Cartesian product sizing, k-neighbors ordering, empty-axis validation, and import-lightness.
- Add tests for the factory module covering index construction, metric propagation, vector and all-pairs blocker creation, error handling, and a slow smoke test for the real SentenceTransformerEmbedder path.

### DIFF
--- a/src/langres/core/autoresearch/factory.py
+++ b/src/langres/core/autoresearch/factory.py
@@ -1,0 +1,118 @@
+"""Turn one autoresearch config dict into a runnable blocker.
+
+Companion to ``search_space``: :class:`SearchSpace` enumerates config dicts and
+this module builds the blocker each one describes. Two pure functions, no global
+mutable state — index reuse across ``k`` values is the *caller's* job (it builds
+one index and threads it into :func:`build_blocker_from_config`; see the
+k-innermost ordering contract on ``SearchSpace.configs``), not a cache here.
+
+**NOT import-light.** This module imports the [semantic] stack
+(faiss/sentence-transformers, via the vector index + embedder) at module top, so
+it must only ever be imported **lazily** — on demand when a search actually runs,
+never from a bare ``import langres`` (unlike the sibling ``search_space``, which
+stays pure so it can live on the public API surface). The autoresearch package
+``__init__`` intentionally exports nothing, so importing ``langres`` does not
+reach this module.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from typing import Any, Literal, cast
+
+from langres.core.blocker import Blocker
+from langres.core.blockers.all_pairs import AllPairsBlocker
+from langres.core.blockers.vector import VectorBlocker
+from langres.core.embeddings import EmbeddingProvider, SentenceTransformerEmbedder
+from langres.core.indexes.vector_index import FAISSIndex, VectorIndex
+
+
+def build_index(
+    embedding_model: str,
+    metric: str,
+    texts: Sequence[str],
+    *,
+    embedder: EmbeddingProvider | None = None,
+) -> VectorIndex:
+    """Build a FAISS index over ``texts``, ready for search.
+
+    Mirrors the canonical VectorBlocker build path used across the codebase
+    (embedder → ``FAISSIndex(embedder=..., metric=...)`` → ``create_index``;
+    see ``core/presets.py:_build_vector_blocker`` and
+    ``data/_benchmark_utils.py:sweep_blocking_k``). ``create_index`` is called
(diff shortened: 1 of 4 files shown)


fxd24:
Thanks @sourcery-ai. Addressed the two points:

- **`build_index` metric validation** (fixed in d56549f): now rejects an unknown metric with a clear, local `ValueError` before any embedding work, instead of deferring to a FAISS-internal error — the metric comes from a user's `SearchSpace.metric`, so failing fast at the boundary is friendlier. The `str -> Literal` cast is now justified by the preceding guard; added a covering test.
- **Iterating `dataclasses.fields()` in `configs()`/`__len__`**: kept explicit by design. The `itertools.product(...)` positional call pins the load-bearing "k_neighbors is the innermost dimension" ordering contract visibly at the call site; a generic `fields()` iteration would make that invariant depend implicitly on field-declaration order, which is easy to break silently when adding an axis. `__len__` stays symmetric with it.

PR #165, opened 2026-07-16
fix(data): harden failure-mode diagnostic + denoise feature_specs parity + honest framing

fxd24:
Six verified findings from a cross-model code review (Codex gpt-5.5 + a fresh Claude adversarial pass) of the langres data-preparation layer. Each fix is surgical and traces to a reproduced finding. The recipe-lift proof (`examples/recipe_lift_proof.py`) itself was verified **valid** by both reviewers and is **unchanged** — it still prints recipe 0.5833 / baseline 0.2043 / margin +0.3790, beats 10/10 draws.

## Fixes

**A — record-key str-normalization (`failure_mode.py`)** — the join stringifies judgement ids, but the `records` map was looked up with those stringified ids against the caller's raw-keyed map. An int-keyed map missed every lookup silently: field-emptiness read all-empty and source slices were skipped, corrupting the headline diagnostic (reproduced: `empty:name` reported n=2 lift=1.0 vs the str-keyed truth n=1 lift=2.0). Fix: normalize the records keys to `str` once at entry (`_normalize_records`, symmetric with `_normalize_gold`).

**B — score-band lower clamp (`failure_mode.py`)** — `min(int(j.score * n_bands), n_bands - 1)` clamped only the upper bound, so `score < 0` produced `buckets[-1]` → KeyError crash (reproduced). Meanwhile the overlay's `_histogram` silently drops out-of-`[0,1]` scores. Fix: a shared `_in_unit_range` predicate excludes out-of-range (and `None`) scores from **both** the slices and the overlay, keeping the two views consistent and impossible to crash.

**C — malformed-gold logging (`failure_mode.py`)** — `_normalize_gold` silently dropped gold pairs without exactly two distinct ids. This section's contract is "omissions logged, never silent." Fix: count drops and `logger.warning` the count when > 0.

**D — `feature_specs` on `denoise_pairs` (`mining.py`)** — `mine_misclassified_pairs` accepts `feature_specs` but `denoise_pairs` hardcoded the full default, so a matcher trained on a feature SUBSET was denoised in a different space. Fix: add `feature_specs` (mirroring `mine_misclassified_pairs`) and thread it through `_resolve_feature_specs`.

**E — denoise-before-mine caveat (doc)** — confident learning cannot perfectly separate label noise from genuine hard positives, so denoise-before-mine competes with hard-positive mining. Added an honest caveat to `denoise_pairs`'s docstring and a one-line comment in the recipe's `assemble_recipe` (which accepts the tension because class balance dominates the lift). No behavior/order change.

**F — honest Cleanlab framing (doc)** — the denoiser is the confident-JOINT core (Northcutt et al.), not Cleanlab's full prune pipeline. Reworded the test cross-check docstring to "confident-joint core, partial overlap (jaccard ≥ 0.5), not byte-parity." Also tightened the example's "equal label budget" → "equal training budget" (both arms sit on the same fully-labeled pool; they differ in selection at equal training size N).

## Gates
- TDD: tests written first (red) for A/B/C/D in `test_failure_mode.py` + `test_mining.py`, then implemented.
- Coverage: `failure_mode.py` **100%** (225 stmts / 60 branches), `mining.py` **100%** (121 stmts / 38 branches).
- `tests/data/` + `tests/test_import_budget.py`: 664 passed, import-light budget green (sklearn stays lazy).
- `ruff format` + `ruff check` clean; `mypy src/` clean on touched files.
- `examples/recipe_lift_proof.py` prints the same numbers — proof unchanged.

## Summary by Sourcery

Harden failure-mode diagnostics and denoising behavior in the data-preparation layer, and align documentation and tests with the intended semantics.

Bug Fixes:
- Normalize record keys to strings in failure-mode profiling so field-emptiness and source slices work correctly with non-string ids.
- Exclude out-of-[0,1] and missing scores from score-band slices and overlays to prevent crashes and keep the two views consistent.
- Log and count malformed gold pairs that are dropped during normalization instead of silently omitting them.
- Ensure denoise_pairs respects the caller-provided feature_specs so denoising happens in the same feature space as training.

Enhancements:
- Add a shared score-range predicate to unify score handling between overlays and score-band slices.
- Improve docstrings and comments around denoise_pairs, Cleanlab cross-checking, and the recipe proof to more accurately describe budgets, caveats, and method scope.

Documentation:
- Clarify denoise_pairs caveats about trading off label denoising against hard-positive mining.
- Reframe Cleanlab comparison docs to emphasize confident-joint core overlap rather than exact parity.
- Tighten recipe_lift_proof wording to describe an equal training budget drawn from a fully labeled pool and note the denoise-before-mine tradeoff.

Tests:
- Add targeted tests for record-key normalization, score-band handling of out-of-range scores, malformed-gold logging, and denoise_pairs feature_specs propagation.

### DIFF
--- a/examples/recipe_lift_proof.py
+++ b/examples/recipe_lift_proof.py
@@ -1,4 +1,4 @@
-"""Proof: a curated training recipe beats random pairs at an equal label budget.
+"""Proof: a curated training recipe beats random pairs at an equal training budget.
 
 This is the acceptance demonstration for the data-preparation layer. It asks one
 question and answers it with numbers, on a real benchmark, for $0:
@@ -6,6 +6,10 @@ question and answers it with numbers, on a real benchmark, for $0:
     Given a fixed budget of N training pairs drawn from a blocked candidate pool,
     does *curating* those N (balance + hard positives + denoise + augmentation)
     beat drawing them at random?
+
+Both arms draw the same N pairs from the *same fully-labeled pool*: the budget is
+training size N, not a labeling budget, and the arms differ only in how those N are
+*selected*.
 
 Why it works -- honestly. A blocked entity-resolution pool is dominated by
 non-matches (here an all-pairs blocked pool runs ~1:500 positive:negative). A
@@ -104,6 +108,8 @@ def assemble_recipe(
     Returns ``(recipe, n_flagged)`` -- the assembled pairs and how many the
     denoiser dropped.
     """
+    # Denoise-before-mine can strip genuine hard positives (denoise and hard-positive
+    # mining compete); the recipe accepts that here because the class-balance lift dominates.
     clean, flagged = denoise_pairs(train_pool, seed=SEED)
     hard = mine_misclassified_pairs(clean, cap=POS_CAP, seed=SEED)
     augmented = [
@@ -202,7 +208,7 @@ def main() -> None:
 
     verdict = "LIFT" if margin > 0 else "NO LIFT"
     print(
-        f"\n=> {verdict}: at an equal {budget}-pair budget, curation "
+        f"\n=> {verdict}: at an equal {budget}-pair training budget, curation "
         f"{'beats' if margin > 0 else 'does not beat'} random by {margin:+.4f} F1."
     )
     print(
--- a/src/langres/data/data_profile/failure_mode.py
+++ b/src/langres/data/data_profile/failure_mode.py
@@ -360,10 +360,14 @@ def profile_failure_mode(
 
     score_edges, error_counts, success_counts = _score_overlay(confident, n_score_bands)
 
+    # Normalize record keys to str ONCE here (symmetric with _normalize_gold), so
+    # the slice lookups -- keyed by the stringified judgement ids -- hit even when
+    # the caller passed an int-keyed map.
+    slice_records = _normalize_records(records) if records is not None else None
     all_slices = _build_slices(
         confident,
         overall_rate,
-        records=records,
+        records=slice_records,
         id_key=id_key,
         source_key=source_key,
         n_score_bands=n_score_bands,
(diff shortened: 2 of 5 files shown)


fxd24:
Thanks @sourcery-ai. Dispositions on the three findings:

**1. `str(key)` collision (e.g. `1` vs `"1"`) — intentional, not fixed.** The entire failure-mode diagnostic is str-keyed *by design*: `_join_gold` and `_normalize_gold` already stringify every judgement/gold id via `str()`, and the slice lookups key on those stringified ids. `_normalize_records` is deliberately symmetric with that established behavior — a records map is expected to share the id space (`str(id)`), so there is no collision to guard against. Adding collision-detection would be speculative error-handling for a case that can't arise in this pipeline (Sourcery's `1` vs `True` example actually stringifies to distinct `"1"`/`"True"`).

**2. `_normalize_records` return type — documented (commit 4e45099).** The out-keys are indeed always `str`, but narrowing the return to `dict[str, ...]` breaks the call to `_build_slices` under mypy strict: `Mapping`'s key type is *invariant*, so `dict[str, V]` is not assignable to `Mapping[Hashable, V]`. Narrowing honestly would require widening `_build_slices` + four downstream helper signatures to `Mapping[str, ...]`, which is out of scope for this surgical hardening PR. `dict[Hashable, ...]` is type-*sound* (`str` is a subtype) and matches the parameter; I added a docstring note explaining the constraint.

**3. `n_bands <= 0` KeyError — pre-existing, out-of-scope, unreachable via the entry point.** Fix B targets out-of-`[0,1]` *scores*, not a misconfigured *band count*. And `n_score_bands <= 0` already dies earlier: `_score_overlay` (called before the slice path) computes `edges = [i / n_bands ...]`, raising `ZeroDivisionError` for `n_bands == 0` whenever any in-range score exists — so `_score_band_slices` is never reached with `n_bands == 0` and a scored pair. Guarding only the slice path would be an incomplete fix for a separate, pre-existing config-validation gap; flagging it here rather than expanding this PR's scope.

PR #167, opened 2026-07-16
feat(trackers): local-first Trackio ExperimentTracker adapter

fxd24:
## Summary

Adds `TrackioTracker`, a third `ExperimentTracker` backend alongside `MlflowTracker`/`WandbTracker`, giving the autoresearch loop (`langres.optimize`) durable, off-laptop persistence via [Trackio](https://github.com/gradio-app/trackio).

- **Local-first by default.** `resolve_tracker("trackio")` / `TrackioTracker()` writes to a local trackio SQLite store — no credentials, no network, works out of the box.
- **HF Space sync is opt-in, not built-in.** Setting `space_id` (constructor arg, or `TRACKIO_SPACE_ID`) additionally syncs the run to a Hugging Face Space (+ optional `dataset_id`/`TRACKIO_DATASET_ID` for a persistent Dataset). This is **gated**: if `space_id` is set but no HF write token is available (`HF_TOKEN`/`HUGGING_FACE_HUB_TOKEN` env, or a stored `hf auth login`), `start_run` raises an actionable `ValueError` *before* `trackio.init` is ever called. Nothing is deployed or hardcoded — no Space is created unless the caller explicitly configures one with valid credentials.
- **Registered in the existing seam**: `core/trackers/__init__.py`'s `_ADAPTERS`/`__getattr__`, so `resolve_tracker("trackio")` and `from langres.core.trackers import TrackioTracker` both work, with the same "missing extra" `ImportError` pattern as `mlflow`/`wandb`.
- **`create_trackio_tracker()`** in `clients/tracking.py` mirrors `create_wandb_tracker`'s settings-driven construction (builds an unstarted `TrackioTracker` from `Settings`/explicit overrides — trackio's own `import` stays local to that function body so a `langres[wandb]`-only install never needs `trackio` importable, and vice versa).
- New `[trackio]` extra (pinned `trackio>=0.20,<1`) + dev-group entry; `trackio`/`huggingface_hub` added to the import-budget heavy-module list (both confirmed absent from a bare `import langres`).

## Real installed trackio API (verified, not assumed — 0.20.2)

- `trackio.init(project, name=None, space_id=None, dataset_id=None, config=None, ...) -> Run`. **`dataset_id` is the real, current param name — not `bucket_id`** (no trace of `bucket_id` anywhere in 0.20.2; that was an unverified guess going in).
- `Run` has `.config` (a plain flat `dict`), `.name`, `.url`, and exactly four methods: `log(metrics, step=None)`, `log_system`, `alert`, `finish()` — **no tags API, no artifact API, `finish()` takes no status argument.** `log_artifact`/`set_tags` fold into `run.log`/`run.config` (closest honest mapping); the langres terminal `status` is stamped as a `langres_status` log entry before `finish()`.
- `Run.url` on a Space run is literally the raw `"user/space"` id, not a browsable URL — `run_url` builds `https://{user}-{space}.hf.space/` itself (matches trackio's own internal `deploy.SPACE_HOST_URL` format). `None` for local runs (no static URL outside a notebook embed).
- `trackio.init(space_id=...)` internally raises `huggingface_hub.errors.LocalTokenNotFoundError` for a missing token — the adapter's own `_require_hf_token` guard preempts that with an actionable, project-branded `ValueError` first.
- Both `trackio` and `huggingface_hub` ship `py.typed`, so no `[[tool.mypy.overrides]]` entry was needed (unlike mlflow/litellm/etc.).

## Run-it evidence ($0, offline, real local trackio)

Ran `optimize(space, objective, "amazon_google", tracker=TrackioTracker(project="ar-trackio-smoke"))` over a 3-config k-sweep (k=5/40/80, same RR-budget objective as `examples/research/blocking_recall_autoresearch.py`):

- Local storage: `~/.cache/huggingface/trackio/ar-trackio-smoke.db` (trackio's own default `TRACKIO_DIR`).
- 3 trials -> 3 distinct trackio runs (`optimize_blocking:amazon_google`, `dainty-sunset-0`, `brave-forest-1` — trackio auto-renames on a same-name collision since every trial in one loop shares `context.experiment`).
- **Rejected trials confirmed durable in trackio itself** (not just the in-memory `LoopResult`): `brave-forest-1` (k=80) landed with `"accepted": 0.0, "reduction_ratio": 0.9776` (< the 0.985 budget) read straight back out of trackio's SQLite log table — a genuine over-budget reject, captured end to end through `capture_run` -> `TrackioTracker`.

## Test plan

- [x] `uv run pytest tests/test_trackio_tracker.py tests/test_trackers.py tests/clients/test_tracking.py tests/test_import_budget.py -q` — green (78 fast + 1 slow real-local smoke).
- [x] `uv run pytest --cov=src/langres/core/trackers --cov=src/langres/clients ...` — 100% on `trackio_tracker.py` and `clients/tracking.py`.
- [x] `uv run ruff check .` / `uv run ruff format --check .` / `uv run mypy src` — all clean.
- [x] Full fast suite (`pytest -m "not slow and not integration"`) — 3462 passed, no regressions.
- [x] Import-budget subprocess check — bare `import langres` pulls neither `trackio` nor `huggingface_hub`.
- [x] Docs updated in the same change: `CHANGELOG.md`, `docs/ROADMAP.md`, `docs/EXPERIMENTS.md` (usage snippet), `docs/TECHNICAL_OVERVIEW.md`.

## Known trade-off (flagged, not silently accepted)

`create_trackio_tracker()` lives in `clients/tracking.py` alongside the pre-existing `create_wandb_tracker` (per spec — this keeps the two factories symmetric and in one place). That file has an unconditional top-level `import wandb`, which is load-bearing for `wandb`'s existing "helpful missing-extra `ImportError`" behavior (removing it would regress that). `trackio`'s own import is kept function-local inside `create_trackio_tracker` to avoid the reverse problem for wandb-only installs — but a hypothetical `pip install langres[trackio]`-only (no `[wandb]`, no dev group) caller of `create_trackio_tracker()` specifically (not `resolve_tracker("trackio")` or constructing `TrackioTracker` directly, both of which never touch this file) would still need `wandb` importable. This is documented in the module's docstring. Not a regression for this repo (wandb/mlflow/trackio are all in the dev group together already), but worth knowing about for a future backend-isolation pass.

## Out of scope (confirmed not built)

- No Space deployment, no hardcoded Space name — HF sync is purely a caller opt-in.
- No models-on-Hub / artifact upload beyond the `log_artifact` metadata-entry mapping.

## Summary by Sourcery

Add Trackio as a local-first experiment tracking backend and wire it into the existing ExperimentTracker seam and client factory system, with optional HF Space/Dataset sync and maintained lazy-import behavior.

New Features:
- Introduce TrackioTracker adapter implementing the ExperimentTracker protocol with local SQLite persistence and optional Hugging Face Space/Dataset sync.
- Add create_trackio_tracker factory to build TrackioTracker instances from Settings and explicit overrides, parallel to create_wandb_tracker.
- Expose resolve_tracker("trackio") and TrackioTracker via the trackers package, and add HF/Trackio-related settings fields for token and Space/Dataset configuration.

Enhancements:
- Extend client and tracker lazy-loading infrastructure to include Trackio while preserving import-budget guarantees.
- Update experiment-tracking docs (EXPERIMENTS, TECHNICAL_OVERVIEW, ROADMAP, CHANGELOG) to document Trackio usage and the new off-laptop dashboard option.

Build:
- Define a new [trackio] extra and include trackio in the dev dependency group with version pinning aligned to the verified API.

Tests:
- Add comprehensive tests for the TrackioTracker adapter, including credential-guard behavior, logging semantics, resolve_tracker integration, and a real local smoke test.
- Add tests for the create_trackio_tracker factory and for helpful ImportError behavior when the Trackio extra is missing.
- Extend import-budget tests to ensure trackio and huggingface_hub remain lazily imported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,10 @@
 A `propose → run → evaluate → keep-if-better` hill-climber that tunes a pipeline
 against a **loss-like** objective instead of a saturated F1. M1 proves the loop on
 the **blocking** vertical; the matching vertical (`log_loss` / AUC-PR steering) and
-small-LM fine-tuning are deferred, as is a durable off-laptop **Trackio + Hugging
-Face** dashboard (an optional `tracker=` hook is wired but no-op by default) and an
-Optuna/LLAMBO proposer swap.
+small-LM fine-tuning are deferred, as is an Optuna/LLAMBO proposer swap. A durable
+off-laptop dashboard is no longer deferred: `tracker=resolve_tracker("trackio")`
+(local-first; an HF Space/Dataset sync is a `space_id`/`HF_TOKEN` opt-in) plugs
+straight into `optimize`/`run_loop` via the existing `tracker=` hook (see below).
 
 #### Added
 
@@ -718,10 +719,14 @@ still pulls no `mlflow`/`wandb`).
   exit, and sets the `current_run` contextvar. **`store=None` writes nothing.**
 - **`ExperimentTracker` Protocol + adapters** (`langres.core.trackers`) — an
   Accelerate-style seam (`NoOpTracker` null default, `MultiTracker` fan-out to run
-  MLflow *and* W&B at once, `resolve_tracker` dispatch) with lazy **MLflow** and **W&B**
-  adapters behind the `[mlflow]` / `[wandb]` extras (a missing extra raises a helpful
-  `pip install 'langres[<backend>]'` `ImportError`). MLflow defaults to a local file
-  store out of the box; W&B supports keyless `offline`/`disabled` runs for CI/no-key use.
+  MLflow *and* W&B at once, `resolve_tracker` dispatch) with lazy **MLflow**, **W&B**,
+  and **Trackio** adapters behind the `[mlflow]` / `[wandb]` / `[trackio]` extras (a
+  missing extra raises a helpful `pip install 'langres[<backend>]'` `ImportError`).
+  MLflow defaults to a local file store out of the box; W&B supports keyless
+  `offline`/`disabled` runs for CI/no-key use; **Trackio is local-first** (a local
+  SQLite store, zero credentials) with an opt-in HF Space/Dataset sync
+  (`space_id`/`dataset_id`) gated behind an actionable `ValueError` when no
+  `HF_TOKEN` is available — verified against the installed trackio 0.20.2 API.
 - **LLM trace correlation** — `capture_run` sets `current_run`; `JudgementLog` records
   the active `run_id`, and `LLMJudge` injects litellm `metadata` (`langres_attempt_id`
   + pair ids + decision step) on **both** the sync (`forward`) and async
@@ -733,7 +738,9 @@ still pulls no `mlflow`/`wandb`).
   later eval run threads it into `parent_run_id` (compile → eval lineage).
 - **`Settings`** — `RUN_STORE_PATH`, `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT` (the
   MLflow ones consumed by `MlflowTracker`; a zero-config default `store` from
-  `RUN_STORE_PATH` is deferred to the benchmark wrap — pass `store=` explicitly today).
+  `RUN_STORE_PATH` is deferred to the benchmark wrap — pass `store=` explicitly today),
+  and `HF_TOKEN` / `TRACKIO_SPACE_ID` / `TRACKIO_DATASET_ID` (consumed by
+  `TrackioTracker` / `create_trackio_tracker`; all optional -- local runs need none).
   Docs: `docs/EXPERIMENTS.md`; runnable zero-spend
   `examples/research/experiment_tracking_demo.py`.
 
--- a/docs/EXPERIMENTS.md
+++ b/docs/EXPERIMENTS.md
@@ -373,13 +373,25 @@ records = RunStore("tmp/autoresearch/ag_blocking.jsonl").read()   # last-wins pe
 accepted = [r for r in records if (r.metrics or {}).get("accepted") == 1.0]
 ```
 
-This persistence is **local-only for now.** A durable, off-laptop dashboard
-(Trackio + Hugging Face, with the winning artifact pushed to the Hub) is
-**deferred** — an optional `tracker=` hook is wired into `optimize`/`run_loop` but
-defaults to a no-op. Also deferred: swapping the exhaustive grid proposer for an
-Optuna/LLAMBO one, and the **matching vertical** (steering a judge on `log_loss` /
-AUC-PR) plus small-LM fine-tuning. **M1 proves the loop on the blocking vertical
-only** (epic #145).
+The `RunStore` JSONL is the durable local record; `store=None` skips it entirely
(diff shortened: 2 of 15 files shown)


fxd24:
## Review fix batch applied (commit 4282ba5)

**Addressed:**

- **[A · correctness] `create_trackio_tracker` no longer transitively needs the `wandb` extra.** `create_wandb_tracker` now imports `wandb` lazily inside its own body (mirroring the trackio factory's lazy adapter import), so importing `langres.clients.tracking` — which `from langres.clients import create_trackio_tracker` does — never requires `wandb`. A wandb-less install surfaces the branded `pip install 'langres[wandb]'` `ImportError` at call time. Existing wandb factory tests re-target the `wandb.init` seam via `sys.modules["wandb"]` (7 patch sites) and still pass; added `test_needs_no_wandb_extra` proving `create_trackio_tracker` works with `wandb` absent.
  - *Minor deferral, flagged:* the branded wandb `ImportError` now fires at first use (`start_run`→`create_wandb_tracker`) rather than at `resolve_tracker("wandb")` time. No test relied on resolve-time; the message stays branded/actionable. Kept surgical (didn't add an `import wandb` to `wandb_tracker.py`, which would be an unused-import).
- **[B · correctness, core-tier] `log_params`/`set_tags` now actually persist.** Verified against trackio 0.20.2 source: `Run.config` is flushed to SQLite only on the *first* `run.log` (one-way `_config_logged` latch) and never on `finish()`. Both methods now also emit via `run.log` under `param.`/`tag.` prefixes (config still updated for the in-memory `native` view). Regression tests assert a param/tag logged *after* a metric still goes through `run.log`. (`capture_run`/`_RunHandle` never call these, so the autoresearch loop was unaffected — this only bit direct/`MultiTracker` callers.)
- **[C · hardening] `run_url` no longer raises on a namespace-less `space_id`** — returns `None` before the `"/".split` unpack. Test added.
- **[D · minor] `dataset_id` without `space_id`** now raises an actionable `ValueError` before `trackio.init`, matching the docstring + trackio's own contract. Test added.
- **[E · test hygiene] `test_local_first_when_nothing_configured` made hermetic** — drops ambient `TRACKIO_SPACE_ID`/`TRACKIO_DATASET_ID` (same for the default-settings test).
- **[F · coverage] HF_TOKEN non-override test added** — a pre-exported `HF_TOKEN` wins over `Settings(hf_token=...)` (locks in the `setdefault` precedence).
- **[G · doc] `pyproject` trackio-extra comment** corrected to the verified 0.20.2 API (`dataset_id`; no `bucket_id`).

**Intentionally skipped (per review guidance):**
- Extract a shared `_flatten` helper — mirrors wandb's exactly and extracting touches the existing wandb adapter (out of this PR's surgical scope); follow-on.
- Centralize `space_id`/`dataset_id` resolution — a ~2-line expression at each site; marginal.
- `:174` "comment contradicts behavior" — the `setdefault` letting an explicit ambient token win *is* the intended precedence; comment is accurate (now also locked by test F).

**Verify:** `ruff check .` / `ruff format` / `mypy src` clean; `trackio_tracker.py` and `clients/tracking.py` at **100%** coverage; tracker + runs + optimize + import-budget suites green; slow real-local trackio smoke passes.

PR #169, opened 2026-07-16
refactor(core): split __init__ exports into per-domain fragments

fxd24:
## Why: the worst merge-conflict hotspot in the repo

`src/langres/core/__init__.py` was touched **21 times in the last 30 days**. It carried an alphabetically sorted **~100-name `__all__`** plus the `_LAZY_SUBMODULES` / `_LAZY_SYMBOLS` / `_EXTRA_BY_SYMBOL` dicts.

A sorted list is the exact shape git cannot auto-merge: N concurrent streams each inserting a name **at its sorted position** = N guaranteed conflicts. `src/langres/__init__.py` had the same disease **plus an eager import block whose lines belong to four different work-streams** (verbs/presets, optimize, flywheel, training) — so any two of them collided on one file.

This turns both into **thin aggregators** over per-domain fragments, so N parallel streams edit **disjoint files**.

## What changed

- **`src/langres/core/_exports/`** — 11 fragments (`_models`, `_blocking`, `_matchers`, `_semantic`, `_clustering`, `_resolver`, `_methods`, `_training`, `_tracking`, `_flywheel`, `_eval`), domains derived from the actual module structure.
- **`src/langres/_exports/`** — 6 fragments (`_verbs`, `_optimize`, `_core`, `_flywheel`, `_training`, `_data`), splitting the **eager import block** along work-stream lines.
- Each fragment owns its eager imports + `__all__`, and its slice of the lazy maps. `NAMES` is *derived* (`(*__all__, *LAZY_SUBMODULES, *LAZY_SYMBOLS)`), never hand-maintained.
- Both `__init__.py` keep the PEP-562 `__getattr__` **unchanged** and now hold **zero per-name content**: `core/__init__.py` 325 → 104 lines.
- `.gitattributes`: `merge=union` on the append-mostly fragments, with the aggregators explicitly reset to default 3-way merge (the `_*.py` glob would otherwise match `__init__.py`, where union would duplicate imports or split the `__getattr__` body). Verified with `git check-attr`.

- **`tests/test_export_fragments.py`** — the fragment contract, tested rather than trusted (added in response to the Sourcery review; see the review-response comment). Fragments are discovered via `pkgutil`, so new ones are covered automatically. **Every guard was verified by mutation** — each was seen to *fail* before being allowed to pass:

| Injected fault | Caught by |
|---|---|
| a name owned by two fragments (silent last-wins on `**`-merge) | `AssertionError: 'Resolver' is declared by BOTH ..._clustering and ..._resolver` |
| a core lazy symbol with no `[extra]` | `core lazy symbols without a [extra] hint: ['VectorBlocker']` |
| an `__all__` entry with no import | `AttributeError: module '..._blocking' has no attribute 'GhostBlocker'` (at import) |
| hand-edited `NAMES` | `test_names_is_derived_not_hand_maintained` |

  That last-but-one row matters: it's exactly the `merge=union` downside, and it dies **loudly at import**. The core-extra row closes a trap this refactor itself introduced — `core.__getattr__` does `_EXTRA_BY_SYMBOL[name]` (a direct index); the pre-split dict comprehension defaulted to `"semantic"` so totality was automatic, and explicit per-fragment extras removed that net.

**Adding an export now touches one fragment file.** Only a brand-new *domain* touches an aggregator.

## Zero behaviour change — the set-identity proof

A script captured the surface before and after (`__all__`, both lazy maps, the eager/lazy split, `dir()`) and diffed them:

```
IDENTICAL  core_all                 (103 names)
IDENTICAL  core_eager_names         <- the eager/lazy split is preserved exactly
IDENTICAL  core_extra_by_symbol
IDENTICAL  core_lazy_submodules
IDENTICAL  core_lazy_symbols
IDENTICAL  root_all                 (36 names)
IDENTICAL  root_eager_names
IDENTICAL  root_extra_by_symbol
IDENTICAL  root_lazy_symbols
DIFF       core_dir: only-before=['TYPE_CHECKING'] only-after=['_exports']
DIFF       root_dir: only-before=['TYPE_CHECKING'] only-after=['_exports']

PUBLIC SURFACE SET-IDENTICAL: True
```

The only movement is in `dir()`, and only in **private** names not in `__all__`:
- `-TYPE_CHECKING` — an accidental re-export of the `typing` helper (the `TYPE_CHECKING` blocks moved into the fragments). Verified nothing references `langres.core.TYPE_CHECKING`.
- `+_exports` — the new private fragment package.

`core_eager_names` being identical is the load-bearing one: **no lazy import silently became eager**, and no eager one became lazy.

## Verification

- **`tests/test_import_budget.py` + `tests/test_public_surface_dod.py`: 37 passed, 12 skipped** — identical to the pre-change baseline (skips are uninstalled extras). The import-budget test is subprocess-based and is exactly what a broken lazy seam or an import cycle would trip.
- **`uv run mypy src/`: `Success: no issues found in 147 source files`.** (A first pass showed 9 errors both before *and* after — diffed error-for-error, zero new; they turned out to be artifacts of the extras being absent from a fresh venv, and vanish once `--all-extras` is installed.)
- **`uv run ruff format src/ && uv run ruff check src/`**: clean.
- **Full local suite (`uv run pytest`, all extras): 3574 passed, 4 skipped, 0 failed** (16m29s).
- **Clean-room check** — built the wheel and installed it into a fresh **core-only** venv (no extras), i.e. the real user path the local suite can't exercise (it has extras installed):
  ```
  import langres            : OK
  langres.__all__           : 36 names
  langres.core.__all__      : 103 names
  heavy modules leaked      : none
  eager surface works       : AllPairsBlocker / Resolver / dedupe
  lazy [semantic] hint      : langres.core.VectorBlocker requires the 'semantic' extra: pip install 'langres[semantic]'
  lazy [trained] hint       : langres.derive_threshold requires the 'trained' extra: pip install 'langres[trained]'
  ```
  All 19 fragment files ship in the wheel (hatchling auto-discovers the new subpackages).
- **CI: green, 0 failing checks** — lint, CodeQL, Analyze, build, test (3.13), test-core-only, test-finetune, Sourcery.

### The mechanism actually works (simulated two concurrent streams)

| Scenario | Result |
|---|---|
| **Old** design — 2 streams insert into one sorted `__all__` | **CONFLICT** (2 markers) |
| **New** — 2 streams → *different* fragments | **clean merge** |
| **New** — 2 streams → *same* fragment (`merge=union`) | **clean merge**, both names kept |

Union merge is safe here precisely because the fragments are append-mostly and aggregation is order-insensitive and de-duplicating (`sorted({...})` over every fragment's `NAMES`).

**The honest trade-off**, also documented in `.gitattributes`: union merge is not semantic. If one stream *removes* an export while another edits the same fragment, union resurrects the `__all__` entry without its import — and `from ._x import *` then raises `AttributeError` at import. That failure is **immediate and loud** (collection dies; the import-budget test is the gate), never silent corruption. Removals are rare; additions are the hot path.

## Notes for the reviewer

- mypy's re-export rules were **probed, not assumed**: star-import re-exports under `--strict` (which implies `no_implicit_reexport`), so `from langres.core import AllPairsBlocker` keeps type-checking. A modules-as-`Protocol` variant to collapse the composition was tried and **rejected** — mypy rejects a tuple of modules against a Protocol.
- Docs updated in the same change (`CLAUDE.md`, `docs/TECHNICAL_OVERVIEW.md`): contributors are pointed at the fragments, not the sorted `__all__`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,32 @@
+# Merge policy for the per-domain export fragments (W-1a).
+#
+# langres/__init__.py and langres/core/__init__.py used to carry one sorted
+# ~100-name __all__ plus a shared eager import block -- the single worst
+# merge-conflict hotspot in the repo (21 touches in 30 days). A sorted list is
+# the exact shape git cannot auto-merge: N concurrent streams each inserting a
+# name at its sorted position = N guaranteed conflicts. The list now lives in
+# per-domain fragments so streams edit disjoint FILES.
+#
+# `merge=union` (a built-in git driver) closes the remaining case: two streams
+# appending to the SAME fragment. Instead of a conflict, git keeps both sides'
+# lines -- which is correct here precisely because the fragments are
+# append-mostly and the aggregation is order-insensitive and de-duplicating
+# (`sorted({...})` over every fragment's NAMES).
+#
+# The trade-off, stated honestly: union merge is not semantic. If one stream
+# REMOVES an export while another edits the same fragment, union resurrects the
+# removed __all__ entry without its import -- and `from ._x import *` then
+# raises AttributeError on import. That failure is immediate and loud (every
+# test collection dies; tests/test_import_budget.py is the gate), never silent
+# corruption -- which is what makes the trade acceptable for append-mostly
+# files. Removals are rare; additions are the hot path.
+src/langres/_exports/_*.py             merge=union
+src/langres/core/_exports/_*.py        merge=union
+
+# The `_*.py` globs above also match `__init__.py`. The aggregators are NOT
+# append-mostly -- they carry import blocks and the PEP-562 __getattr__, where
+# keeping both sides' lines would duplicate imports or split a function body.
+# Reset them to git's default 3-way text merge (conflict markers, human
+# resolves).
+src/langres/_exports/__init__.py       !merge
+src/langres/core/_exports/__init__.py  !merge
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -49,7 +49,9 @@ langres/
 │   ├── optimize.py     # langres.optimize / score_blocking: import-light autoresearch facade over blocking search
 │   ├── eval.py         # Curated evaluation facade (lazy): evaluate, list_benchmarks/get_benchmark, ER metrics
 │   ├── cli.py          # langres CLI: review / export-csv / import-csv (labeling loop)
+│   ├── _exports/       # per-domain fragments composing the ROOT __all__ + lazy maps (add a root export HERE, not in __init__.py)
 │   ├── core/           # Low-level primitives + the Resolver
+│   │   ├── _exports/       # same, for langres.core (add a core export HERE, not in core/__init__.py)
 │   │   ├── resolver.py     # Resolver.from_schema / resolve / save / load
 │   │   ├── presets.py      # judge presets ("auto" fail-fast/string/embedding/zero_shot_llm/prompt_llm), DEFAULT_AUTO_MODEL, NoJudgeAvailableError, spend cap
 │   │   ├── method_registry.py  # ONE MethodSpec registry: judge/method name -> builder + identity (all three dispatch paths resolve here)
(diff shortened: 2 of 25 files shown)


fxd24:
## Review response (@sourcery-ai)

Both points addressed in 6aa2f74 — thanks, the first is a **real bug risk** and I'd under-guarded it.

### 1. Extra-mapping conflicts — agreed on the substance, different shape

You're right that `**`-merging silently last-wins on a duplicate key, and it's worse than it looks: this PR marks the fragments `merge=union` in `.gitattributes`, so git deliberately keeps **both** sides' lines rather than conflicting — which makes accidental duplication *more* likely, not less.

I did **not** take the suggested runtime check in `_exports/__init__.py`, for two reasons:
- it taxes every production `import langres` forever to catch a contributor mistake, and
- it only guards `EXTRA_BY_SYMBOL`, while the same merge silently last-wins on `LAZY_SYMBOLS` (wrong owning module), and several other contract holes have no guard at all.

Instead the contract is now a **test** (`tests/test_export_fragments.py`) — zero runtime cost, catches every hole in CI. Fragments are discovered via `pkgutil`, so new ones are covered automatically (including "added a fragment, forgot to wire it").

**Every guard was verified by mutation** — each was seen to *fail* before being allowed to pass:

| Injected fault | Caught by |
|---|---|
| `Resolver` declared by two fragments | `AssertionError: 'Resolver' is declared by BOTH ..._clustering and ..._resolver` |
| core lazy symbol with no extra | `core lazy symbols without a [extra] hint: ['VectorBlocker']` |
| `__all__` entry with no import | `AttributeError: module '..._blocking' has no attribute 'GhostBlocker'` (at import) |
| hand-edited `NAMES` | `test_names_is_derived_not_hand_maintained` |

That third row is worth calling out: it's exactly the `merge=union` downside (a removal + an edit interleaving can resurrect an `__all__` entry without its import). It fails **immediately and loudly at import**, never silently — which is what makes the union trade acceptable.

Your prompt also surfaced a trap I'd introduced without noticing: `core.__getattr__` does `_EXTRA_BY_SYMBOL[name]` (a direct index, not `.get()`). The pre-split code built that map with a comprehension defaulting to `"semantic"`, so totality was automatic; my explicit per-fragment extras removed that safety net. A future lazy symbol without an extra would `KeyError` *inside* the `except ImportError` handler. Now locked by `TestCoreExtraBySymbolIsTotal`.

### 2. Root vs. core fragment overlap — documented

Fair; it reads as duplication. It's re-export with distinct ownership, now spelled out in `langres/_exports/__init__.py`:
- `langres/core/_exports/*` **owns** the export (imports the implementation module, decides eager-vs-lazy).
- `langres/_exports/*` **re-surfaces** a curated subset at the root, mostly importing from `langres.core` itself. The root is a smaller, opinionated front door — **103 core names vs 36 root** — so it is deliberately *not* a mirror. Adding a name to `core` does not imply adding it here.

PR #178, opened 2026-07-16
fix(core): give Resolver a per-instance spend cap (B1)

fxd24:
Clears **fork blocker B1**. Both a refactor prerequisite and a live product bug that could cost real money today.

## The bug

`Resolver` — the entire low-level public API — had **no budget guard at all**. `_SpendCappedMatcher` was applied at exactly two sites, both in `core/presets.py`, which `Resolver` is architecturally forbidden to import (`resolver.py` ~L162-166). Only `link`/`dedupe` were ever capped. `from_schema`'s own docstring advertised the hole:

> **Caution**: … runs **UNCAPPED** — there is no `budget_usd` on this method and nothing stops a runaway bill.

Two further bugs in the cap itself, both harmless for a one-shot verb call and wrong for a long-lived object:

1. The `SpendMonitor` was built **inside `forward()`** ⇒ per-call, not per-instance. `r.resolve(a); r.resolve(b)` = **2× budget**; N resolves = N × budget.
2. No pre-check ⇒ an already-tripped cap paid for one more call before refusing.

## The guarantee (stated honestly)

**Spend is bounded by `budget_usd` plus the cost of at most one further call.** An LLM call's cost is not knowable until it has been made, so "check before" can only mean *"am I already at/over budget? then refuse the NEXT call"*. No doc, test, or comment in this PR claims more. `SpendMonitor.check()` is `spent > budget` (strictly), so landing exactly on the budget is not a breach.

## Where `_SpendCappedMatcher` went, and why

Two new **core leaves**:

| module | holds | W4 target pkg |
|---|---|---|
| `core/spend.py` | `SpendMonitor`, `BudgetExceeded` (moved out of `clients/openrouter.py`) | `core` |
| `core/spend_cap.py` | `SpendCappedMatcher`, `DEFAULT_BUDGET_USD`, `UNCAPPED_BUDGET_USD`, `effective_budget` | `architectures` |

The cap must sit **beneath both** `presets` (above `Resolver`) and `Resolver`. Putting it in a new module that imports `clients.openrouter` does **not** work, and I measured rather than assumed:

```
spend_cap -> clients.openrouter -> core.benchmark -> core.presets -> spend_cap
```

That closes a cycle and grows the all-edges SCC **10 → 11**. So the ledger had to come with it. That move is also just *right*: `SpendMonitor` is pure USD arithmetic with nothing OpenRouter-specific about it ("Pure — no I/O beyond `logging`"), and `BudgetExceeded.partial_judgements` already depended on `core.models` — a `clients` module owning a primitive `core` must enforce was backwards. `clients.openrouter` re-exports both names, so `from langres.clients.openrouter import SpendMonitor, BudgetExceeded` and `langres.BudgetExceeded` are unchanged.

The re-export is **lazy** (PEP 562 `__getattr__`), and that is measured too: an eager one is the first *toplevel* `clients -> core` edge in the package and drags `clients` into the refactor's **runtime cross-package cycle (9 → 10 packages)**. `openrouter` no longer *uses* these names — it only forwards them — so a lazy alias is the honest shape as well as the cheap one.

## Ratchet: unchanged

| view | baseline `29e55bd` | this PR |
|---|---|---|
| runtime cycles | **0** | **0** |
| all-edges SCC | **[10, 9, 3, 2]** | **[10, 9, 3, 2]** |
| W4 counterfactual (all-edges) | [13] | [13] |
| W4 counterfactual (toplevel) | [9] | [9] |

`tools/refactor_target_packages.json` gains the two modules (the mapping must be total — *"no wave may discover a homeless file"*), with the measurement recorded in `_comment`.

## Design calls

- **The monitor, not the wrapper, is the durable thing.** `Resolver` owns one `SpendMonitor` and re-wraps `self.module` per scoring pass. `self.module` is *reassignable* — `dedupe()` wraps it in a `LoggingMatcher`, `distil()` replaces it outright — so a wrapper pinned at `__init__` would meter a stale matcher.
- **The cap never enters the `module` slot.** `self.module` stays the raw matcher, so `fit()`'s `isinstance(..., SupervisedFitMixin)` checks, `save()`'s registry lookup, and `.model`/`.config` access all keep working. Wrapping the slot would have broken `save()` for every Resolver.
- **Exactly one cap on the `dedupe` path.** `build_resolver` hands the Resolver the *unwrapped* scorer plus `budget_usd`; `resolve_judge` still wraps for `link` (which scores one pair directly, no Resolver). Threading `budget_usd` through is **mandatory, not cosmetic**: a Resolver defaulting to $1.00 under a `dedupe(budget_usd=5.0)` would trip at $1.00 and silently break the verb's budget.
- **`None` means the default, never "uncapped"** — you cannot disable the cap by forgetting to pass it. `UNCAPPED_BUDGET_USD` (`float("inf")`) is the deliberate, in-writing opt-out.

## Verbs: unchanged

`link` keeps `LoggingMatcher(Cap(judge))` untouched. `dedupe` becomes `Cap(LoggingMatcher(judge))` — verified equivalent: `LoggingMatcher` logs *before* it yields, so every paid judgement, including the one that trips the cap, still lands in the log exactly once. All 3652 fast tests pass.

## Behavior change (CHANGELOG'd)

A paid `Matcher` on a `Resolver` that previously ran unbounded now raises `BudgetExceeded` past $1.00. Free matchers meter $0 and never trip. This is the point of the fix.

## Proof the cap binds

The load-bearing test is `test_two_resolve_calls_share_one_budget`. Verified by **mutation** — reinstating the pre-B1 per-`forward()` monitor fails 7 of the 8 per-instance tests:

```
FAILED tests/core/test_resolver_spend_cap.py::TestResolverCapIsPerInstance::test_two_resolve_calls_share_one_budget
FAILED tests/core/test_resolver_spend_cap.py::TestResolverCapIsPerInstance::test_a_tripped_resolver_costs_nothing_on_the_next_resolve
FAILED tests/core/test_resolver_spend_cap.py::TestResolverCapIsPerInstance::test_predict_shares_the_same_ledger_as_resolve
FAILED tests/core/test_resolver_spend_cap.py::TestResolverCapIsPerInstance::test_budget_survives_module_reassignment
FAILED tests/core/test_spend_cap.py::TestSpendCapIsPerInstanceNotPerCall::test_budget_binds_across_two_forward_calls
FAILED tests/core/test_spend_cap.py::TestSpendCapIsPerInstanceNotPerCall::test_a_tripped_cap_costs_nothing_on_the_next_call
FAILED tests/core/test_spend_cap.py::TestSpendCapIsPerInstanceNotPerCall::test_two_wrappers_sharing_one_monitor_share_one_budget
========================= 7 failed, 1 passed in 0.38s ==========================
```

The 8th (`test_separate_resolvers_do_not_share_a_budget`) passes either way **by design** — it guards the opposite over-correction, a global ledger.

Edges covered per the `core` tier: no key (unchanged fail-fast), over budget, **exactly at** budget, **zero** budget (refuses the first paid call; still runs free matchers), `None` budget, uncapped, missing/`None` `cost_usd`, and `budget_usd`+`monitor` both passed.

## Gates (real output)

- `uv run pytest -m "not slow and not integration"` → **3652 passed, 1 skipped**
- `uv run mypy src/` → **Success: no issues found in 152 source files**
- `uv run ruff check src/ tests/ examples/` → **All checks passed!**; `ruff format --check` → **421 files already formatted**
- `core` contract gate (`--include="src/langres/core/*" --fail-under=98`) → **98.27%** ✅; `spend.py` and `spend_cap.py` both **100.00%**

## Zero spend

No paid path was ever invoked. Every "paid" matcher in these tests is a fake stamping a made-up `provenance["cost_usd"]` — the cap only ever reads the number a judgement reports, so a fake proves the arithmetic exactly as well as a real model would, for $0.

## Docs that had to stop lying (same PR)

`from_schema`'s docstring (advertised UNCAPPED), `docs/TECHNICAL_OVERVIEW.md`, `CascadeMatcher`'s docstring (justified "caps belong outside" with the now-fixed per-call reset), `examples/flywheel_closed_loop.py`'s `_OuterSpendCap` (documented the fresh-monitor-per-`forward` as *deliberate* — it is now redundant with the core cap and says so), `CLAUDE.md`, `.claude/rules/component-design.md`, `CHANGELOG.md`.

## Known gap (scoped deliberately, named not hidden)

The cap meters the **scoring** path — `resolve()` / `predict()`. `fit(method=Platt()/Isotonic())` and `distil()` still score through `self.module.forward` directly, **uncapped** (`resolver.py:954`, `:1114`, `:1181`). `fit(..., pairs=...)`'s validation pass is uncapped too, though its matcher must be a `SupervisedFitMixin` (sklearn, $0).

Left open rather than half-closed: capping `distil()` properly means metering DSPy's compile calls, which never reach `self.module.forward` at all — a cap on its validation pass alone would read as protection while covering little. Named in the CHANGELOG, `__init__`'s docstring and `TECHNICAL_OVERVIEW` instead. A cap doc that overstates its own reach is the same class of bug as the one B1 fixes. Happy to close it here or as a follow-up if you'd rather.


fxd24:
@sourcery-ai review

fxd24:
@sourcery-ai review

PR #179, opened 2026-07-17
refactor(core): clear the two W4 fork blockers — fit() topology (B5) + save/load identity (B4)

fxd24:
Clears the last two fork blockers gating **W4** (the `ERModel`/`architectures` wave). Both defend the plan's invariant **I2 — "one architecture = one class = one identity"** — which would otherwise be false on day one. Two sequential commits in one worktree (both touch `resolver.py`).

**No `ERModel` / `architectures/` here — that is W4.** This only removes the blockers.

## B5 — `fit()` mutated topology, falsifying I2

`_fit_finetune` repoints the matcher slot (`self.module = LLMMatcher(...)`), so `FuzzyString().fit(method=QLoRA(...))` would silently become LLM-backed while still being class `FuzzyString`.

A Resolver subclass now declares `accepted_method_kinds`; `fit()` refuses anything else **at the boundary, before dispatch**:

```
UnsupportedMethodKind: FuzzyString does not accept method kind 'finetune'
(fine-tune qwen/Qwen3-0.6B (4-bit QLoRA r=16)); FuzzyString accepts: 'calibrate'.
Fitting with 'finetune' would change the pipeline's topology, leaving FuzzyString
naming a pipeline it no longer is. Use a plain Resolver (which accepts every kind)
or an architecture whose identity includes 'finetune'.
```

- Base `Resolver` stays `None` = **permissive** — it makes no identity claim; no existing fit path changes.
- Real kinds are `prompt` / `finetune` / `calibrate`, **verified at runtime** off `Method.kind` (a `ClassVar`), not guessed.
- `UnsupportedMethodKind` subclasses `TypeError` — `kind` is strategy-*type* identity, and it matches the sibling `TypeError` `_fit_finetune` already raises.
- Rejected alternatives (new-instance-returning `fit()`; advisory I2) are recorded **in-code** with why, so this is cheap to flip.

## B4 — `save`/`load` erased architecture identity

`save` now stamps an **optional** `model_class`; `load` reconstructs that class. Fresh-process proof:

```
--- resolver.json ---            --- FRESH PROCESS: Resolver.load(...) ---
"artifact_version": "1"            loaded class : FixtureFuzzyString
"model_class": "fixture_fuzzy_string"   accepted_method_kinds: ['calibrate']
```

- **Absent ⇒ plain `Resolver`** (every pre-0.4 artifact; an unregistered subclass degrades to today's behaviour, not an error).
- **`ARTIFACT_VERSION` stays `"1"`** — `_check_versions` rejects older *and* newer, so a bump would break existing 0.3.0 artifacts in both directions to buy a read that costs one `if`.
- **Models get their own namespace in the existing `registry.py`** (`register_model`/`get_model`/`model_type_name`), not a parallel registry: a component *fills* a slot, a model *owns* the slots — one dict would make `"type_name": "fuzzy_string"` resolvable in a blocker slot. `registry.py` already keeps schemas separate for exactly this reason.
- `get_model` returns a bare `type` like `get_component`: naming `Resolver` there (even under `TYPE_CHECKING`) would knot `registry <-> resolver` and grow the SCC.
- `config_dict()` still emits only `components`, so `model_class` stays **outside** `recipe_id`'s hash domain — no recipe hash forks.

## Gates (measured, not assumed)

| Gate | Result |
|---|---|
| Import ratchet | runtime cycles **0**, all-edges SCC **[10, 9, 3, 2]** — unchanged |
| `core` coverage (CI cmd) | **98.27%** vs gate 98 — exit 0 |
| Tests | **3681 passed**, 1 skipped (`-m "not slow and not integration"`) |
| `ruff check` / `format --check` | clean / 362 formatted |
| `mypy src/` | Success, 152 files |
| import_budget + public_surface_dod | 45 passed, 1 skipped |

Docs updated in the same PR (`docs/TECHNICAL_OVERVIEW.md`: both contracts) plus `fit`/`save`/`load` docstrings.

## Summary by Sourcery

Introduce architecture identity preservation in resolver save/load and enforce method kind constraints for resolver subclasses to protect pipeline topology.

New Features:
- Add model registry namespace for Resolver subclasses so saved artifacts can reconstruct their architecture class via model_class.
- Expose architecture registration and lookup helpers (register_model, get_model, model_type_name) through the core exports API.
- Allow Resolver subclasses to declare accepted_method_kinds, enabling topology-preserving fit contracts for named architectures.

Enhancements:
- Extend ArtifactManifest and Resolver.save/load so resolver.json includes an optional model_class field without changing artifact_version.
- Add UnsupportedMethodKind error and a pre-dispatch gate in Resolver.fit to reject method kinds outside a subclass’s accepted_method_kinds while keeping the base Resolver permissive.
- Document the save/load model_class semantics and accepted_method_kinds contract in TECHNICAL_OVERVIEW and resolver-related docstrings.

Tests:
- Add resolver save/load round-trip tests that verify architecture identity is preserved, including fresh-process checks and failure modes for unknown model_class values.
- Add tests for the model registry namespace, registration, lookup, suggestions, and separation from component registry.
- Add fit-method tests covering accepted_method_kinds behavior, UnsupportedMethodKind semantics, and ensuring the base Resolver’s existing fit paths remain unchanged.
- Introduce test fixture architectures (registered and unregistered Resolver subclasses) to exercise architecture identity and method kind constraints.

### DIFF
--- a/docs/TECHNICAL_OVERVIEW.md
+++ b/docs/TECHNICAL_OVERVIEW.md
@@ -800,6 +800,33 @@ unchanged):
   map with pure NumPy). `Platt()`/`Isotonic()` are import-light (no scikit-learn
   until the fit runs).
 
+#### Which kinds a class accepts: `accepted_method_kinds`
+
+Note the fine-tune bullet above: `fit(method=QLoRA(...))` **repoints the matcher
+slot**. That is deliberate and correct for a plain `Resolver`, which claims no
+identity — "a resolver" is not a topology, so a topology change falsifies
+nothing. It stops being free the moment a *class* names an architecture: a
+`FuzzyString` that quietly became LLM-backed while still being class
+`FuzzyString` is a name that describes a pipeline it no longer is.
+
+So a Resolver subclass may declare the kinds it can absorb without ceasing to be
+itself:
+
+```python
+class FuzzyString(Resolver):
+    accepted_method_kinds = frozenset({"calibrate"})   # no prompt/finetune
+```
+
+`fit()` checks it at the boundary, before dispatch, and refuses anything else
+with `langres.core.UnsupportedMethodKind` (a `TypeError` — `Method.kind` is a
+`ClassVar`, i.e. strategy-*type* identity, so rejecting a kind is rejecting a
+type) naming the class, the kind, `method.describe()`, and the accepted kinds.
+The three real kinds are `"prompt"`, `"finetune"`, `"calibrate"`.
+
+The base `Resolver` leaves `accepted_method_kinds = None` — **permissive**, every
+kind accepted, every fit path unchanged. An empty `frozenset()` is the opposite
+extreme: a fully frozen topology that takes no `method=` at all.
+
 `Resolver.describe()` is the pre-fit honesty device: a per-component string
 tagging each pipeline role **TRAINABLE** (implements a `langres.core.fit`
 Protocol, or is a prompt-compilable `DSPyMatcher`) or **frozen**. A pure string
--- a/docs/TECHNICAL_OVERVIEW.md
+++ b/docs/TECHNICAL_OVERVIEW.md
@@ -204,6 +204,15 @@ It is **optional on purpose**, and the compatibility rules follow from that:
 - `config_dict()` still emits only `components`, so `model_class` stays **outside**
   `recipe_id`'s hash domain and no existing recipe hash forks.
 
+> **Constraint on a registered model:** `load` reconstructs by calling the class
+> with `Resolver.__init__`'s component keywords (`blocker=`, `comparator=`,
+> `matcher=`, `clusterer=`, `calibrator=`). An architecture that narrows
+> `__init__` to its own signature (e.g. `FuzzyString(threshold=0.8)`) raises
+> `TypeError: ... unexpected keyword argument 'blocker'` on load. It must keep
+> accepting the component keywords, or grow an explicit from-components hook.
+> This could not bite before `model_class`, since `load` always built the base
+> `Resolver`.
+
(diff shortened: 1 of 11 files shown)


fxd24:
Thanks @sourcery-ai — triaged all three. Two taken, one declined with reasoning.

**1. Validate `target` is a `Resolver` subclass — TAKEN (real bug, reproduced).**

I probed it before fixing rather than take it on faith, and it reproduces exactly as described:

```
load() returned: NotAResolver
is it a Resolver? False
-> load() silently handed back a non-Resolver
```

Root cause is structural: `registry.py` sits *beneath* `resolver.py`, so it can neither import `Resolver` nor annotate `get_model` with it (doing so would knot the two into an import cycle the ratchet counts). The registry therefore *cannot* type-check its own entries, and `load` deserializes a file — so the check belongs exactly where you put it. After:

```
TypeError: Artifact model_class 'probe_not_a_resolver' is registered to
__main__.NotAResolver, which is not a Resolver subclass; register_model is for
Resolver subclasses (architectures).
```

One deviation: raised as `TypeError`, **not** the suggested `UnknownModelType`. That type is a `KeyError`, and this is a wrong-*type* error, not an unknown-*name* one — reusing it would make `except KeyError` swallow a type bug. Bonus: the guard narrows `target` to `type[Resolver]`, so mypy correctly flagged the `cast` it replaced as redundant and it's now gone.

**2. `UnsupportedMethodKind` message length — TAKEN (trimmed).**

It repeated the kind three times. Now once, while still naming class + kind + `describe()` + accepted kinds:

```
FuzzyString does not accept method kind 'finetune' (fine-tune qwen/Qwen3-0.6B
(4-bit QLoRA r=16)); it accepts: 'calibrate'. That fit would change the pipeline's
topology, leaving FuzzyString naming a pipeline it no longer is. Use a plain
Resolver (which accepts every kind) or an architecture built for it.
```

**3. Reverse map for `model_type_name`'s O(n) scan — DECLINED.**

`model_type_name` runs **once per `save()`** — an operation that writes a directory to disk — over what will be a handful of registered architectures (W4 adds ~2). The scan is unmeasurable against the I/O it accompanies, while a second dict to keep in sync with `_MODEL_REGISTRY` is a genuine bug surface (silent drift → `save` stamps the wrong name). Declining per the repo's "clean over premature optimization" rule; happy to revisit if architectures ever number in the thousands.

Gates after the fixes: **3682 passed**, core coverage **98.27%** (gate 98, exit 0), import ratchet unchanged (runtime **0**, all-edges **[10, 9, 3, 2]**), `mypy` clean, `ruff` clean.

PR #180, opened 2026-07-17
test(packaging): gate the wheel/sdist against shipping unlicensed third-party data

fxd24:
## The exposure (re-measured on this branch, not taken on faith)

`[tool.hatch.build].exclude` keeps the third-party DeepMatcher/Magellan corpora (vendored for research reproducibility; `ATTRIBUTION.md` but **no explicit redistribution licence**) out of the PyPI artifacts. It is **14 hand-written path literals with no test behind them**. Rename `src/langres/data/datasets/` and every literal quietly matches nothing.

**The build then succeeds with zero warnings and ships the corpora.** That silence is the whole problem. This repo has already shipped a 91%-third-party-data wheel once, so this is a regression of a solved problem.

Measured by renaming the directory on disk and rebuilding:

| | wheel (compressed) | wheel (uncompressed) | CSVs | CSV bytes |
|---|---|---|---|---|
| excludes matching (today) | 745,757 B | 2,163,667 B | 13 | 189,707 B (8.8%) |
| excludes dead (post-rename) | 5,902,391 B | 16,072,058 B | 43 | 14,095,044 B (**87.7%**) |

**The sdist carries the identical exposure** (2.26 MB → 7.39 MB; 13 → 43 CSVs). `[tool.hatch.build]` is target-agnostic and `publish.yml` runs `uv build` + `uv publish`, which uploads **both**. Gated both.

## The invariant, and why it is not derived

The redistribution decision is **not derivable from any property of a file**. Every mechanical rule is measurably wrong on this tree:

- *"no `.csv` ships"* — false: 13 CSVs ship deliberately.
- *"small CSVs ship"* — false: `wdc_computers/valid.csv` (6,106 B) is excluded while `fodors_zagat/fodors.csv` (44,963 B) ships. **No threshold separates them.**
- *"a dir with `ATTRIBUTION.md` ships nothing"* — false: `tiny_fixture` has one and ships all 5 CSVs; `abt_buy`/`amazon_google` have one and ship `peeters_sampled_test.csv`.
- *"only id/label pair files ship"* — false: `fodors.csv` and `person_a.csv` are full record text and ship.

It is a per-file human licence judgement, so the expectation must be a per-file human declaration.

**The tempting derivation is vacuous.** Deriving expected-contents from the excludes themselves (*files on disk minus files matching a pattern*) fails open by construction: after a rename the patterns match nothing, so the derived expectation becomes *"everything ships"* — and the fat wheel **passes**. An expectation regenerated from the thing that broke cannot detect it breaking.

So: **two independent descriptions of one decision** — hatchling's deny patterns, and an allow manifest (`SHIPPED_NON_PY_FILES`) — cross-checked against a real `uv build`. Direction of rot is the point: **a deny list rots open; an allow list rots closed.** It is not the 14 literals restated — it is their complement, and it catches drift the literals cannot see (a rename, a *new* unlicensed dataset, a deleted exclude, a backend swap). Scoped to *non-`.py`* rather than `.csv` so a future `.parquet`/`.json` corpus is caught too.

## What is in the gate

1. `test_every_exclude_pattern_still_matches_real_files` — no build; fails naming the **dead literal**. Catches a rename at its source.
2. `test_wheel_ships_exactly_the_declared_data_files` / `..._sdist_...` — ground truth, against the real artifacts.
3. `test_wheel_stays_under_the_size_ceiling` — generous catch-all (4 MB vs 2.16 MB actual) for payload the manifest cannot see (e.g. a corpus inlined into a `.py` literal). Deliberately **secondary**.

## Mutation proof — it bites

**Real directory rename** (`datasets/` → `benchmarks/`, pyproject untouched) — the exact W2/Stream B scenario:
```
AssertionError: 14 exclude pattern(s) in [tool.hatch.build] match no file on disk:
    - src/langres/data/datasets/dblp_acm/*.csv
    ... (all 14)
AssertionError: langres-0.3.0-py3-none-any.whl ships 52 data file(s) that are NOT
declared redistributable in SHIPPED_NON_PY_FILES:
    + data/benchmarks/abt_buy/tableA.csv ...
AssertionError: ... is 16,072,162 B uncompressed, over the 4,000,000 B ceiling.
=== 4 failed in 2.50s ===
```
**Single mangled literal** (`abt_buy/tableA.csv` → `tableA_RENAMED.csv`) — minimal mutation:
```
AssertionError: 1 exclude pattern(s) ... match no file on disk:
    - src/langres/data/datasets/abt_buy/tableA_RENAMED.csv
AssertionError: ... ships 1 data file(s) that are NOT declared redistributable:
    + data/datasets/abt_buy/tableA.csv
=== 3 failed, 1 passed in 1.33s ===
```
The one that passed is the **size check** — a 324 KB CSV does not breach a 4 MB ceiling. That is exactly why size is the secondary signal and contents are the primary one. Unmutated: `4 passed in 2.47s`.

## Which job runs it — verified, not assumed

Collection tested against the **exact marker expressions in `test.yml`**:

| job | marker expr | trigger | collects all 4? |
|---|---|---|---|
| `test` | `not integration and not slow and not finetune` | `pull_request` | ✅ |
| `test-full` | `not integration and not finetune` | `push` → `[main, feat/arch-refactor]` + `workflow_dispatch` | ✅ |

**Not marked `slow`** — deliberately, and this departs from the brief's expectation that a wheel build would be slow. Measured: **the whole file is ~2.5s** (`uv build` is ~1.6s warm; `uv sync` warms the build env in every CI job). This repo's `slow` marker means torch inference / ~10 min. Marking it `slow` would have **removed it from the per-PR job**, so a rename would only be caught *after* merge. Unmarked, it runs in **both** — the rename is caught on the PR that does it.

## Not done, on purpose

No rename, no move of `data/`, no restructure. This is the gate that makes that move safe; the move is a separate, undecided wave.

## Verification

`3661 passed, 1 skipped, 114 deselected` (99% coverage) · `ruff check` clean · `ruff format --check` 423 files clean · `mypy src/` clean (152 files) · `test_import_tangle.py` unaffected (no new `src` imports).

## Summary by Sourcery

Add a packaging licence gate that ensures built wheels and sdists only include explicitly approved non-Python data files and that Hatch build exclude patterns remain effective.

Documentation:
- Document in `pyproject.toml` how Hatch exclude patterns interact with the new tests and the SHIPPED_NON_PY_FILES manifest, and the requirement to update both when datasets move or change.

Tests:
- Add tests that build real wheel/sdist artifacts with `uv` and assert their non-`.py` contents match a declared allowlist of redistributable data files.
- Add a test ensuring every `[tool.hatch.build].exclude` path pattern still matches files on disk, catching silent failures from dataset renames or moves.
- Add a size-based test that the uncompressed wheel remains under a configurable ceiling as a secondary guard against unexpected payload bloat.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,6 +141,15 @@ build-backend = "hatchling.build"
 # (synthetic, BSD-3), fodors_zagat (small), and the peeters_sampled_test.csv
 # pair sets (id/label triples only -- no third-party record text, and the
 # replication seam's sampling stays inspectable from the wheel).
+#
+# !! THESE ARE PATH LITERALS AND THEY FAIL SILENTLY !! Rename a directory below
+# and every pattern quietly matches nothing: the build still succeeds, with zero
+# warnings, and publishes the corpora (measured: 2.16 MB -> 16.07 MB uncompressed
+# wheel, 87.7% third-party CSVs). tests/test_wheel_contents.py is the gate --
+# it builds the real artifacts and compares them against SHIPPED_NON_PY_FILES,
+# an independent allow manifest of what may ship, and separately asserts every
+# pattern here still matches a real file. Moving a dataset means updating BOTH
+# this list and that manifest; the test tells you if you missed one.
 exclude = [
     "src/langres/data/datasets/dblp_acm/*.csv",
     "src/langres/data/datasets/dblp_scholar/*.csv",
--- a/tests/test_wheel_contents.py
+++ b/tests/test_wheel_contents.py
@@ -221,6 +221,17 @@ def test_sdist_ships_exactly_the_declared_data_files(
     with tarfile.open(sdist) as tf:
         names = [m.name for m in tf.getmembers() if m.isfile()]
 
+    # Locating the package by a path marker already fails closed -- if the sdist
+    # layout moved, nothing matches and every manifest entry reports as missing.
+    # But "23 files missing" would misdiagnose a layout change as a deletion, and
+    # a gate whose message points at the wrong cause wastes the incident. Say so.
+    assert any(marker in name for name in names), (
+        f"{sdist.name} contains no member under '{marker}', so this test cannot see "
+        "the package at all. The sdist layout changed (a src-layout move, or a "
+        "different build backend) -- update the marker; do NOT assume the data is "
+        "simply gone."
+    )
+
     shipped = {
         name.split(marker, 1)[1] for name in names if marker in name and not name.endswith(".py")
     }
@@ -228,6 +239,17 @@ def test_sdist_ships_exactly_the_declared_data_files(
 
 
 def _assert_manifest(shipped: set[str], *, artifact: str) -> None:
+    # Both callers filter `.py` out of `shipped`, so a `.py` entry in the manifest
+    # could never match and would surface as a bogus "missing file" -- a real
+    # licence decision is not being made about langres's own source. Name the
+    # actual mistake instead.
+    stray_py = {p for p in SHIPPED_NON_PY_FILES if p.endswith(".py")}
+    assert not stray_py, (
+        f"SHIPPED_NON_PY_FILES lists {len(stray_py)} `.py` path(s): {sorted(stray_py)}. "
+        "The manifest declares the non-`.py` data payload only; `.py` files are "
+        "langres's own source and are not part of this gate."
+    )
+
     unexpected = shipped - SHIPPED_NON_PY_FILES
     missing = SHIPPED_NON_PY_FILES - shipped
     assert not unexpected, (
--- a/tests/test_wheel_contents.py
+++ b/tests/test_wheel_contents.py
@@ -1,0 +1,261 @@
+"""Licence gate: the published artifacts must ship only data we may redistribute.
+
+`[tool.hatch.build].exclude` in `pyproject.toml` keeps the large third-party
+DeepMatcher/Magellan benchmark corpora (vendored for research reproducibility,
+with ATTRIBUTION.md but **no explicit redistribution licence**) out of the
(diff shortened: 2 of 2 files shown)


fxd24:
## Review response (@sourcery-ai)

**Taken — both were the gate misdiagnosing its own failure**, which matters more than usual in a file whose value *is* the failure message. Fixed in `7a235d5`, each mutation-checked:

- **`.py` in the manifest** — both callers filter `.py` out of the shipped set, so a stray `.py` entry could never match and surfaced as a bogus *"missing file"*. Now: `SHIPPED_NON_PY_FILES lists 1 '.py' path(s): ['verbs.py']. The manifest declares the non-'.py' data payload only`.
- **sdist marker** — it already failed *closed* (marker matches nothing → all 23 entries report missing), but that misreads a **layout change as a mass deletion**. Now asserts the marker was found first: `contains no member under '/lib/langres/' ... The sdist layout changed -- update the marker; do NOT assume the data is simply gone.` I kept it to one assertion rather than the suggested helper — the layout assumption is one line, and a helper around it would be indirection without a second caller.

**Not taken — skipping when `uv` is absent.** This is the one change that would defeat the PR. A licence gate that silently skips itself is not a gate, and this repo has been burned by precisely that failure three times (a CI step *named* "95% coverage gate" that enforced 90; a cancelled run read as green; a review bot reporting pass without fetching the diff). `pytest.skip` converts "we cannot verify what we publish" into a green tick — the exact silence this PR exists to remove. `uv run pytest` is the repo's paved road (`CLAUDE.md`, `.claude/rules/python-style.md`) and every CI job installs uv via `setup-uv`, so the failure mode is a misconfigured dev env, which should be loud. The `pytest.fail` message says so and names the fix.

PR #188, opened 2026-07-17
refactor(report): extract the HTML/SVG render seam out of langres.core

fxd24:
Moves the HTML/SVG rendering slice out of `langres.core` into a new `langres/report/` package. All numbers below are **measured on this branch**, not quoted from the brief.

## Why `report`, and why these three

`_report_html` had **zero** importers in `core` — all 8 were `langres.data.data_profile.*`. `_svg`'s only `core` importer was `eval_report` itself. These are shared rendering primitives that `langres.data` uses more than `core` ever did: in `core` by accident of history, not design. `core` is the contracts floor; a chart renderer is not a contract.

**On the name:** the brief invited me to argue for a better one. I'm not going to — `tools/refactor_target_packages.json` (which I'm forbidden to edit) **already maps all three to `"report"`**, alongside `fit_report` and `reports.py` for a later wave. Inventing a different name would desync from the gate file. `report` it is.

## Measured outcomes

| metric | before (`b70e54e`) | after |
|---|---|---|
| `core` lines | 30,343 | **28,962** (−1,381) |
| `core` % of `src` | 67.1% | **64.0%** |
| `report` lines | — | 1,414 |
| runtime SCC (`import_graph.py kinds`) | `[]` | **`[]`** |
| all-edges SCC | `[9, 9, 3, 2]` | **`[9, 9, 3, 2]`** (identical membership) |
| root `len(langres.__all__)` | 36 | **36** |
| `len(langres.core.__all__)` | 76 | **76** |

**Package edges — every direction one-way, as predicted:**

```
core   -> report:  0 edges ( 0 runtime)   <- the point of the exercise
report -> core  :  4 edges ( 4 runtime)
data   -> report: 13 edges (13 runtime)
report -> data  :  0 edges ( 0 runtime)
```

Runtime import cycles stay at **ZERO**. The tangle-gate baselines (`largest_scc=0, tangled=frozenset()` runtime; `9` all-edges) are **untouched** — none of the three modules was ever in an SCC. No baseline raised, no re-export relay interposed.

**Tests:** 3,641 passed, 4 failed — all 4 are shared gate files I am not allowed to edit (below). Ruff clean. Mypy: 9 errors in 5 files, **identical before and after** (missing optional extras: torch/sentence_transformers/openai/fastembed) — pre-existing, none in `report/`.

## Correction to the brief's API framing

The brief says to check `len(langres.core.__all__) == 76` before/after. Worth stating plainly: **`EvalReport` was never in `core.__all__`.** Verified:

```
EvalReport in root __all__: True     (langres/_exports/_flywheel.py, lazy)
EvalReport in core __all__: False
EvalReport in eval __all__: True     (langres/eval.py, lazy)
```

So `core.__all__` is structurally unaffected by this move. The only real API wiring was two lazy-map strings (`_flywheel.py:51`, `eval.py:82`) plus their TYPE_CHECKING imports. Both `langres.EvalReport` and `langres.eval.EvalReport` resolve to the identical class and still cache on first access.

## Breaking: the deep import path (needs your call)

`README.md:259` advertised `from langres.core.eval_report import EvalReport`. The brief's API constraint named `langres.eval` as the path that must keep working but did not mention this one — I suspect it went unnoticed. That path **is broken** by this PR.

I updated README, `docs/reference/eval-report.md` (mkdocstrings target), and `examples/quickstart_eval.py` in this same change. 0.x semver permits it, and the repo has precedent (the judge→Matcher rename). **If you would rather not break it**, the alternative is a PEP 562 `__getattr__` shim at `core/eval_report.py` — that keeps `core -> report` at 0 *runtime* edges (function-local only), exactly the pattern `clients/openrouter.py` already uses to forward `spend`/`spend_cap`. Say the word and I will add it; I left it out because the mapping file maps the module wholesale to `report` with no shim entry, unlike `model_ref` which lists both paths.

`CHANGELOG.md` 0.3.0 entries still name `langres.core.eval_report` — **deliberately left**: they are a historical record of what shipped at that path, not live docs.

## Gate files — 4 failures, all yours to land

I did not touch any file on the forbidden list. The move needs these:

**1. `tests/test_import_budget.py` — 3 hard failures + 1 silent fail-open**

| line | current | needs |
|---|---|---|
| 264 | `import langres.core.eval_report; import langres.core._svg` | `langres.report.…` |
| 297 | `import langres.core._report_html` | `langres.report._report_html` |
| 670 | `from langres.core.eval_report import EvalReport` | `langres.report.eval_report` |
| **391** | `_ROOT_LAZY_MODULES = ["langres.core.eval_report", …]` | `langres.report.eval_report` |

**Line 391 is the dangerous one — it does not fail. It passes vacuously.** It asserts the module is *absent* from `sys.modules` after `import langres`; a module that no longer exists can never appear, so the gate goes green while checking nothing for `EvalReport`. Proven:

```
langres.core.eval_report            module exists? False   <- gate still checks this
langres.core.benchmark              module exists? True
langres.core.calibration            module exists? True
langres.data.data_profile           module exists? True
```

That is the "deny lists rot open" / gate-decoupled-from-what-it-checks pattern. The other three tests fail loudly; this one needs a human to notice.

For the record, **all four properties that gate protects still hold at the new paths** (verified in fresh processes): bare `import langres` pulls no `langres.report.*`; `import langres.eval` stays lazy; `report.eval_report`+`_svg` leak no heavy dep; `data_profile`+`report._report_html` leak no heavy dep. **matplotlib does not become eager.** Only the path strings are stale.

**2. `tools/refactor_target_packages.json` — mapping totality**

`tests/test_import_graph.py::test_shipped_refactor_mapping_covers_every_module` ("no wave may discover a homeless file") fails: `langres.report`, `.report._svg`, `.report._report_html`, `.report.eval_report` are unmapped. The fix is the same one-liner shape as your `b70e54e`:

```json
"prefix": { "langres.report": "report", … }
```

…and the three now-stale `exact` keys (`langres.core._report_html`, `._svg`, `.eval_report`) can be dropped — nothing tests for stale keys, only for totality, so they would rot silently.

**3. `.github/workflows/test.yml:213-214` — the coverage glob** (you said you would handle it)

`--include="src/langres/core/*"` silently stops covering the moved files. I measured the impact rather than just flagging it. The three files carry **98–100%** coverage (`_report_html` 100%, `_svg` 98%, `eval_report` 99%), so they were *lifting* the core average — removing them moves the gate **down**:

```
baseline b70e54e : 97.91%  (7705 stmts)
after move       : 97.85%  (7307 stmts)   delta -0.06pp
```

Apples-to-apples, identical command. Both sit under 98 because the **fast** suite is not CI's `test-full` (which adds slow tests); the *delta* is the meaningful figure. Recommend adding `src/langres/report/*` to a gate so ~1,400 lines at 99% do not fall out of coverage entirely.

## Pre-existing defect found — reported, NOT fixed (out of scope)

`report/_report_html.py:safe_auc` returns `None` and filters non-finite values; `report/eval_report.py:_safe_auc` returns `nan` and does not. Two functions, same job, divergent contracts — now visible as neighbours in one package. **I did not unify them**: silently changing either is a behaviour change. Follow-up.

## Notes

- `report/__init__.py` is deliberately **import-free** — but see the correction below for why.

### Correction to my own claim (measured, and I was wrong)

An earlier version of this PR body claimed the new path is *"lighter than before"*. **That is false, and I am striking it.** Measured on both trees:

```
BASE  (4679489):  from langres.core   import _svg  -> 69 langres modules
MINE  (b7b8efe):  from langres.report import _svg  -> 70 langres modules
```

It is **+1 module heavier**, not lighter — the extra being `langres.report` itself (an empty package `__init__`). My reasoning was that `from langres.core import _svg` executes `core/__init__`'s fragment aggregator and the new path would skip it. Wrong: `langres/__init__.py` eagerly imports `langres.core`, so that aggregator runs no matter which langres submodule you touch. Nothing is saved.

The import-free `__init__` is still the right call, but for the *other* reason, which does measure out — re-exporting `EvalReport` there would add **+3** modules to every `data_profile` import of `_svg`:

```
_svg alone     : 70 modules
+ eval_report  : 73 modules  (+3)
extra pulled   : ['langres.core.benchmark', 'langres.core.usage', 'langres.report.eval_report']
```

So: keep it import-free to avoid +3 on the `$0` profile path; do not credit this PR with an import-weight win it does not deliver.
- Tests moved with the code to `tests/report/` (found by content — the tree does not mirror source).
- Rebased onto `b70e54e`; clean, no overlap with your gate commits.


fxd24:
### Follow-up detail on the `safe_auc` / `_safe_auc` collision (still NOT fixed here)

I re-verified this rather than take the brief's word for it, and it is **sharper than "divergent contracts"** — `_safe_auc` does not honour its own name or docstring.

`report/eval_report.py:694` documents itself as *"ROC-AUC that returns `nan` (not a raise) on an empty/degenerate input"*. It **raises** on a non-finite score:

```
non-finite score present  (labels=[T,F,T,F], scores=[0.9, 0.1, nan, 0.2]):
  _report_html.safe_auc ->  1.0                     # drops the nan pair, renders
  eval_report._safe_auc  -> RAISES ValueError: scores must be finite, got nan at index 2

empty input:
  _report_html.safe_auc ->  None
  eval_report._safe_auc  ->  nan
```

Three axes of divergence, not one:

| | `_report_html.safe_auc` | `eval_report._safe_auc` |
|---|---|---|
| return type | `float \| None` | `float` |
| empty input | `None` | `nan` |
| **non-finite score** | **filtered, returns a number** | **raises `ValueError`** |

The third is the real bug: a non-finite score means a broken ranking, which is precisely when you want the tearsheet to still render a `"n/a"` rather than blow up mid-report. `_safe_auc` only guards `if not scores`, so its "safe" is empty-input-only.

**Still deliberately not fixing it here** — out of scope, and either direction is a behaviour change (`EvalReport.from_log` on a judge that emitted a nan score currently raises; making it render `n/a` changes what users see). Flagging it with evidence so the follow-up can be decided on facts rather than on the two functions' names.

fxd24:
### CI status + rebase note

Rebased onto `4679489` (the base moved twice while this was in flight — `b70e54e` then `4679489`). Re-verified on the new base: `core` **28,962** lines, runtime SCC **`[]`**, all-edges SCC **`[9, 9, 3, 2]`**, root `__all__` **36**, `core.__all__` **76**. Unchanged.

**Diff touches 26 files and ZERO files on the forbidden list** (verified with `git diff --name-only`).

| check | result |
|---|---|
| **Docs** (mkdocs `--strict`) | **pass** — the retargeted `::: langres.report.eval_report` mkdocstrings reference resolves |
| **Lint** | **pass** — the gate for moved symbols / stale re-exports |
| **build** | **pass** |
| **test-finetune** | **pass** |
| **test-core-only** | **fail — `tests/test_import_budget.py` only** |
| test-full | skipped (push-only) |

`test-core-only` fails at exactly one step:

```
success  Install dependencies (core only -- no [semantic]/[llm] extras)
failure  Run import-budget tests against the real core-only install   <-- the gate file
skipped  Verify the core-only FuzzyString().dedupe() path still works
skipped  Verify the core-only eval tearsheet ($0 EvalReport -> to_html) works
skipped  Verify the core-only flywheel loop (examples/flywheel_min.py) works
```

The three example steps were **skipped, not failed** — they never ran because the gate step aborted the job. I ran all three locally on this branch and **all three pass**, including `quickstart_eval.py` (which I retargeted) rendering its tearsheet, and `flywheel_min.py` resolving `EvalReport` through the root lazy path. They will go green once the four `test_import_budget.py` paths are updated.

**Wheel verified** (a new package silently omitted from packaging would be invisible): hatchling auto-discovers via `[tool.hatch.build].exclude`, and the built wheel contains

```
langres/report/__init__.py
langres/report/_report_html.py
langres/report/_svg.py
langres/report/eval_report.py
stale core render modules still shipping? NONE
```

Nothing here needs a fix from me — the only red is the gate file you own.

fxd24:
### Final CI — all checks landed

`test (3.13)`: **4 failed, 3639 passed**. The four are *exactly* the four predicted in the PR body, and every one is a shared gate file I am not permitted to edit:

```
FAILED tests/test_import_budget.py::test_eval_report_stays_import_light
FAILED tests/test_import_budget.py::test_data_profile_stays_import_light
FAILED tests/test_import_budget.py::TestRootLazyGetattr::test_eval_report_resolves_and_caches
FAILED tests/test_import_graph.py::test_shipped_refactor_mapping_covers_every_module
```

Zero unexpected failures. Nothing in `src/` or the moved tests is red.

| check | result |
|---|---|
| Docs (mkdocs `--strict`) | pass |
| lint | pass |
| build | pass |
| test-finetune | pass |
| test (3.13) | fail — 4 gate tests only |
| test-core-only | fail — `test_import_budget.py` step only |
| test-full | skipped (push-only) |

**This PR cannot go green without you** — the four fixes live in files you own and update at integration (the exact deltas are in the PR body). Everything on my side is done.

⚠️ **`Sourcery review` reports `pass` but reviewed nothing.** Its actual output:

> Sorry @fxd24, you have reached your weekly rate limit of 2500000 diff characters.

33s, green check, zero review. That is a *new* variant of the Sourcery false-green (not the ~5s never-fetched-the-diff case) — a quota exhaustion that still reports success. **This PR has had no Sourcery review**; do not read that check as one. Given how many waves are landing this week, it is likely silently green on the others too.

PR #191, opened 2026-07-17
refactor(core): thin langres.core 67% -> 62% of src, and cut a 9-module import knot

fxd24:
Thins `langres.core` from **67.1% → 62.0%** of the source tree, and cuts one of the two 9-module import knots. Four sub-PRs (#187, #188, #189, #190), each CI-verified and reviewed before merge.

## Measured on this branch — not reported, measured

| | before (`d63443a`) | after |
|---|---|---|
| `core` | 30,343 lines (67.1% of src) | **28,223 (62.0%)** |
| `core/resolver.py` | 2,330 — largest file in the repo | **1,258 (−46%)** |
| all-edges SCCs | `[9, 9, 3, 2]` — tangled **23** | **`[9, 3, 2, 2, 2]` — tangled 18** |
| runtime SCCs | `[]` | **`[]`** |
| `langres.__all__` / `langres.core.__all__` | 36 / 76 | **36 / 76** |

Counted by **SCC membership**, never mutual pairs — this repo has measured that a re-export relay shrinks the pair count while *growing* the SCC (`tests/test_import_tangle.py`, PR #169). **No shims were interposed anywhere.**

## What landed

- **#189 — `ERModel` split** (`resolver.py` 2,330 → 1,258). A 1,975-line class becomes `_model_state` (what a model *is*) ← `_model_run` (how it runs) / `_model_persist` (how it persists), over an `_artifacts` leaf. Verified by AST diff: 46 of 51 functions byte-identical, every symbol has a home, `save`/`load` round-trip proven live at $0 on a `FuzzyString`.
- **#188 — the render seam → `langres.report`.** `_svg` (a pure leaf), `_report_html` (all 8 importers were `langres.data`, zero from core), `eval_report`. `core → report` = **0**, one-way. `reports.py`'s value types stayed in `core` — they're contract surface.
- **#187 — the autoresearch engine → `langres.optimize`.** Zero backward edges of any kind; a pure move.
- **#190 — `CostTrack` → the `core.usage` leaf.** The one that mattered: `clients/openrouter.py` (an HTTP client) imported `CostTrack` from a 1,769-line benchmark harness — the floor importing the ceiling. Both edges were non-toplevel, which is why runtime read 0 while all-edges carried a 9-knot. Deleting it freed `openrouter`, `finetune`, `matchers.cascade`, `matchers.llm_judge`, `method_registry` — **tangled 23 → 18**.

## Breaking — deep imports only, no shims

Every symbol still exists; every **supported** path is unchanged (`langres.EvalReport`, `langres.eval`, `langres.optimize()`, `score_blocking`). Only deep `langres.core.*` paths break, and none were ever in an `__all__`. Logged in `CHANGELOG.md`. No compatibility shims: a relay would re-create the exact edges these moves remove.

- `langres.core.eval_report` → `langres.report.eval_report`
- `langres.core.optimizers` → `langres.optimize.blocker_optimizer`

## Gate work

- **`lint.yml` ran on none of this at first.** `pull_request.branches` filters by *target* branch; `feat/core-thin` wasn't listed, so every sub-PR would have merged with **no ruff and no mypy** — the one gate that catches what moving code breaks. Fixed, reverted here.
- **A gate was guarding nothing.** `_ROOT_LAZY_MODULES` is a *deny* list: it asserts modules are absent from `sys.modules`, so its stale `langres.core.eval_report` entry passed trivially against a module that no longer existed. Green, checking nothing. Now enforced by `test_root_lazy_module_paths_all_exist` — verified by injecting a dead path and confirming red. **A deny list rots open; this one now rots closed.**
- **Coverage glob** extended to `report/*` + `optimize/*`. Both measure ~98–100%, so they were *lifting* core's average — dropping them would have un-gated well-covered public surface while making the number look better. Verified against coverage's real `GlobMatcher`: 95 files matched, no leakage.
- **lint concurrency SHA-keyed** (kept): grouping on the ref alone means merging one PR cancels another's in-flight run, and cancelled ≠ green.

## Two things worth your attention

1. **Sourcery reviewed almost none of this.** Its weekly 2.5M diff-char quota is exhausted, so it reports `pass` having read nothing. #187 got a real review (47s); #188 a rate-limited no-op (**42s**) — five seconds apart, opposite truths. Every PR here got a real substitute review instead. This likely affected other PRs this week.
2. **The coverage ratchet is quietly tightening.** `core/*` measures **98.43%** against a floor of **98** that was set from a measured **98.84%**. ~0.4pp of headroom left. The next person to add a few uncovered lines to `core/` will trip it and reasonably think they broke something. Recorded in the workflow + testing rule; a real follow-up, not fixed here.

## Not done, deliberately

- **The fit family (645 lines) stayed in `resolver.py`.** Extracting it measures `tangled 23→24` — it would spread the existing knot across more modules, which is the PR #169 shape the gate exists to catch. Refused on measurement, not preference.
- **The surviving 9-module SCC is entirely `data.data_profile.*`** — self-contained, unrelated to core, untouched.
- `_report_html.safe_auc` vs `eval_report._safe_auc` diverge on **three** axes (one returns `None` and filters non-finite; the other is *documented* as returning nan but actually **raises**). Now colocated and tempting to unify — **don't** silently; it's a behaviour change. Reported, not fixed.


fxd24:
One diff line worth explaining before you hit it in review.

`.github/workflows/test.yml` loses `feat/arch-refactor` from its `push.branches` gate list. That is **not** part of this refactor — it is dead config I removed while reverting my own temporary `feat/core-thin` entry from the same list, and I only kept it removed after verifying the branch is a strict ancestor of `main`:

```
git merge-base --is-ancestor origin/feat/arch-refactor origin/main   # true
gh pr list --base feat/arch-refactor --state open                     # 0
```

Fully merged, nothing targeting it. The list entry could only ever fire on a branch that can no longer diverge. The file's own comment says to revert to `[main]` "once the work has landed and the integration branch is deleted" — landed is true, deleted is not, so this is the one clause I am not honouring to the letter. Say the word and I will put it back; I judged dead config in a gate list to be worse than an undeleted branch.

Same applies to `feat/core-thin` itself: **do not delete this branch until #191 is merged**, or the PR loses its head.

Issue #193, opened 2026-07-17
Epic: the langres core — one carrier, one op, topology as code


fxd24:
## Why

**langres should be to entity resolution what PyTorch is to machine learning:** you build the model, and architectures the authors never imagined are expressible without touching the core.

Today it can't be, for one specific reason. `Blocker`, `Comparator`, `Matcher`, `Clusterer` are names *we invented* for positions in a pipeline *we chose* — and they live in the type system. The four components have four different types on the arrows:

```
records → candidates → vectors → judgements → clusters
```

Every arrow lines up exactly one way. That isn't a pipeline we picked; **it's the only pipeline the types permit.**

`nn.Module` composes for exactly one reason: `forward` is `Tensor → Tensor`. Input type equals output type, so any module can follow any module. PyTorch has no `nn.Encoder` base class — a transformer's encoder and decoder are both just `Module`, and `nn.Linear` doesn't know it's layer three. **Position lives in `forward()`, never in the type.** That is precisely why PyTorch could host ResNet, transformers, and diffusion without a core change.

## What we found

**Every defect traces back to one of the four names.** All verified against `origin/main`:

| The name | What it forced | Where |
|---|---|---|
| `Matcher` as a mandatory slot | `EmbeddingScoreMatcher` — a component that computes nothing, copying `similarity_score` → `score` | `core/matchers/embedding_score.py` |
| `Blocker` owning the score | a `type_name == "vector_blocker"` sniff walking the blocker tree on **every** `resolve()` | `core/resolver.py:219` |
| `Clusterer` owning the threshold | `t` passed twice in one topology; closure overrides negatives we already measured | `architectures/vector_llm_cascade.py` |
| Fixed order, each slot once | CLER's matcher-aware blocking depth (PVLDB 17(3)) is **inexpressible** | `core/resolver.py:585` |
| Set-wise work in a pairwise slot | `GroupwiseMatcher` is set-wise **in**, pairwise **out** — the joint decision is flattened away | `core/matcher.py:217` |
| Position in the type | "tune the blocker's embedder" ≠ "tune the matcher's embedder" — two code paths, one operation | — |

**And the four types are already one type.** From `core/models.py`:

```
ERCandidate       = left,    right,     blocker_name,   similarity_score: float|None
PairwiseJudgement = left_id, right_id,  decision_step,  score: float|None, score_type
```

A pair, a score, and a note about who touched it last — twice, in two spellings. `EmbeddingScoreMatcher` is the memcpy between them.

[...]

PR #211, opened 2026-07-18
W3-c: reify the scoring prefix into _stages() + a generic fold driver — #193

fxd24:
## What

W3-c of epic #193 (the "one carrier, one op" core refactor). A **byte-parity mechanical refactor** of the scoring spine in `src/langres/core/_model_run.py` — no behavior change.

Consolidates the scattered inline Op-adapter construction into **one per-call factory** + **one generic driver**:

- **`_stages(self, *, log=None) -> list[Source[Any] | Score[Any]]`** — the ordered scoring pipeline built from the four slots: `[BlockerSource(self.blocker)]` + (`ComparatorScore(self.comparator)` if configured) + `MatcherScore(self._scorer(log=log), out_space="unknown")` (reuses the existing `_matcher_score`, keeping one MatcherScore construction site).
- **`_run_stages(records, stages)`** — a small module-level driver that folds `.forward` across the stage list (`Records → Pairs → … → Pairs`). Two narrowing `cast`s encode the Source-first shape `_stages` guarantees (the first stage takes records, the rest take pairs — they cannot share one loosely-typed loop variable); no `Any` accumulator.
- `_candidates` folds `_stages()` **minus** its trailing `MatcherScore` (block+compare prefix); `_scored_pairs` folds the **full** `_stages`. `_ensure_index_built` still runs first in both, outside the fold.

## Honesty-only sentinel swap

`out_space="heuristic"` → `out_space="unknown"` (the W3-b sentinel). A `Matcher` has no class-level `score_type` — each row is stamped its own family per-judgement in `MatcherScore._rescore` — so the family is genuinely not knowable pre-run. `out_space` is **inert in the run path** (nothing reads `Score.out_space`; `_rescore` stamps the row's own family), so there is **no behavioral effect**. Updated the one asserting test (`test_matcher_score_is_a_score_with_declared_out_space`); the other `test_op_adapters` uses of `"heuristic"` are non-asserting placeholders over *known* rapidfuzz matchers whose real family IS `heuristic`, left as-is.

## Guarantees preserved

- **Per-call rebuild, never cached** — `_stages` takes `log` and wraps `self._scorer(log=log)` per call, so `fit()`'s reassigned matcher and the per-call log are honoured (a cached stage list would pin a stale matcher / empty the log).
- **SpendCappedMatcher(LoggingMatcher(raw)) wrapping intact** — `_stages` wraps the same `self._scorer(log=log)`; `MatcherScore._rescore` drains the capped generator, so the one-call overshoot bound / `$0`-next-resolve / `partial_judgements` all survive.
- **`compare()` untouched** — still terminates at scoring via `_matcher_score` and raises `MatcherAbstainedError` on abstain; never routes through the clusterer.
- Calibration, `ClustererStage`, `CanonicalizeFinalize`, the abstain guard: untouched.
- No new source module (no `tools/refactor_target_packages.json` change).

## Verification (in worktree)

- Parity gates green: `test_behavior_parity_w0` (FuzzyString golden), `test_behavior_parity_cascade_w3d` (mixed-family cascade golden), `test_groupwise_spine_w3a` (predict() deterministic order); plus `test_op_adapters` (23), `test_op`, `test_resolver_spend_cap` (the `.module.forward` AST-ban), import budget/tangle gates.
- **Full fast suite: 3773 passed, 2 skipped** (`--all-extras`, OpenMP guard).
- `mypy --strict` clean across all 89 core files; `ruff format --check` + `ruff check` clean.
- $0 end-to-end: `FuzzyString().dedupe(...)` clusters `{1,2}`/`{3,4}`, drops the singleton; `DedupeResult.score_type == "heuristic"` (the row's real family, proving the `"unknown"` out_space is inert); `resolve()` matches, `compare(1,2)=True`.

https://claude.ai/code/session_011XYuhyiigi1Bw2YnWCxMr9

### DIFF
--- a/src/langres/core/_model_run.py
+++ b/src/langres/core/_model_run.py
@@ -130,11 +130,13 @@ def _run_stages(records: list[Any], stages: list[Source[Any] | Score[Any]]) -> P
     """Fold ``forward`` across ``stages`` to turn ``records`` into a scored ``Pairs``.
 
     The generic driver behind :meth:`ModelRun._candidates` and
-    :meth:`ModelRun._scored_pairs`. A stage list :meth:`ModelRun._stages` builds
-    always leads with the :class:`~langres.core.op.Source` (records -> pairs) and
-    continues with :class:`~langres.core.op.Score`\\ s (pairs -> pairs), so the
-    fold walks ``Records -> Pairs -> ... -> Pairs``. The two narrowing ``cast``\\ s
-    encode that Source-first shape (which ``_stages`` guarantees) precisely --
+    :meth:`ModelRun._scored_pairs`. Both stage lists it is handed --
+    :meth:`ModelRun._stages` (full) and its matcher-free prefix
+    :meth:`ModelRun._block_compare_stages` -- lead with the
+    :class:`~langres.core.op.Source` (records -> pairs) and continue with
+    :class:`~langres.core.op.Score`\\ s (pairs -> pairs), so the fold walks
+    ``Records -> Pairs -> ... -> Pairs``. The two narrowing ``cast``\\ s
+    encode that Source-first shape (which the factories guarantee) precisely --
     ``Source.forward`` takes records while ``Score.forward`` takes a ``Pairs``, so
     the heterogeneous first stage cannot share one loosely-typed loop variable
     with the rest, and the cast is a typed assertion of the shape, not an ``Any``.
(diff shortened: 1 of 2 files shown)


fxd24:
Addressed review (decoupling refinement, `25bfe32`): `_candidates` no longer builds the full `_stages()` and discards the matcher — that allocated a throwaway `SpendCappedMatcher(LoggingMatcher)` on every `candidates()` call (hot fit/eval path) and widened `_candidates`'s dependency to `self.module` + `self.clusterer.threshold`. Extracted the matcher-free `_block_compare_stages()` (BlockerSource + optional ComparatorScore, blocker/comparator slots only); `_candidates` folds it directly and `_stages()` appends the MatcherScore — one construction site, no duplication, and the blocking op no longer mandates a scoring slot (keeps matcher-free/retrieval-only topologies expressible, per the epic goal). Byte-identical stage list; pure decoupling. Re-verified: 3 parity gates + test_op_adapters + test_resolver_spend_cap green, full fast suite 3773 passed / 2 skipped, mypy --strict + ruff clean.

PR #236, opened 2026-07-27
docs: make the docs true (fine-tuning ships, PyPI/licence resolved, a first command that runs)

fxd24:
Docs-only. Every claim here was checked against the code or a command **before** being written, because the whole reason this pass exists is that claims stopped matching reality. Evidence is named inline so a reviewer can re-check cheaply.

## The headline fix

`README.md` claimed **fine-tuning a small LM was `🚧 roadmap`. It is not — it ships.** `src/langres/training/finetune.py` has `run_finetune()`, `finetune()`, `class QLoRA(Method)` and `class QLoRATrainer`; `QLoRA`/`run_finetune`/`finetune` resolve off the root `langres` namespace (verified at runtime, not by reading `_exports`); there is an example (`examples/finetune_capstone.py`) and two test modules. This one line had actively misled the project's own maintainer into believing the feature was missing.

It is now stated as shipped **with the honest limitation attached**: `QLoRATrainer.train` is `# pragma: no cover` (`finetune.py:346`) because real training needs a GPU — the fast suite drives the orchestration with a fake trainer, and the real stack runs in the dedicated `test-finetune` job (`test.yml:298`). The same false claim appeared a second time in the flywheel prose; both are fixed.

## A first command that actually runs

`README.md` and `docs/index.md` both led with `examples/research/first_experiment.py`. A reader cannot run it:

- `examples/` is **not in the wheel** — verified by building it: the wheel contains only `langres/` and the dist-info, zero `examples/` entries.
- research `Retrieve` constructs a Qdrant index (`resources/retrieve.py:13,61`) and `qdrant-client` is in `[semantic]` (`pyproject.toml:62`), not core `dependencies`.

Measured both ends: on a **bare core-only install** the script does not error loudly, it reports a silently `failed` cell with no reason; with `[semantic]` it completes and prints `pair_f1 = 0.0000` (a `FakeEmbedder`).

Both pages now lead with `examples/quickstart_models.py`, verified end-to-end in a throwaway core-only venv (zero of torch/litellm/faiss/scikit-learn/qdrant/sentence-transformers installed) where it prints real clusters — the same script the core-only CI job runs on a bare `uv sync` (`test.yml:106`). The research runner is kept, with its real prerequisites stated rather than implied.

## TODOS.md

- **"PyPI publish — not published today"** and **"License choice (TBD)"** were both false. `pypi.org/pypi/langres/json` reports **0.3.0 under Apache-2.0** (releases 0.1.0, 0.3.0), and `pyproject.toml:6` is `license = "Apache-2.0"`.
- Dropped the `CascadeModule` → `CascadeJudge` migration item: **neither name exists anywhere in `src/`** any more; `methods.py` uses `CascadeChainMatcher`.
- Records the three deferred cleanups with **measured** counts so the follow-up starts from evidence, not a re-survey.

## docs/RESOURCES.md — six false claims

Mostly find-and-replace collateral from the `judge` → `Matcher` rename: **`nn.Matcher`** (PyTorch's class is `nn.Module`); a *"two-layer API (tasks vs. core)"* that was never built; `Flows`/`Tasks`/`Optimizer` presented as existing components; and DSPy / sentence-transformers / optuna each called a **"core dependency"** when they are `[llm]`, `[semantic]`, and the **dev-only** group respectively (optuna is not reachable via any `pip install langres[...]`).

## AGENTS.md, mkdocs.yml

- `plotting/` was undocumented. Now documented — including that **`matplotlib` is declared nowhere in `pyproject.toml`** (no extra, not even the dev group), so `langres.plotting` requires callers to bring their own.
- Corrected a stale tree entry: `module.py, modules/, judges/` → `matcher.py, matchers/` (the former paths do not exist).
- `docs/RESOURCE_API.md` was in neither `nav` nor `not_in_nav`, so mkdocs reported it as unlisted; added to the nav.

## Deliberately NOT done

- **No source file deleted or moved.** The 21 shims, 71 examples and the `bootstrap/` question are counted and recorded only.
- `docs/DOCS.md` left alone — it is **already** excluded from the site (`mkdocs.yml:74-78`), so retiring it would be zero user-visible change.
- `docs/RESOURCES.md` stays in `not_in_nav`: it is a bibliography of external reading, not the resources-concept reference. `RESOURCE_API.md` is that reference.
- `CLAUDE.md`/`.claude/**` untouched — a parallel unmerged commit rewrites them.

## Verification

`4262 passed, 2 skipped` (identical pass count to main; docs-only diff), `ruff format --check` + `ruff check` clean, `mypy src` clean, `mkdocs build --strict` clean, and all three core-only smoke scripts run.

## Known, not fixed here

`docs/GETTING_STARTED.md:319` still says *"Fine-tuning a small LM on these labels is the roadmap's next rung"* — the same false claim as the README row. That file is owned by a parallel stream, so it is flagged rather than edited to avoid a collision.

## Summary by Sourcery

Align documentation and TODOs with the current shipped feature set, installation paths, and package metadata, ensuring the first-run examples and fine-tuning claims match reality.

Documentation:
- Update README and docs index to provide a core-only, runnable quickstart script and clarify that example scripts require a git checkout and, for the research runner, the semantic extra.
- Correct documentation about fine-tuning, stating that QLoRA-based fine-tuning is shipped with a noted testing caveat and specifying the finetune extra and example entry point.
- Fix outdated statements in RESOURCES.md about PyTorch abstractions, langres architecture layers, and dependencies for DSPy, sentence-transformers, and Optuna, reflecting their actual roles and extras.
- Refresh AGENTS.md module tree to describe the matcher-based core, note back-compat shims, and document the optional plotting module and its matplotlib requirement.
- Update mkdocs navigation to include the Model Resources (RESOURCE_API.md) page and keep the RESOURCES bibliography out of the main nav.

Chores:
- Remove outdated TODO items about PyPI publication, license choice, and a non-existent CascadeModule→CascadeJudge migration, and add a measured inventory of deferred cleanups for back-compat shims, example scripts, and the phantom bootstrap package.


fxd24:
### Second wave: the clean-install doc gate

Stream D's gate (`tools/doc_snippets.py` — build the real wheel, install it into a fresh extras-free venv, execute every ` ```python ` block with that interpreter, cumulatively per page) reported **19 failures** across the three landing docs. I ran it locally against my branch and fixed **every failure in the two files this branch owns**.

`README.md` and `docs/index.md` are now **zero-failure**, with 5 python blocks genuinely executing against a real clean-install wheel:

```
PASS  README.md:40      python      <- new, runs on a bare `pip install`
PASS  README.md:228     python
PASS  README.md:248     python
PASS  README.md:293     python
PASS  docs/index.md:21  python      <- new, runs on a bare `pip install`
SKIP  ... requires-repo x4, requires-network x1, illustrative x1
```

**Rewrote rather than exempted on the landing path.** Both pages now lead with `pip install langres` plus an inline snippet that actually runs on a bare core install — verified against a throwaway core-only venv (zero of torch/litellm/faiss/sklearn/qdrant present), where it prints `DedupeResult([{'1', '2'}], architecture='FuzzyString', backbone=None, score_type='heuristic', threshold=0.6)`. The `examples/` script form is kept directly below for git-checkout readers. An exemption marker is honest, but a README whose first snippet is `requires-repo` still fails the reader — it just fails them legibly.

**Exemptions use the narrowest true reason.** Worth flagging one judgement call: the `VectorLLMCascade` block fails first on `ModuleNotFoundError: faiss`, so `requires-extra=semantic` looks right — but it is **wrong**, because the block calls `.dedupe()`, so installing the extra still would not make it run; it needs a key and a real paid call. It is marked `requires-network`. The marker should name the strongest blocker, not the first traceback.

**A trap worth recording for whoever fixes the rest:** the gate's `examples/...` rule runs *before* the exemption check and is satisfied **only** by `requires-repo`. So naming an `examples/*.py` path inside a block marked `illustrative` still fails. I hit this by putting a helpful `examples/quickstart_eval.py` reference inside a code comment; the comment now points at the linked script instead of repeating its path.

**Not fixed here:** the remaining 13 failures are all in `docs/GETTING_STARTED.md`, which a parallel stream owns — 5 `examples/` path references, 4 needing `[semantic]`, 2 needing `[trained]`, and 2 fragments (`student`, `resolver`) that are downstream of the extras-only blocks and so will **not** go green just by marking B and C.

fxd24:
@codex review

Re-requesting on the final head (4bd0dc9) — the earlier automated review covered f37cd65 and several markdown commits have landed since. This PR is docs-only; the highest-value check is whether any factual claim it asserts is contradicted by the code.

PR #241, opened 2026-07-27
feat(fit): fit() derives the match threshold from labels -- and races it against the incumbent

fxd24:
## The gap this closes

`derive_threshold` was **not** "wired to nothing" — it already has two production
callers (`curation/harvest.py`, whose module docstring calls itself "its first
production caller", and `data/fixed_split_pair_benchmark.py:309`). The real gap
is narrower and worse: **`ERModel.fit()` never called it.** A user who supplied
labels still resolved at whatever constant the constructor happened to set.

`fit(data, pairs=..., derive_threshold=True)` now fits the match cut from those
labels, reusing the existing `derive_threshold_from_pairs` seam rather than
re-implementing it.

## It races the derived cut and keeps the winner

Deriving a threshold from labels is **not** unconditionally an improvement, and
this PR has the measurement to prove it. `derive_threshold` maximizes Youden's J
(`tpr - fpr`), which ties flat across a whole separating gap. On a fixture where
the default `0.5` already separates cleanly, J returns a cut high enough to drop
a genuine twin:

| fixture | incumbent | derived | train pair-F1 | held-out pair-F1 | kept |
|---|---|---|---|---|---|
| default 0.5 is wrong (hard negatives share a token) | 0.5000 | 0.8824 | 0.5217 → **1.0000** | 0.5714 → **1.0000** | derived |
| default 0.5 already right (abbreviated twins, dissimilar negatives) | 0.5000 | 0.8824 | 1.0000 → 1.0000 | 1.0000 → **0.8000** | **incumbent** |

The second row reproduces at seeds 0/1/4/5. A derive-and-apply implementation —
which is what this PR was originally scoped as — would have shipped that
regression as an improvement.

So `fit()` **races** the derived cut against the threshold already in force and
keeps whichever wins, recording both in `FitReport.threshold_fit`.

## Which split did the selecting (the leakage question)

**Selection runs on `train`. `valid` is never read to decide anything.**

`train` was already spent deriving the candidate, so selecting there adds no
leakage that deriving had not already incurred — and it leaves *both*
`held_out_f1` numbers in the report as clean estimates of a split nothing ever
tuned against. The honest before/after of a threshold fit is
`previous.held_out_f1` vs `candidate.held_out_f1`, and both held-out columns
above are exactly that, not in-sample numbers.

`test_selection_never_reads_the_held_out_split` is the guarantee: re-fitting at
different seeds reshuffles which entity components land in `valid`, and the
chosen cut must not move with it.

**The residual gap, stated plainly:** there is no clean estimate *of the
selection itself*. Choosing between two cuts on `train` and then reporting
`valid` is honest about `valid`, but the selection decision has no untouched
third split behind it. A proper three-way split (derive / select / report) is
the fix and is named as a follow-up in `docs/EXPERIMENTS.md` — it is more than
this PR should carry.

When `split=None`, every labeled pair is train: `held_out` is `False`,
`held_out_f1` is `None`, and the Markdown digest prints the word **IN-SAMPLE**
in capitals rather than letting the number pass for held-out.

## New capability: `fit()` could not run on an `_ops` model at all before this

This is not a bug fix or a new branch on an existing path. On `main`,
`ERModel.fit()` reads `type(self.module).__name__` unconditionally as the first
statement of the fit body; `from_topology` "leaves the four slots empty" and the
`module` property raises `_require_bound`. So **every** `fit()` call on an
explicit Op topology raised before any fit logic ran, regardless of arguments.

`_fit_chain_threshold` is the new path. An Op chain has no matcher slot to train
(its `Score`s are topology, not a fittable slot) and no clusterer to read — its
one fittable parameter is the terminal `ThresholdSelect`. Calling `fit()` on such
a model without `derive_threshold=True` now raises a message that says exactly
that, instead of the misleading unbound-model error.

## Both threshold seams, and how each is tested

| topology | where the cut lives | test |
|---|---|---|
| classic four-slot | `clusterer.threshold` | `_classic()` fixture — derive, decline, persist, guard tests |
| explicit `_ops` chain | terminal `ThresholdSelect.threshold` | `_explicit()` fixture — `applied_to == "threshold_select"`, the no-`ThresholdSelect` error, `TopKSelect`-only chains |

The branch is not written twice. `_model_state.py` gains a topology-neutral trio
— `_threshold_seam()` / `_match_threshold()` / `_set_match_threshold()` — and the
reader and the writer are derived from one `_chain_threshold_select()` helper so
they cannot disagree about which `Select` is the cut.

One subtlety the `_ops` path had to get right: deriving from the *post-threshold*
rows would only recover the cut that already filtered the sample. So
`_prethreshold_pairs` runs the chain once **truncated at** the fitted
`ThresholdSelect`, and both splits are read off that single pass (a chain's
`Source` owns retrieval, so unlike a classic model it cannot score an arbitrary
candidate subset). That is one extra full pass — documented as a cost. See
finding #3 below for why it truncates rather than merely omitting the select.

## No silent fallback when scikit-learn is missing

`training/calibration.py` imports sklearn at module scope, so on a bare install
the lazy import inside `derive_threshold_from_pairs` used to surface as a raw
`ImportError`. It now raises a directed one naming the problem, the cause, and
the fix (`pip install 'langres[trained]'`), and says explicitly that langres will
**not** substitute an underived default — identical code must not produce
different artifacts depending on which extras happen to be installed.
`docs/GETTING_STARTED.md` already sells "No key, no silent fallback"; this keeps
that promise on the extras axis too.

Note `test-core-only` is *not* a gate for this: scikit-learn is duplicated in
`[dependency-groups] dev`, so `uv sync` installs it. The test simulates absence
via `monkeypatch.setitem(sys.modules, ...)`.

## Provenance, and what survives save/load

`FitReport.threshold_fit` is a new `ThresholdFit` carrying `source`
(`derived`/`declined`/`not_fitted`), `method`, `n_pairs`, `held_out`, `applied_to`
(`clusterer` or `threshold_select`), `selected_on`, and a `ThresholdCandidate`
for each of `previous` and `candidate` with that candidate's selection F1 and
held-out F1.

**`fit_report_` is deliberately never serialized** (`_model_state.py:160`). The
asymmetry is now stated in the code: `save()`/`load()` carry the *threshold
value* — it lives on the clusterer or the `ThresholdSelect`, both of which
persist — but **not** the provenance of how it was chosen. A loaded model
resolves at the derived cut and cannot tell you it was derived. Keep the
`FitReport` (it is JSON-serializable) if you need the audit trail.

## Youden's J is symmetric in cost; ER is not

Called out in the `fit()` docstring and in `docs/EXPERIMENTS.md` at the point of
choice: J weights a false positive and a false negative equally, but in entity
resolution a false merge can poison an entire cluster through transitive closure
while a false split is recoverable by a later merge. A user who cares about that
asymmetry should pass their own threshold, or call `derive_threshold` directly
with a method that prices it.

## The default stays `False`

`derive_threshold: bool = False`. The measured decline above is the reason: a
feature that sometimes helps and sometimes hurts does not belong on by default,
and flipping it would silently move the cut for every existing `fit(pairs=...)`
caller. The race makes the opt-in safe; it does not make it free.

## Scope notes

- `fit(pairs=..., derive_threshold=True)` is now accepted for matchers that are
  **not** `SupervisedFitMixin` (e.g. `WeightedAverageMatcher`) — previously
  `fit(pairs=...)` raised for those, even though a fixed scorer is exactly where
  threshold fitting is the *only* useful thing to fit. `fit(method=...)` and
  `derive_threshold=True` together are refused: two different fits of the same
  parameter.
- The walkthrough went in **`docs/EXPERIMENTS.md`, deliberately, not in
  `GETTING_STARTED.md`** — `tools/doc_snippets.py` executes python blocks from
  `README.md` / `docs/index.md` / `docs/GETTING_STARTED.md` only, against a bare
  wheel in a fresh venv. A `[trained]`-dependent snippet in that set would trip
  the clean-install gate. `EXPERIMENTS.md` is where `derive_threshold` is already
  documented, so the new section sits beside the primitive it wraps. This was a
  placement decision, not an oversight.
- No constructor default was touched: the six `threshold: float = 0.5` values in
  `src/langres/architectures/` are unchanged.

## Two P1s cross-review caught, and how each is now guarded

**1. The refusal ran *after* the spend.** The "no `ThresholdSelect` to write"
`ValueError` was raised inside `_select_threshold`, one full Source+body pass
downstream of `_prethreshold_pairs`. So `from_topology([BlockerSource,
MatcherScore(paid_llm), ClustererStage])` + `fit(derive_threshold=True)` billed
the entire dataset and then threw the scores away. **Measured: 435 scoring calls
over 30 records before the refusal.** The check is hoisted to the top of
`_fit_chain_threshold`; the guard inside `_select_threshold` stays, because it is
the invariant for the classic path too.

The existing test asserted only the exception — which passes whether the refusal
happens before or after the spend, i.e. a check decoupled from what it guards. It
now runs a counting matcher and asserts `calls == 0`.

**2. Fabricated held-out metrics on a Score-less chain.** The metrics gated on
`aligned.valid.candidates` while computing over `judgements`, which is filtered to
`score_type is not None`. A blocker similarity lands as an *unscored* row score
(`PairRow.score_type is None` — "blocked, not yet scored"), so
`[BlockerSource(sim_blocker), ThresholdSelect, ClustererStage]` is permitted
topology whose `judgements` are legitimately empty — and `classify_pairs([], gold,
t)` does not return `None`, it returns a real `PairMetrics` of zeros with
`fn=len(gold)`. That printed a fully-populated "Held-out pair metrics" table of
`0.0000` for a fit that measured nothing. Now gated on `judgements`, matching
`_described`'s own `if valid_judgements else None` convention.

**Both new assertions were confirmed to FAIL against the unfixed code** before
being kept — I reverted each fix, watched the test go red (435 calls; a
`PairMetrics(..., f1=0.0, fn=4)` where `None` was required), then restored it.

## Four more from codex review, all verified and fixed

Each was checked against source before touching anything —
`predicted_match`'s decision precedence, `classify_pairs`' use of it,
`_dedupe_explicit`'s `normalize_records` call, `PairRow`'s unscored-similarity
state. All four were real.

**3. A `Score` after the fitted `Select` contaminated the derivation.**
`_prethreshold_pairs` *removed* the fitted `ThresholdSelect` and ran everything
else. In the supported reranker topology — `Source -> ScoreA -> ThresholdSelect ->
ScoreB -> ClusterStage` — `ScoreB` then overwrote the score column, so the fit
derived a cut from `ScoreB` values and wrote it into a `Select` that thresholds
`ScoreA` values. It now **truncates at** the select instead of removing it. Test
puts a constant-0.01 scorer downstream and asserts both `calls == 0` and that the
derived cut did not collapse toward it.

**4. `fit()` on an explicit chain skipped `normalize_records`.** Every other door
into a chain Source (`_dedupe_explicit`, `execute`) normalizes first; the fit path
was handing it the raw dicts the caller passed to `fit()`.

**5. Every non-deriving pair-fit claimed the cut was a constructor default.**
`ThresholdFit(source="default")` was stamped on all `fit(pairs=...)` calls, so a
model whose *earlier* `fit(derive_threshold=True)` derived its cut had that cut
relabelled a default by the next plain fit — a false record in the one field
whose entire job is provenance. `fit_report_` is not serialized, so a later fit
cannot recover the earlier one's origin. Renamed to **`"not_fitted"`**, which
claims only what is true: *this* fit did not touch the threshold. The test asserts
the rendered digest never contains the word "default".

**6. A decider's threshold fit was silently inert.** `predicted_match` gives
`decision` precedence over `score`, so for a matcher emitting both, moving the cut
changes nothing at resolve time — the fit would derive a number, race it, report an
improvement, and resolve identically either way. That is precisely the failure
class this feature exists to prevent, so it now raises. (A *pure* decider was
already refused by `derive_threshold_from_pairs`; this catches the both-emitting
case, which has scores and would otherwise sail through.)

**7. Unscored blocker-similarity rows made the fit inert.** `PairRow.predicted_match`
deliberately refuses to threshold a row with `score_type is None` ("only a SCORED
row's `score` is a judge score") and returns `self.decision` — `None` — so
`ThresholdSelect.forward` drops **every** such row whatever value is fitted. A
Score-less chain therefore cannot have its cut fitted at all, and now raises. My
own earlier test asserted this case merely reported no held-out metrics; that was
the wrong bar — it was validating an inert fit that still stamped
`source="derived"` and an applied threshold.

**8. Chains that gate twice** — and the guard for it initially repeated the same
"bill, then refuse" mistake as #1. A `ClustererStage`'s clusterer keeps its own
threshold and gates the projected judgements independently of the
`ThresholdSelect`. Lowering the select to a winning 0.80 while the clusterer
still cuts at 0.90 reports a held-out improvement `resolve()` never realizes.
Refused rather than half-fitted — and the refusal is decided **before** the
scoring pass, because it is decidable from the topology alone. That is now the
stated rule in the code, after this PR hit the same shape three times: *every
refusal decidable from topology alone is decided before a paid `Score` runs*.
The one guard that genuinely cannot be hoisted (unscored rows) is annotated as
such. (`ClusterStage` is the abstract contract and carries no clusterer, so a
*custom* stage correctly reads as "no second gate" — tested.)

**9. The classic seam was left on the old convention.** `_fit_from_pairs` still
gated its held-out metrics on `aligned.valid.candidates` — the same pattern
fixed above in `_fit_chain_threshold`. **No shipped matcher can trigger it**
(every one yields exactly one judgement per candidate; an abstention is a
judgement with `score=None`, not a skipped yield), so this is hardening rather
than a live bug — but `Matcher` is a documented bring-your-own extension point
with nothing enforcing that cardinality, and leaving two sibling paths 280 lines
apart on different conventions, one just corrected, is how the next reader
concludes the divergence was deliberate. All three sites (`_described`, classic,
chain) now read the same way, and the test uses a custom matcher that skips the
held-out candidates.

## Known adjacent issue, not fixed here

On the **non-deriving** path, `_fit_from_pairs` grades held-out judgements on the
raw `_scorer()` scale while a fitted `Calibrator` may have rescaled the cut. That
is pre-existing, its docstring records the byte-identical behaviour as
deliberate, and changing it would alter results for every existing
`fit(pairs=...)` caller. Recorded here rather than silently carried. (The
*deriving* path does handle it — `_cut_scale_scores` / `_on_cut_scale` put both
sides on the cut's scale, applied only when `derive_threshold=True` so the old
path stays byte-identical.)

## Verification

- `uv run pytest -m "not integration and not slow and not finetune"` — **4543 passed**, 2 skipped
- `uv run pytest -m "not integration and not finetune"` (full, incl. slow) — **4642 passed**, 2 skipped, 16m31s
- `uv run ruff format --check . && uv run ruff check . && uv run mypy src` — clean
- `docs-clean-install` (the `tools/doc_snippets.py` gate) — **pass** in CI on this PR

Every new assertion in this PR that guards a specific defect was confirmed to
**fail against the unfixed code** before being kept — the counting matcher (435
calls), the Score-less chain (`PairMetrics(f1=0.0, fn=4)` where `None` was
required), and the classic-seam skipping matcher. A test that has never been
seen to fail is a hypothesis, not a safety net.

### Coverage delta on the contract glob

Measured locally, **both sides with the identical command below**, on the fast
subset — because `test-full`, which carries the real gate, is
`if: github.event_name != 'pull_request'` and structurally cannot run on a PR.
Baseline is this PR's merge-base, `82a222f`. The absolute is therefore not CI's
`test-full` number; the **delta** is the comparable quantity.

Reproduce (run once on this branch, once on `82a222f`):

```bash
uv run pytest -m "not integration and not slow and not finetune" -q -p no:randomly --cov-fail-under=0
uv run coverage report --precision=2 --include="src/langres/core/*,src/langres/metrics/*,src/langres/curation/anchor_store.py,src/langres/curation/canonicalizer.py,src/langres/curation/harvest.py,src/langres/curation/review.py,src/langres/tracking/judgement_log.py,src/langres/tracking/runs.py,src/langres/tracking/trackers/*,src/langres/training/*,src/langres/report/*,src/langres/autoresearch/*,src/langres/optimize.py"
```

| | statements | missing | branches | partial | total |
|---|---|---|---|---|---|
| `main` @ 82a222f | 9023 | 206 | 2616 | 130 | 96.94% |
| this branch @ 3b8bd51 | 9194 | **206** | 2678 | **130** | **97.00%** |

**Δ +0.06pp.** Missing statements *and* partial branches are **identical** on both
sides (206 / 130): this adds 171 statements and 62 branches to the contract glob
with **zero** newly uncovered lines and **zero** newly partial branches.

Per touched file: `training/fit_report.py` 85.00% → 90.74%, `core/_model_state.py`
93.93% → 95.32%, `core/resolver.py` 96.64% → 97.09%, `core/_model_run.py` 89.85%
→ 89.95%, `curation/harvest.py` 100% → 100%. Every one up or flat.

The one line this change originally added to the uncovered set was a dead branch
in `_prethreshold_pairs` (a classic-model arm whose only caller runs when
`_ops is not None`). It is deleted rather than covered or `pragma`'d.

**Why a green PR is not evidence here.** `test-full` — the job carrying the 98%
contract gate — is `if: github.event_name != 'pull_request'`, so it did not run
on this PR and could not have caught a regression in that number. That is why
the gate was run locally on both sides and the delta quoted above, rather than
inferred from the green checks.

Separately: `main` is currently **below** the 98 contract floor for reasons
predating this branch, and that is owned elsewhere — this PR moves the number
up, it does not reach the floor.

### DIFF
--- a/docs/EXPERIMENTS.md
+++ b/docs/EXPERIMENTS.md
@@ -426,6 +426,55 @@ threshold = derive_threshold(scores, labels, method="youden")   # or "percentile
 band and report on **held-out test** so the threshold isn't tuned on the pairs it's
 measured on — see `examples/research/m4_calibration.py` for the honest held-out version.
 
+### …or let `fit()` do it: `fit(derive_threshold=True)`
+
+The snippet above is the primitive. If your labels are id-keyed, the model can
+derive and apply the cut itself, holding out for you:
+
+```python
+resolver.fit(records, pairs=labeled_pairs, split=0.3, seed=0, derive_threshold=True)
+
+print(resolver.fit_report_.to_markdown())
+# - Threshold: 0.8824 (derived from 0.5000 by youden on 18 labeled pairs,
+#                      held-out, applied to the clusterer)
+```
+
+It is **opt-in**: the default leaves the constructed threshold alone, which is the
+right no-data fallback. What it changes is that a user who *does* have labels no
+longer resolves at a constant nobody measured.
+
+Four things worth knowing:
+
+- **It needs `pairs=`, not `labels=`.** `pairs=` is what carries the
+  entity-disjoint `split=`. With a split, the cut is derived on `train` and the
+  report's P/R/F1 grade it on `valid`, which the cut never saw. Without one the
+  cut is still derived — and the report says `IN-SAMPLE`, and computes no
+  metrics at all, because an in-sample number here would only flatter the cut.
+- **It fits matchers that have no fit hook.** `WeightedAverageMatcher`,
+  `RapidfuzzMatcher` and the LLM judges have nothing to train, so
+  `fit(pairs=...)` refuses them — but their *threshold* is a real fittable
+  parameter, and those are exactly the pipelines with nothing else to fit.
+- **It knows where your model keeps its cut.** A classic four-slot model writes
+  `clusterer.threshold`; a research recipe / `from_topology` chain has no
+  clusterer at all and writes its terminal `ThresholdSelect`.
+  `fit_report_.threshold_fit.applied_to` says which.
+- **Youden's J is symmetric in cost; ER usually is not.** It maximizes
+  `tpr - fpr`, weighting a false merge exactly as badly as a false split. In
+  entity resolution a false merge propagates through transitive closure and can
+  poison a whole cluster, while a false split leaves two records that can still
+  be merged later. If that asymmetry matters to you, treat the derived cut as a
+  starting point and move it **up**, or call `derive_threshold` yourself with
+  `method="percentile"`.
+
+Needs the `[trained]` extra (scikit-learn). On a core-only install it raises a
+directed `ImportError` rather than quietly keeping the default — identical code
+must not produce a different threshold depending on which extras are installed.
+
+The derived threshold survives `save`/`load`; its **provenance does not**
+(`fit_report_` is a fit-time artifact and is never serialized). Keep
+`fit_report_.model_dump_json()` beside the artifact if you need to show where the
+number came from.
+
 ## Budget monitoring (`SpendMonitor`, ≤ $5)
 
 ```python
--- a/docs/EXPERIMENTS.md
+++ b/docs/EXPERIMENTS.md
@@ -436,14 +436,43 @@ resolver.fit(records, pairs=labeled_pairs, split=0.3, seed=0, derive_threshold=T
 
 print(resolver.fit_report_.to_markdown())
 # - Threshold: 0.8824 (derived from 0.5000 by youden on 18 labeled pairs,
-#                      held-out, applied to the clusterer)
+#                      chosen on train, held-out, applied to the clusterer)
+#
+# ## Threshold selection (chosen on train)
+# - incumbent 0.5000: selection F1 0.5217, held-out F1 0.5714
+# - derived 0.8824 — KEPT: selection F1 1.0000, held-out F1 1.0000
 ```
 
 It is **opt-in**: the default leaves the constructed threshold alone, which is the
 right no-data fallback. What it changes is that a user who *does* have labels no
 longer resolves at a constant nobody measured.
 
-Four things worth knowing:
+**It derives, then races — it does not just apply.** A derived cut is not
+automatically a better cut. Youden's J maximizes `tpr - fpr`, which is flat
+across a wide separating gap, so on cleanly-separated data it happily returns a
+cut that ties on `train` and is *worse* on unseen pairs. We measured exactly
+that: held-out pair-F1 **1.0000 → 0.8000** on a fixture where `0.5` was already
+optimal. So the candidate has to beat the incumbent's pair-F1 to be applied; a
+tie keeps the incumbent, and `fit_report_.threshold_fit.source` reads
+`"declined"` with the rejected candidate still reported:
+
+```
+# - Threshold: 0.5000 (kept: the cut derived from 16 labeled pairs 0.8824
+#                      did not beat it on train)
+```
+
+**The race runs on `train`, never on `valid`.** Picking the winner on the
+held-out split would make that split a selection set and silently downgrade the
+reported P/R/F1 from a held-out estimate to an optimistic one. `train` was
+already spent deriving the candidate, so selecting there costs no further
+honesty — and both `held_out_f1` values in the report stay clean. The residual
+gap, stated rather than hidden: a cut chosen on `train` can still fail to
+generalise; selecting there bounds the risk, it does not remove it. A three-way
+derive/select/report split would close it, and is deliberately left as follow-on
+work — at the label counts this targets (a handful of corrections out of a review
+loop) three splits would make every number noise.
+
+Five more things worth knowing:
 
 - **It needs `pairs=`, not `labels=`.** `pairs=` is what carries the
   entity-disjoint `split=`. With a split, the cut is derived on `train` and the
@@ -464,7 +493,13 @@ Four things worth knowing:
   poison a whole cluster, while a false split leaves two records that can still
   be merged later. If that asymmetry matters to you, treat the derived cut as a
   starting point and move it **up**, or call `derive_threshold` yourself with
-  `method="percentile"`.
+  `method="percentile"`. (The race uses pair-F1, which inherits the same
+  symmetry.)
+- **On an explicit `_ops` chain, deriving costs a full pass over `data`.** A
+  chain's Source owns retrieval, so — unlike a classic model, which scores only
+  the labeled candidates — it must run the whole chain over every record. Cheap
+  for a local scorer; with a paid `Score` in a large chain, size `budget_usd`
+  for it.
 
 Needs the `[trained]` extra (scikit-learn). On a core-only install it raises a
 directed `ImportError` rather than quietly keeping the default — identical code
--- a/docs/EXPERIMENTS.md
+++ b/docs/EXPERIMENTS.md
@@ -480,9 +480,15 @@ Five more things worth knowing:
   cut is still derived — and the report says `IN-SAMPLE`, and computes no
   metrics at all, because an in-sample number here would only flatter the cut.
 - **It fits matchers that have no fit hook.** `WeightedAverageMatcher`,
-  `RapidfuzzMatcher` and the LLM judges have nothing to train, so
+  `RapidfuzzMatcher` and the **score-only** LLM judges have nothing to train, so
   `fit(pairs=...)` refuses them — but their *threshold* is a real fittable
   parameter, and those are exactly the pipelines with nothing else to fit.
+  A *decision-based* judge is the exception and is refused: an
+  `LLMMatcher(response_parser="binary_yes_no")` emits an explicit `decision`, and
+  `predicted_match` gives that precedence over the score, so no cut you fit could
+  change what `resolve()` does. Note the refusal happens after the labeled pairs
+  are scored on the classic path, so on a paid judge it costs those calls before
+  it fails.
 - **It knows where your model keeps its cut.** A classic four-slot model writes
   `clusterer.threshold`; a research recipe / `from_topology` chain has no
   clusterer at all and writes its terminal `ThresholdSelect`.
@@ -497,9 +503,13 @@ Five more things worth knowing:
   symmetry.)
 - **On an explicit `_ops` chain, deriving costs a full pass over `data`.** A
   chain's Source owns retrieval, so — unlike a classic model, which scores only
-  the labeled candidates — it must run the whole chain over every record. Cheap
-  for a local scorer; with a paid `Score` in a large chain, size `budget_usd`
-  for it.
+  the labeled candidates — it must run over every record. It runs the *prefix*
+  ending at the fitted `ThresholdSelect`, not the whole chain: anything
+  downstream of the cut (a reranker `Score`, say) is skipped, both because it
+  cannot influence a threshold applied before it and because its scores would
+  overwrite the ones the cut actually reads. So budget for every `Score` *above*
+  the fitted cut, and none below it. Cheap for a local scorer; with a paid
+  `Score` in that prefix, size `budget_usd` for it.
 
 Needs the `[trained]` extra (scikit-learn). On a core-only install it raises a
 directed `ImportError` rather than quietly keeping the default — identical code
--- a/docs/TECHNICAL_OVERVIEW.md
+++ b/docs/TECHNICAL_OVERVIEW.md
@@ -927,9 +927,9 @@ signature:
   `langres.training.calibration` implements it, consumed by
   `Resolver.fit(method=Platt()/Isotonic())` — see the `method=` seam below.
 
-`Resolver.fit(data, labels=None, *, pairs=None, split=None, seed=0)` consumes the
-**matcher** mixins, detected with `isinstance(module, SupervisedFitMixin)` /
-`isinstance(module, UnsupervisedFitMixin)`:
+`Resolver.fit(data, labels=None, *, pairs=None, split=None, seed=0, method=None,
+derive_threshold=False)` consumes the **matcher** mixins, detected with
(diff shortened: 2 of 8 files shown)


fxd24:
@codex review

fxd24:
Pushed three commits since the initial review request — two P1 fixes from cross-review plus a dead-branch removal. @codex review

fxd24:
All four Codex P2s verified against source and fixed in 952d010 (Score-after-Select contamination, missing `normalize_records`, the false `source="default"` provenance claim, and the silently-inert decider fit). Details and the reproducible coverage command are in the PR body. @codex review

fxd24:
Also applied the classic-seam convention fix (`_fit_from_pairs` was still gating held-out metrics on `aligned.valid.candidates`) in 001ca75 — no shipped matcher can trigger it, but all three sites now read the same way. Coverage delta and the reproducible command are in the body; note `test-full` cannot run on a PR, so the 98% contract gate was measured locally on both sides. @codex review

fxd24:
Both P1s from the 001ca75 review are fixed in 5656f01 + e6b4e13, and the two remaining coverage gaps I had introduced are closed in f62edb3 + 2ec52c4.

The Score-less-chain one was a good catch against my own test: it asserted that such a fit merely reported no held-out metrics, when the fit was in fact **inert** — `PairRow.predicted_match` returns `self.decision` (None) for `score_type=None`, so `ThresholdSelect` drops every blocker-similarity row whatever cut is fitted, while the report still said `source="derived"` with an applied threshold. It now raises, and the test asserts the refusal.

Coverage on the contract glob is back to parity on both counts: 206 missing statements and 130 partial branches, **identical to `main`**, with 171 statements and 62 branches added. Δ +0.06pp (96.94% → 97.00%); command is in the body.

@codex review

fxd24:
Fixed the new P1 in 3b8bd51: my own double-gate guard sat **after** `_prethreshold_pairs`, repeating the exact "bill the dataset, then refuse" bug I fixed earlier for the missing-`ThresholdSelect` case. Hoisted, and its test now asserts `calls == 0` (confirmed to fail against the un-hoisted version).

Three instances of that shape in one PR, so the rule is now written down in the code: **every refusal decidable from topology alone is decided before a paid `Score` runs.** The one guard that genuinely needs the rows (unscored blocker similarities) is annotated as the exception.

The other items in the latest review are re-reports of findings already fixed in earlier commits (decision precedence → `_refuse_deciders`; `normalize_records` → in `_prethreshold_pairs`; unscored rows and the clusterer gate → both now raise) — the bot re-reviews the cumulative diff each round.

Two I have **not** taken, with reasons:
- *Reject empty `pairs=[]` before scoring* — real, but `pairs` also accepts a path, so a complete guard means loading labels before blocking; that is a restructure of the fit entry, not a hoist. Worth a follow-up, not this PR.
- *Bind lazy architectures before pair-based fitting* — out of my write-lock (`architectures/`), and orthogonal to the threshold seam.

CI green, coverage at parity with main on both counts. @codex review

fxd24:
Latest round: the two P1s are **stale re-reports** — both are fixed at HEAD (the unscored-blocker-similarity one even links `001ca75` in its citation). Verified rather than assumed: `fit(derive_threshold=True)` on a Score-less chain raises "no judge score" (`test_a_score_less_chain_refuses_to_derive`), and a chain with a non-zero `ClustererStage` cut raises "gates twice" (`test_a_chain_that_gates_twice_refuses_to_derive`, asserting `calls == 0`). The bot re-reviews the cumulative diff each round, so findings resurface after they are fixed.

The three **doc** findings were real and are fixed in 8426c1f. All three were drift my own later commits created:
- `ThresholdFit.source` documented as `default` after I renamed it `not_fitted` — code branching on the documented name would never match, and Pydantic would reject `source="default"`.
- "it fits ... the LLM judges" became false once `_refuse_deciders` landed; now qualified to score-only judges, with the classic path's score-then-refuse cost stated.
- "must run the whole chain over every record" became false once `_prethreshold_pairs` started truncating at the fitted cut; it told users to budget for calls that cannot occur.

Good catch by the reviewer on all three — that is exactly the failure mode `CLAUDE.md` asks to be fixed in the same change, and I had missed it.

@codex review

PR #242, opened 2026-07-27
fix(embeddings): correct the ladder's half-driven claim, make the asymmetric prompt recipe safe, and measure the OSI-licensed field

fxd24:
## What this is

Three things, in the order they matter:

1. **A correction to a claim that is currently on `main`** and in the merged #239 body — the headline embedder number was *not* measured half-driven.
2. **The defect that claim was hiding**, fixed: the asymmetric prompt recipe works, but nothing checked the two halves agree and one index silently discarded a prompt.
3. **The ladder's OSI-licensed field, measured** — the five models stranded at `metric_revision=0`, which under David's "Gemma is a documented opt-in" call are the deciding measurement.

**It does not change `DEFAULT_EMBEDDING_MODEL`.** It publishes a recommendation and states the licences; the default is David's call.

---

## 1. The correction: `+0.2079` came from the `none` arm, not a half-driven one

#239's body says, verbatim:

> Every number below, including the **+0.2079** on `wdc_computers`, was measured with only the query side driven.

**That is false**, and it is verifiable two ways without re-running anything:

- the table it is published under is headed `## Is it better than what ships today? (k=20, **no instruction**)` (`20260727_embedder_ladder.md:135`; the row is `:149`);
- `render_report` selects that table on `prompt_arm == "none"` — *neither* side prompted.

`none` is **langres's default configuration**: a `VectorBlocker` with no `query_prompt` over an embedder with no `prompt_name`. So the correction runs in the direction that makes the number **stronger** — it was measured in the configuration langres actually ships, not a crippled one.

The genuinely half-driven arm is `instruct`, and it is a different and much smaller number: on `wdc_computers` it moves EmbeddingGemma by **+0.0084**, CI `[-0.0036, +0.0209]` — spanning zero.

The correction is rendered **into the report by the generator**, not hand-edited into the `.md` — the file is regenerated from the JSONL after every model, so a hand edit would have been erased by the next run. The `instruct` number it quotes is looked up from the rows for the same reason: a correction that hand-types the number it corrects is the same defect.

A correction comment is also posted on #239 itself.

## 2. The second false claim, and the real defect underneath it

The report also said the asymmetric recipe was **"not expressible through the blocking API"**, because `create_index(texts)` takes no prompt. Also false — the document prefix does not travel through that argument. It rides on the **embedder**: `SentenceTransformerEmbedder` forwards `prompts=`/`prompt_name=` as `default_prompt_name`, and sentence-transformers resolves `encode(prompt=None)` back to it (`base/model.py::_resolve_prompt`). `search_all(query_prompt=…)` then passes an *explicit* prompt, which takes precedence.

Verified empirically before anything was written, then frozen as a test: the recipe through the shipped API is **byte-identical** to the harness's prefix-the-corpus-by-hand workaround.

**`create_index` is unchanged.** `VectorIndex` is a public *structural* protocol with at least six known test doubles in this repo alone; a required new kwarg breaks every external implementer for a recipe that already works.

What was genuinely broken, and is fixed here:

| | Defect | Consequence |
|---|---|---|
| P1 | `QdrantHybridIndex.search_all` accepted `query_prompt` and **discarded** it | a sweep over the axis returned *identical* numbers at every prompt. Identical numbers read as "the prompt does not help" — this repo published exactly that table once, from the same shape of no-op in `FAISSIndex.search_all`. It now raises `NotImplementedError`, **before** issuing any query |
| P2 | nothing checked the two halves of the recipe agree | `prompt_name` set + `query_prompt` unset is *worse* than prompting neither side: `search_all(query_prompt=None)` reuses the document-prompted corpus vectors as queries, so queries carry the **document** prefix. `VectorBlocker` now warns on exactly that direction |
| P3 | the recipe was undocumented | now a worked example in `VectorBlocker`'s class docstring and on `SentenceTransformerEmbedder.prompt_name` — the two places a user meets the two halves |

The coherence check is duck-typed and degrades to "cannot tell" rather than raising, precisely because `VectorIndex` is structural. It stays **silent** on query-only (a legitimate, measured arm) and on the bare default; it fires only in the direction that is actually wrong.

**Every prompt assertion is a result comparison, never a plumbing one** — including two controls that require the prompt to *change* the answer. A `query_prompt` that forwarded correctly and did nothing is what shipped last time.

## 3. The ladder: the OSI-licensed field, measured

Five models were stranded at `metric_revision=0` — measured under a retired metric definition, so excluded from every table. All five are re-measured here at revision 1. They are also, exactly, the **OSI-licensed** field, which is what makes this the deciding measurement rather than a completeness chore: langres is Apache-2.0 and EmbeddingGemma is not.

`ModelSpec` now carries the licence **each model card declares** — read from the cards on 2026-07-27, not remembered — and the report splits its recommendation in two: the OSI-licensed field (the only candidates for a default) and the use-restricted checkpoints (documented opt-in, never a silent default). `OSI_APPROVED_LICENSES` is an **allow** list, so a licence nobody has classified reads as "not clearly OSI" instead of passing by absence from a deny list.

Winners, counts and the coverage denominator are all derived from the rows.

## What did NOT run, as named rows

**Seven of the fourteen models in the ladder have a row at the current metric revision.** The report says so in its own recommendation section, derived from the rows rather than typed. Measured at revision 1: `all-MiniLM-L6-v2`, `all-MiniLM-L12-v2`, `BAAI/bge-small-en-v1.5`, `BAAI/bge-base-en-v1.5`, `all-mpnet-base-v2`, `intfloat/e5-base-v2` — 40 rows each — and `google/embeddinggemma-300m` at 60 (it carries the third `documented` arm; see §4).

Seven models were never run at all and are named in the report's own derived "What did not run" table: `Alibaba-NLP/gte-base-en-v1.5`, `nomic-ai/nomic-embed-text-v1.5`, `BAAI/bge-large-en-v1.5`, `mixedbread-ai/mxbai-embed-large-v1`, `Qwen/Qwen3-Embedding-0.6B`, `Qwen/Qwen3-Embedding-4B`, `Qwen/Qwen3-Embedding-8B`. **The ladder is not finished** and the report says so in its own recommendation section, with the denominator spelled out.

## 4. The `documented` arm, measured for the first time — and it settles the thesis

The ladder shipped a third `documented` arm (the checkpoint's **own** document *and* query prefixes, read from its `config_sentence_transformers.json`) that **had never produced a single row**. That is this repo's recurring shape: a code path whose output nobody had ever seen. Since the whole PR is about the asymmetric recipe, an unmeasured "does the documented recipe work?" arm was the one gap worth closing. It is $0 and offline, so I ran it.

It fires (verified by measurement, not inference: a 2-arm model builds **1** index per benchmark, gemma builds **2** — the extra one is the document-prefixed corpus), and the result is decisive. All at k=20, deltas vs. the same model's own bare `none` arm, cluster-resampled CIs:

| benchmark | `none` | `instruct` (query-only) Δ | `documented` (both sides) Δ |
|---|---:|---|---|
| abt_buy | 0.9741 | **−0.0049** `[−0.0108, +0.0010]` spans 0 | **+0.0088** `[+0.0039, +0.0147]` clear |
| amazon_google | 0.8014 | +0.0123 `[+0.0052, +0.0206]` clear | **+0.0371** `[+0.0263, +0.0485]` clear |
| fodors_zagat | 1.0000 | 0.0000 (saturated) | 0.0000 (saturated) |
| walmart_amazon | 0.8755 | 0.0000 `[−0.0044, +0.0044]` spans 0 | **+0.0077** `[+0.0022, +0.0143]` clear |
| wdc_computers | 0.7786 | +0.0084 `[−0.0039, +0.0206]` spans 0 | **+0.0394** `[+0.0247, +0.0545]` clear |

**Driving both sides wins on 4 of 5 with the interval clear of zero. Driving only the query side wins on 1 of 5, and on `abt_buy` it is negative.** On `wdc_computers` the full recipe is 4.7× the half-driven one and clear of zero where the half-driven one is not.

That is the empirical case for every other change in this PR: the recipe is real and worth ~+0.04 recall where it bites, half-driving it measures almost nothing, and `prompt_name`-without-`query_prompt` — the case the new `VectorBlocker` warning catches — is the *worse-than-neither* corner of the same axis.

Two honest limits: `fodors_zagat` is saturated at 1.0000, so its `0.0000` is vacuous, not evidence. And `documented` rows carry **no** vs-incumbent delta, because that comparison needs a matching reference cell and the reference model (`all-MiniLM-L6-v2`) documents no recipe — there is nothing to pair against. The within-model comparison above is the one that is defined.

## 5. `QdrantHybridIndex` claimed a protocol conformance it does not have

Found in review, same defect family as the `query_prompt` discard. Its docstring said *"Implements VectorIndex protocol"*, while `search` accepts `str | list[str]` against the protocol's `str | list[str] | np.ndarray`. A narrower parameter type is not substitutable, so a type checker rejects the assignment — the claim was simply false, and nothing could contradict it.

**I corrected the claim rather than widening the signature.** A hybrid index cannot serve a pure-vector query: the sparse side has to encode the *text*. Widening `search` to accept `np.ndarray` and then raising would make mypy pass over a claim that is still false at runtime — the same defect one layer down. The docstring now states exactly which parts conform and which do not, and why.

This is not a separate errand: the protocol's `np.ndarray` branch **is** the pre-computed-embeddings path, and `QdrantHybridIndex` re-spells that same capability as the private `_dense_embeddings`. Two names for one capability is why neither a reader nor a type checker noticed the `query_prompt` discard living between them.

**The check is runtime signature introspection, not a mypy assertion, and that is deliberate**: CI runs `uv run mypy src` (`lint.yml:52`) — *src only*, so a mypy assertion in `tests/` would be checked by nothing. `tests/core/indexes/test_vector_index_conformance.py` compares each implementation against the protocol at runtime.

**I verified the gate can fail before trusting it.** Flipping the expectation back to the old false claim makes it fail, naming `numpy.ndarray` as precisely what is missing. It also carries a premise guard: if `VectorIndex.search` ever loses its `ndarray` branch, every implementation would conform and every assertion would pass vacuously — so that is asserted too.

## Also fixed while running it

Three things the sweep itself surfaced, each committed with the reason:

1. **`run_ladder.sh` could not run from a bare checkout.** It hard-required `uv run --env-file .env`; `.env` is gitignored, so a fresh checkout or worktree exited 2 before the harness started — on a **$0, offline** sweep that needs no key. The flag is now conditional, and `LADDER_MODELS` drives a subset without editing the model list in place.

2. **Making `.env` optional exposed a dependency nobody could see, and it bit immediately.** The OpenMP settings `docs/FRICTION_LOG.md` documents live *only* in that gitignored file. Without them torch/faiss/sklearn load two `libomp.dylib` runtimes and the sweep **deadlocks** in `__kmp_join_barrier` — 0% CPU forever, no exception, so the script's own failure path never fires and the sweep just silently stops. It hit `all-mpnet-base-v2` after three models had already succeeded (sampled: two libomp base addresses, main thread parked in `__kmp_internal_join`; 6s of CPU in 16 minutes). `run_ladder.sh` now defaults `OMP_NUM_THREADS=1` / `KMP_DUPLICATE_LIB_OK=1` when unset, so `.env` and an explicit caller still win. A script whose documented prerequisite is invisible is a script that hangs.

3. **`examples/research/blocking_evaluation_with_instructions.py` raised `TypeError` on its first setup call** — pre-existing, found by review during this change. It passed `query_prompt=` into `FAISSIndex(...)` and `QdrantHybridIndex(...)`; no `VectorIndex` constructor has ever accepted it. Fixed here rather than merely reported because it is the same API this PR documents: the kwarg moves to `VectorBlocker` on the FAISS arm and to the `search()` call on the CrossEncoder arm. The Qdrant-hybrid arm now carries **no** instruction and says so — it reaches the index via `search_all`, which cannot apply one. That arm was previously publishing an un-instructed row inside a "with instructions" comparison.

## 6. The cache-integrity work, and why it took six rounds

Reviewing the ladder's provenance surfaced a hazard the report had answered with a sentence: the embedding cache namespace is model name + dtype with **no Hub revision**, so a checkpoint re-uploaded under the same name still hits it — a warm re-run reads the *new* checkpoint's `parameter_count`/`embedding_dim` while serving the *old* checkpoint's vectors and publishes a row mixing the two. The documented remedy was "delete the namespace first", which is a remedy nothing can observe.

It is now measured. Each run re-encodes a fixed canary string and compares it to the cached entry; a mismatch aborts. That is deliberately **stronger** than the revision pin the reviewer first asked for, not cheaper: a revision answers *is this the same Hub commit*, the canary answers *does this checkpoint still produce these vectors* — the property the measurements actually depend on — so it also catches a pooling change, a tokenizer fix, or a different dtype path, and it invalidates no cached vector to do it.

Worth recording plainly, because it is the more useful finding: **the first five versions of that check were each unable to observe the failure they existed to catch**, and each hole only appeared when the next one was tested against real data rather than read.

| round | the hole | how it surfaced |
|---|---|---|
| 1 | On a pre-existing cache the canary *misses* — encoded fresh, written, compared against itself. Always passes. And **every** namespace in the ladder was pre-existing, so it was vacuous on the only run that mattered. | review |
| 2 | The refusal asked "is the canary here?" by *encoding* it, so a refused run deposited one and the next run sailed through. A refusal you get past by running it twice. | trying to prove it against the real cache |
| 3 | The refusal was retroactive with no way out: 1.8 GB and six namespaces would have to be re-measured. `--trust-existing-cache` vouches once, loudly, and the next run is checked normally. | measuring the real cache state |
| 4 | That flag defaulted to the whole ladder, so one command blessed all six namespaces — from a flag documented as vouching for one. | review |
| 5 | The refusal *deleted the measurements it protects*: it landed in a blanket `except Exception`, became a `status="failed"` row, and `merge_rows()` voided every prior cell for that model. | review |
| 6 | Fixing (5) in Python only **moved** the deletion — `run_ladder.sh` reads any non-zero exit as "process died" and deletes the rows itself. Now a reserved exit code (3) that the driver bails on before the row-deleting path. | review |

Plus: a `NaN` canary read as *agreement* (`NaN > tolerance` is `False`), and the canary vouched for one cache partition while the `instruct`/`documented` arms read prompt-keyed entries it never looked at.

Round 6 is verified end-to-end rather than argued: `--models intfloat/e5-base-v2 --cache-dir tmp/embedder_ladder_cache` exits **3**, names the 70,804-vector namespace, leaves the cache at exactly 70,804 entries, and leaves `docs/research/*_rows.jsonl` untouched.

Two side-effects worth knowing about:

- **`QdrantHybridRerankingIndex` refuses a `query_prompt` it cannot honour.** The `NotImplementedError` this PR adds to `QdrantHybridIndex.search_all` originally recommended that index as the alternative. It is not one: in *its* `search_all` the dense side is served from cached vectors and the sparse side always encodes unprompted, so only the reranking embedder could see the prompt — and the production one (`FastEmbedLateInteractionEmbedder`) ignores prompts outright, as its own source comment says. Following the refusal's advice therefore restored the exact silent no-op the refusal exists to prevent. Now a checked capability (`honours_prompt`), defaulting to `True` so unknown duck-typed embedders are trusted rather than broken.
- **The coherence warning no longer leads with a remedy that crashes.** `VectorBlocker` forwards `query_prompt` to `search_all()`, so "pass `query_prompt=...`" turned a running hybrid configuration into a `NotImplementedError`. It now leads with the remedy that works on every index.

## Verification

Run locally, one `uv run` at a time (two concurrent ones in a single worktree race on `.venv`/`.coverage` and fabricate plausible failures):

| gate | result |
|---|---|
| full suite + coverage | **4635 passed**, 2 skipped, 18 deselected, 1097s — `Required test coverage of 90% reached. Total coverage: 96.05%` |
| `tests/core` (the gated contract) | 1635 passed, 4 deselected, 165s |
| `ruff format --check .` | clean (3 files reformatted in `fb41f21`; layout only, no string content changed, so the generated report is byte-identical) |
| `ruff check .` | `All checks passed!` |
| `mypy src` | `Success: no issues found in 215 source files` |
| `tools/doc_snippets.py` | 8 passed, 0 failed, 28 skipped |

### A finding while measuring that: the on-disk `.coverage` under-reports

I could not honestly quote a full-suite path-scoped percentage, because the measurement is not stable, and the reason is worth recording rather than working around.

`src/langres/core/blockers/vector.py` measures **97.17%** from `tests/core` alone. Reading the **same** file out of the *full* suite's `.coverage` — a strict superset of those tests — gives **64.31%**, and an earlier full run gave **32.51%**. A coverage union cannot shrink when you add tests, so at least one of those numbers is not a measurement of anything. Two more symptoms line up with it: the glob's statement *denominator* differed between two full runs of the same include list (9037 vs 6702), and pytest-cov's own in-process summary (`Total coverage: 96.05%`) disagrees with what `coverage report` reads back from the file it just wrote.

The in-process summary is computed before the file is re-read, so **it is the trustworthy one**; a `coverage report` run afterwards is not. That matters beyond this PR, because the 98% path-scoped ratchet in `.github/workflows/test.yml` is exactly a post-hoc `coverage report` against that file. I have not chased the mechanism (a spawned subprocess writing over the data file is the leading candidate — `examples/flywheel_min.py` shells out to `.venv/bin/python`), because it is pre-existing, out of this track's scope, and does not block this PR: the gate lives on `test-full`, which cannot run on a PR at all. Flagging it because a ratchet that reads an unreliable file is another instance of the pattern this repo keeps hitting — a green light decoupled from the thing it claims to check.

**One invocation caveat, since it cost me a false failure:** run the suite through `uv run pytest`, not `.venv/bin/python -m pytest`. The latter does not put `.venv/bin` on `PATH`, and `tests/examples/test_flywheel_min.py` shells out to the bare `langres` console script — it fails with `FileNotFoundError: 'langres'` and reads exactly like a regression. It passes with `PATH` set.

### Coverage: read this as a delta, not an absolute

`main` is **already below the 98 path-scoped floor (97.28%)** and has been since #229 on 2026-07-21 — six days and roughly seven merges. This branch inherits that; it is not this branch's bug, and a separate agent owns the fix. **The bar here is "does not make it worse."**

Note the gate is on the `test-full` job, which is `if: github.event_name != 'pull_request'` — it **structurally cannot run on a PR**, so a green PR is not evidence either way. Hence the local run above.

Running CI's exact include list (copied verbatim from `.github/workflows/test.yml:288`, not reconstructed) against this branch's coverage data:

| | path-scoped total |
|---|---|
| `main`, per the failing CI run | **97.28%** |
| this branch, local | **97.28%** |

Those two figures come from different machines (CI ubuntu vs. local macOS), so read the match as corroboration rather than as one measurement.

**One trap worth naming, because I nearly reported a regression from it:** this gate must be measured with the **full** suite, not the fast one. Running it under `-m "not slow and not integration and not finetune"` gives **96.95%** — not a regression, just fewer tests executing less code. `test-full` runs the whole suite, so only a full-suite number is comparable to main's. A fast-suite figure quoted against a full-suite baseline manufactures a regression that does not exist.

**The actual proof that this branch cannot have lowered it is structural**, and it is cheap to check:

- this branch touches exactly three files under the gated tree — `core/blockers/vector.py`, `core/indexes/hybrid_vector_index.py`, `core/embeddings.py`;
- `embeddings.py`'s diff is **entirely docstring prose** — zero executable statements;
- `hybrid_vector_index.py` is at **100.00%**;
- `vector.py` is at 97.87%, and every one of its uncovered lines is pre-existing: `616`/`622` are `_index_is_built` branches, `576->579` is in `from_config`, `915->922` is in `inspect_candidates`. The lines this branch adds — `_document_prompt_name` and the coherence warning — are covered.
- `core/indexes/__init__.py` measures **11 statements / 84.62%**, byte-identical to the figure on the failing main run. The `TYPE_CHECKING` conformance block adds **zero** counted statements: `exclude_lines` in `pyproject.toml` lists `if TYPE_CHECKING:` (which excludes the whole block) and `raise NotImplementedError`. Verified by measurement, not by reading the config.

A change that adds only covered statements cannot lower a coverage percentage.

`src/langres/core/indexes/*` is this track's tree and appears in that failing report, so I checked its uncovered lines against the source myself rather than inheriting the summary:

| line | what it is | mine? |
|---|---|---|
| `vector_index.py:417`, `:589` | `raise ValueError(f"Unknown metric: …")` — unreachable; `metric` is a `Literal["L2","cosine"]` | no |
| `vector_index.py:584` | the `faiss.IndexFlatL2(dim)` branch — a genuinely untested *branch*, not a guard | no |
| `vector_index.py:662`, `:701` | `FakeVectorIndex` "Index not built" guards | no |
| `vector_index.py:729-730` | `FakeVectorIndex.build()` — an explicitly **deprecated** back-compat method | no |

All pre-existing; none introduced by the ladder work or by this branch. (Two of them are worth correcting against the earlier summary that called them *all* guards: `:584` is a branch and `:729-730` is deprecated dead API.) I did **not** add assertion-free tests to move the number.

## What cross-model review caught

Codex reviewed three times. It found **six** issues, and one of them landed on this PR's own headline evidence — worth recording, because that is the case *for* running it rather than treating it as a formality.

| | finding | outcome |
|---|---|---|
| P1 | the `documented` rows were published under prose saying every comparison keeps the same index and re-encodes only the query | **correct, and material.** That arm also rebuilds the corpus with the document prefix, so its Δ is document-plus-query vs. neither. Verified by measurement, not re-reading: gemma builds **2** indexes per benchmark where a query-only model builds **1**. All four sites now scope query-only to `instruct` |
| P1 | the example's hybrid arm was labelled unprompted only in the saved JSON; the console still printed `Hybrid` under a "With Instruction Prompts" banner | banner now says "on 2 of 3 arms" and names them; column is `Hybrid*` with a footnote; the arm's run name uses the label |
| P1 | one global `embedding_instruction` saved for three arms with two prompt regimes | replaced with per-arm `query_prompt_by_arm` + `confounded_deltas` |
| P2 | `FakeHybridVectorIndex.search_all` accepted a `query_prompt` the real index now rejects | fake raises too. A double that accepts what production rejects lets a test pass and then fail on the real index — hiding the exact defect the raise exists to surface. **Not** applied to the reranking fake: `QdrantHybridRerankingIndex.search_all` genuinely can apply a prompt, so mirroring there would be the inverse bug |
| P2 | the conformance table reduced non-conformance to a boolean | stores the exact accepted type set, so narrowing `str \| list[str]` to bare `str` now fails |
| P2 | the gate omitted `QdrantHybridRerankingIndex` / `FakeHybridRerankingVectorIndex` | both added, plus the claim at `reranking_vector_index.py:54` and a third site codex did not flag — the module-level claim at `hybrid_vector_index.py:10` |

I also caught myself once, while writing the request asking codex to check it: I had labelled `crossencoder_vs_faiss` `like_for_like_deltas`, when it matches only on the *prompt regime* and still swaps the index and adds a reranker. Renamed. Trading one misleading label for another is the same defect.

## Provenance

The report now generates a **How to reproduce these numbers** block — the command, `k` values, benchmarks, metric revision, and the OpenMP prerequisite — derived from the rows rather than typed. Every result was committed *and pushed* the moment it existed, one commit per model, which is what `run_ladder.sh` exists to guarantee.

Raw logs under `tmp/` are **deliberately discarded**: every number they produced is in `docs/research/20260727_embedder_ladder_rows.jsonl` (tracked), the 3-arm comparison is rendered into the report's own table, and `tmp/embedder_ladder_cache/` is 1.8G of regenerable derived embeddings.

## One thing this PR deliberately does not contain

`QdrantHybridIndex.search_all` now **raises** where it previously ignored a `query_prompt`. That is a user-visible behaviour change and belongs in `CHANGELOG.md` under `[Unreleased]`. `CHANGELOG.md` is owned by another concurrent stream in this batch, and two streams editing it produce a conflict for no benefit — so it is flagged here rather than edited. Suggested entry:

> **Changed** — `QdrantHybridIndex.search_all()` now raises `NotImplementedError` when passed a `query_prompt` instead of silently discarding it. It serves queries from vectors cached at `create_index()` time, so the prompt could never reach the encoder; a discarded prompt makes a prompt sweep return identical numbers at every setting. Use `FAISSIndex`, `QdrantHybridRerankingIndex`, or `search(..., query_prompt=...)`.


fxd24:
@codex review

Three things worth pointing at, because each is a place where a plausible-looking check would not catch the defect:

1. **The correction in `examples/research/embedder_ladder.py` is rendered by the generator, not written into the `.md`.** The report is regenerated from the JSONL after every model, so a hand edit is erased by the next run. Please check the correction actually survives a re-render, and that the `instruct` delta it quotes is *looked up from the rows* rather than hardcoded.

2. **`QdrantHybridIndex.search_all` now raises instead of discarding `query_prompt`.** The failure mode this guards is a prompt sweep returning identical numbers at every setting, which reads as "the prompt does not help". Please check the raise happens *before* any query is issued, and that no legitimate caller in the repo is now broken by it.

3. **The `VectorBlocker` coherence check is duck-typed on a public structural protocol.** It must degrade to "cannot tell" rather than raise for indexes that expose no embedder, stay silent on query-only and on the bare default, and fire only when `prompt_name` is set without `query_prompt`. Please check the decorator-unwrap does not swallow a real disagreement.

Coverage note: this branch reads 97.28% on CI's path-scoped include list, the same figure `main` has had since #229 — that floor was already red before this branch and a separate agent owns it. The structural argument that this change cannot have lowered it is in the PR body.

fxd24:
@codex review

Fixed your P1 in `c9559f6`. You were right that making the hybrid arm honest about carrying no `query_prompt` left the *reporting* dishonest — the script still announced itself as "with instructions" and saved one global `embedding_instruction`, so `hybrid_vs_faiss` mixed an index change with a prompt-regime change under a label that looked like every other delta.

What changed:

- `query_prompt_by_arm` replaces the single global key, so the regime is per-arm **data**, not a caveat in prose.
- `confounded_deltas` / `like_for_like_deltas` name explicitly which comparisons are not like-for-like (`hybrid_vs_faiss` and `crossencoder_vs_hybrid` are; only `crossencoder_vs_faiss` is clean).
- `HYBRID_ARM_LABEL = "Qdrant Hybrid (unprompted)"` puts the regime in the label itself, so it travels with the value into every ranking and saved file rather than living somewhere the reader has to go find.
- the two log lines no longer claim "with instructions" for that arm.

Please check I did not leave a fourth site still implying the arm is prompted, and that `like_for_like_deltas` is actually correct — it asserts `crossencoder_vs_faiss` compares two equally-prompted arms, which is the kind of claim that is easy to state and easy to get backwards.

Note a results commit for `google/embeddinggemma-300m` may land on this branch shortly (docs/rows only, no source change).

fxd24:
@codex review

Head is now `f855254`. Three things landed since your last pass:

**1. Your P1, fixed (`c9559f6`) — and I then caught myself overclaiming in the fix (`82e90ee`).** I had labelled `crossencoder_vs_faiss` as `like_for_like_deltas`. It is not like-for-like; it matches on the **prompt regime** only and still swaps the index backend and adds a reranker. Renamed to `prompt_regime_matched_deltas`. Trading one misleading label for another would have been the same defect you flagged.

**2. A protocol-conformance defect in the same family (`6657d4b`).** `QdrantHybridIndex`'s docstring claimed *"Implements VectorIndex protocol"* while its `search` accepts `str | list[str]` against the protocol's `str | list[str] | np.ndarray`. I **corrected the claim rather than widening the signature** — a hybrid index cannot serve a pure-vector query because the sparse side must encode the text, so widening-and-raising would make mypy pass over a claim still false at runtime.

Worth your attention: the new check is **runtime signature introspection, not a mypy assertion**, because CI runs `uv run mypy src` (`lint.yml:52`) — src only — so a mypy assertion in `tests/` would be checked by nothing. Please sanity-check that reasoning, and that `SEARCH_CONFORMANCE`'s `FakeHybridVectorIndex: False` entry is right rather than copied from its sibling.

**3. The `documented` arm produced rows for the first time (`a74aadb`).** It had **never** emitted a single row in the ladder's history. Following the checkpoint's own document+query recipe beats the bare default on 4 of 5 benchmarks with CIs clear of zero (wdc_computers +0.0394), while the query-only `instruct` arm manages 1 of 5 and is *negative* on abt_buy. That is the measurement behind this PR's whole thesis.

I verified the new conformance gate can actually fail (flipping the expectation to the old false claim fails, naming `numpy.ndarray`). Please check I have not left the gate able to pass vacuously in some other direction.

fxd24:
@codex review

Head is now `2d45da3`. All five findings from your last pass are addressed — thank you, the P1 on the report prose was the most valuable single comment on this PR, because it landed on the headline evidence.

**1. "Describe the documented arm as a two-sided prompt change" (P1) — you were right, and I verified it by measurement rather than by reading.** Four sites called the delta query-only ("same index", "instruct minus none"). The proof it is not: gemma builds **2** FAISS indexes per benchmark where a query-only model builds **1** — the extra one is the document-prefixed corpus. All four sites now scope query-only to `instruct` and state that `documented` is document-plus-query vs. neither.

**2. "Carry the unprompted label into console output" (P1).** The label reached only the saved JSON. Now: the banner says "on 2 of 3 arms" and names which, the comparison column is `Hybrid*` with a footnote directly under it, and the arm's own run name uses `HYBRID_ARM_LABEL`.

**3. "Mirror prompt rejection in the hybrid fake" (P2).** `FakeHybridVectorIndex.search_all` now raises too. Note I did **not** do the same for `FakeHybridRerankingVectorIndex`: `QdrantHybridRerankingIndex.search_all` genuinely *can* apply a query prompt (its reranking embedder encodes queries), so making its fake raise would be the mirror-image bug. Please check that distinction.

**4. "Assert the exact nonconforming signatures" (P2).** Replaced the boolean with an explicit expected type set per class (`TEXT_ONLY` / `TEXT_OR_VECTORS`), so narrowing `str | list[str]` to bare `str` now fails.

**5. "Cover the hybrid reranking indexes" (P2).** Both added, and the false claim at `reranking_vector_index.py:54` corrected — along with a third site you did not flag, the module-level claim at `hybrid_vector_index.py:10`.

Also added, because CI runs `mypy src` and the only real `VectorBlocker(vector_index=QdrantHybridIndex(...))` construction lives under `examples/`: a `TYPE_CHECKING`-only conformance function in `core/indexes/__init__.py`. The non-conforming cases carry `# type: ignore[assignment]` on purpose — `strict = true` implies `warn_unused_ignores`, so widening one to conform makes the ignore unused and **fails mypy**. I verified both directions: removing an ignore gives `[assignment]`, adding a spurious one gives `[unused-ignore]`.

fxd24:
@codex review

Head is now `8680ee2`. Both new findings from your last pass are fixed, and both were against text I had just added — which is the argument for reviewing a docs change, not only code.

**1. "Qualify the offline reproduction claim" (P2) — you are right and my wording was simply false.** The block said "$0 and offline — no API key, no network at measure time". `SentenceTransformerEmbedder` leaves `local_files_only` false, so a cold cache downloads from the Hub. It now says $0 and key-free, states that a cold run needs network and records failure rows without it, and scopes "offline" to a warm re-render.

**2. "Include the checkpoint revision in the cache identity" (P1).** I had written the weaker claim that a re-run after a re-upload "can legitimately differ". Your framing is sharper and correct: the namespace is `name + dtype` only (`_build_embedder`), so a warm cache serves the **old** vectors while `parameter_count`/`embedding_dim` come from the **new** checkpoint — one row mixing two checkpoints, with nothing in it to reveal that. Documented in the report *and* at the namespace that causes it, with the remedy (delete the namespace before re-measuring).

I did **not** add the revision to the namespace, and I want that decision challenged rather than assumed: doing so invalidates every cached vector and forces a full re-measure of the ladder for a hazard that has not occurred. If you think an unsafe-by-default cache key is worth that cost now rather than at first bite, say so.

Note four of the threads on the previous SHA were carried forward by GitHub rather than re-raised — the per-arm provenance, console labelling, and hybrid-fake rejection are all fixed in `c9559f6`/`2d45da3`. Worth confirming you see them as closed.

CI was green on `2d45da3` (8/8 real checks; `test-full` is `if: github.event_name != 'pull_request'` so it cannot run here).

fxd24:
@codex review

Round 4 findings, all addressed at `c028c0c`:

**P1 — Key the embedding cache by checkpoint revision.** You were right to push back on the deferral, and the reason you gave is the load-bearing one: a documented cleanup with no automatic signal is a remedy nothing can observe. Fixed, but with a *stronger* check than a revision pin rather than a cheaper one — `_assert_cache_matches_checkpoint` re-encodes one fixed canary string per model per run and compares it to the cached entry; a mismatch aborts and names the `.db` to delete.

Why not the revision: it answers *is this the same Hub commit*, whereas the property the measurements depend on is *does this checkpoint still produce these vectors*. The canary answers that one directly, so it also catches drift a revision cannot see — a sentence-transformers pooling change, a tokenizer fix, a different dtype path on a new device. It also invalidates nothing, where putting the revision in the namespace would have discarded every cached vector in the ladder.

Three tests, and I watched them fail: forcing `_CANARY_TOLERANCE` negative (always raise) fails all three, so a degenerate always-raising check cannot masquerade as this one. The two controls (cold cache, warm cache from the *same* checkpoint) exist for exactly that.

**P1 — Carry the unprompted label into console output.** Correct, and the comparison-table footnote did not cover it: the KEY INSIGHTS block prints separately and a footnote 400 lines up does not travel with it. Every row in that block now carries its prompt regime, and each delta says whether it is prompt-matched or confounded — the same split already saved to JSON. Two claims went with it: the hybrid-wins recommendation credited "keyword matching" for a difference in which index *and* prompt moved together, and `Instruction prompts: Expected +1-5% quality improvement` cited a number nothing in that run measures (no two arms share an index). It now points at the ladder, which holds the index fixed.

**P2 — Do not classify every default prompt as document-side.** Correct and it was a real false positive on a documented recipe. `intfloat/e5-base-v2`'s model card asks for `"query: "` on *both* sides for symmetric tasks (verified against the card), and e5-base-v2 is one of the models in this ladder — so the warning fired on a correct configuration and called its prompt "document-side". Exempted via a one-element allow-list (`"query"` is what `SentenceTransformer.encode_query` binds), with a set-equality test so the exemption cannot widen into a mute button.

**P2 — Assert the exact nonconforming signatures / cover the reranking indexes.** Both done earlier: `SEARCH_QUERY_TYPES` now pins the exact accepted type set per class (not a superset boolean) and includes `QdrantHybridRerankingIndex` / `FakeHybridRerankingVectorIndex`.

**P2 — Count arbitrary requested models in the denominator.** Fixed at both sites; derived from `MODELS | measured`. The regenerated report is byte-identical, since every committed row names a model already in the tuple.

**P2 — Mirror prompt rejection in the hybrid fake / qualify the offline claim.** Both landed in earlier rounds.

fxd24:
@codex review

Round 5, all three addressed at `2fab4f5`.

**P1 — Reject legacy caches before inserting the canary.** You are right, and this one was a real hole in a check I had just written to close exactly this class of defect. On a pre-existing namespace the canary *misses*: encoded fresh from the checkpoint loaded now, written into the unvouched database, then compared against another fresh encoding of itself. It always matches, while every corpus vector beside it may belong to a different checkpoint. A gate decoupled from what it checks, reintroduced one level up — and it would have failed silently on the very first run, which is the run that matters, since every namespace in the ladder today predates it.

Fixed by reading the entry count *before* the canary is written (`_cache_entry_count`, a direct `SELECT COUNT(*)` — `cache_info()` reports per-process hit/miss counters and is 0 for a cache written last week). A namespace that is non-empty and gains a row when the canary is encoded has never been verified against any checkpoint, so it is refused outright and named for deletion. An empty namespace is the ordinary first run and still passes. New test `test_a_cache_written_before_the_canary_existed_is_refused`; I confirmed it fails without the fix.

**P2 — Avoid recommending query_prompt for QdrantHybridIndex.** Correct, and worse than cosmetic: `VectorBlocker.stream()` forwards `query_prompt` to `search_all()`, which this same PR made raise `NotImplementedError`, so the warning's *primary* remedy converted a running hybrid configuration into a crash. The message now leads with "drop `prompt_name` … (works with any index)" and states the constraint on the other remedy explicitly — "FAISSIndex does, QdrantHybridIndex.search_all() does not and will raise". A test asserts the ordering, so the safe remedy cannot drift back below the unsafe one.

**P2 — Include failed custom models in the ladder denominator.** Correct at both sites. `measured` is derived from `ok` rows, so a custom `--models` name that *failed* dropped out of the denominator while still appearing in the Failures table — the ladder would appear to shrink because a run went badly. The recommendation site now counts `MODELS | models` (`models` is every model in `current`, any status) and the "What did not run" site `MODELS | measured | stale | failed`.

Full suite: 4625 passed / 2 skipped. `ruff check`, `ruff format --check`, `mypy src` all clean.

fxd24:
@codex review

Follow-up at `a30a8d8` — two further defects in the legacy-cache refusal itself, both found by trying to prove it works against the real 1.8 GB ladder cache rather than by reading the code.

**The refusal could be bypassed by running it twice.** It asked *is the canary here?* by **encoding** the canary — answering the question by changing it. The refused run deposited a canary row, so a second run found one present and went straight to the drift comparison and passed. Now the key is read instead (`_canary_is_cached`, via the cache's own `_hash_text` so there is exactly one implementation of the key, not a second one free to drift), and nothing is written on the refusal path. A test runs the refusal twice and asserts the entry count is unchanged.

**The refusal was retroactive with no way out.** Measured, not assumed: all six real namespaces hold 70,804–141,608 vectors and none carries a canary, so every one would be refused and the ladder would demand a full re-measure of a cache written hours ago from the checkpoints as they are now. `--trust-existing-cache` vouches for one, loudly — it logs what is being asserted and by whom, pins the canary, and the *next* run is checked normally. A test proves that is a one-time adoption rather than an off switch: after adopting, a checkpoint swap still raises.

Worth flagging in case it is useful signal: your P1 and this follow-up are the same defect at three depths — a check whose observation of the failure was itself unobservable. The canary was vacuous on a legacy cache; the fix for that was bypassable by repetition; the fix for *that* would have forced discarding a known-good cache. Each layer only showed up when the next one was tested against real data.

`ruff check`, `ruff format --check`, `mypy src` clean; 339 passed across the affected test trees.

fxd24:
@codex review

All four addressed at `b9c7e5e`, each reproduced before fixing.

**P1 — Propagate cache-integrity failures before replacing rows.** Correct, and this was the worst one in the PR: it is a data-loss path, not a reporting nit. `_build_embedder` sits inside `evaluate_model_on_benchmark`'s blanket handler, so the refusal became a `status="failed"` row; `main()` persists it and `merge_rows()` voids every previously recorded cell for that (model, benchmark) — erasing good measurements from the tracked `_rows.jsonl` — and `run_ladder.sh` sees exit 0 and commits the deletion. `StaleEmbeddingCacheError` now escapes. The distinction it encodes: every other exception there is a fact about the *model* (it did not load) and belongs in a row; a cache refusal is a fact about the *harness* and must abort. A control test keeps "just re-raise everything" from passing — an ordinary model failure must still be recorded.

**P1 — Restrict cache adoption to a single model.** Correct. `--models` defaults to all 14 and the flag was forwarded to each, so one invocation would permanently pin all six unverified namespaces from a flag whose own docstring says it vouches for one. Now `parser.error` unless exactly one `--models NAME` is given; verified by running it.

**P1 — Don't recommend a reranker that discards the prompt.** Correct, and I had verified the wrong half. In `QdrantHybridRerankingIndex.search_all` the dense side is served from vectors cached at `create_index()` and the sparse side always encodes with `prompt=None`, so the reranking embedder is the only stage a prompt can reach — and `FastEmbedLateInteractionEmbedder.encode` ignores it, as its own source comment states. So the refusal's suggested alternative restored the exact silent no-op the refusal exists to prevent.

Fixed as a **checked capability** rather than by editing the prose, per your suggestion: `honours_prompt = False` on that embedder, and `search_all` refuses when its reranker does not honour prompts. It defaults to `True` so an unknown or duck-typed embedder is trusted rather than broken — `VectorIndex` is a structural protocol and a fail-closed default would start raising on every test double. Four tests, including a premise guard that fails if the flag ever disappears from the production embedder (without it the check can never fire in production, which is the vacuous-gate shape again).

**P2 — Reject non-finite canary vectors.** Correct. `NaN > tolerance` is `False`, so a non-finite vector made the guard *accept* a cache it could not compare. Non-finite is now maximal drift — unusable is not the same as equal.

Both new gates were watched failing: neutering the re-raise and the finiteness branch fails exactly the two tests written for them.

`ruff check`, `ruff format --check`, `mypy src` clean; 346 passed across the affected test trees; full suite running.

fxd24:
@codex review

Both addressed at `8b25da1`, and the first was verified end-to-end rather than reasoned about.

**P1 — Preserve cache refusals in the outer ladder driver.** Correct, and this is the finding I most needed: re-raising inside Python *moved* the deletion rather than stopping it. `run_ladder.sh` treats any non-zero exit as "the process died", calls `record_process_failure`, strips every existing row for the model and commits the replacement — so the integrity refusal still destroyed exactly the measurements it exists to protect, one layer out.

The refusal now exits with a reserved code (`EXIT_CACHE_INTEGRITY = 3`; 1 is an uncaught exception, 2 is argparse), and the driver bails on that code *before* the row-deleting path, recording nothing.

Proved against the real cache, not asserted: `--models intfloat/e5-base-v2 --cache-dir tmp/embedder_ladder_cache` exits **3**, names the 70,804-vector namespace, leaves the cache at exactly 70,804 entries (the idempotence fix holding on real data), and leaves `docs/research/*_rows.jsonl` untouched. A test pins the shell literal and the Python constant together — they are in different languages and nothing else stops one drifting — and asserts the bail is ordered before `record_process_failure`.

**P1 — Validate prompt-specific cache entries with the canary.** Correct. The key is `(text, prompt)`, so an unprompted canary vouched for one partition while the `instruct` and `documented` arms read prompt-keyed entries the check never looked at — and a change confined to explicit prompt handling (`include_prompt`, prompt-token pooling) leaves bare embeddings byte-identical, which is precisely the drift that would slip through.

The canary now covers `None` plus every query prompt the run will actually use, threaded down from `prompt_arms`. One distinction worth naming: a partition with no canary yet is **cold, not legacy** — the unprompted canary already vouched for the namespace — so adding an arm pins a new canary instead of refusing, which keeps the legacy gate meaning what it says. The control test is the hole itself: with `prompts=[None]` the same prompt-only drift goes unnoticed, so the added coverage is demonstrated rather than claimed.

Full suite 4635 passed / 2 skipped / exit 0; `ruff check`, `ruff format --check`, `mypy src` clean; 532 passed across the affected trees.

fxd24:
@codex review

All four addressed at `a38a8a8`.

**P1 — Refuse populated prompt partitions before pinning a canary.** Correct, and it is the vacuous-canary defect one partition over: a namespace whose *unprompted* canary is present still had prompted partitions that nothing had ever vouched for, and my loop called them "cold" and pinned a fresh canary over stale vectors. The same gate and the same remedy now apply per partition, counting rows via the cache's own `prompt` column.

Worth naming because it nearly bit: the count uses `WHERE prompt IS ?`, not `=`. SQL equality against NULL is NULL, never true, so `prompt = NULL` counts zero rows in the *unprompted* partition — the largest one — and would report every cache as empty, silently disabling the gate. There is a test pinning that specifically.

The coarse namespace-level gate is now gone rather than kept alongside: the loop subsumes it and is strictly more precise, and the old gate had a bug of its own — it `return`ed on the adopt path, skipping the loop entirely, so `--trust-existing-cache` would leave prompted partitions unpinned.

**P2 — Forward the cache-adoption flag from the ladder driver.** Correct: the refusal recommended a flag the documented driver could not accept, and would have silently read as a wait PID. Now `LADDER_TRUST_EXISTING_CACHE=1`, forwarded like `LADDER_DEVICE`, documented in the usage block, and named in the refusal message itself.

**P2 — Avoid ranking symmetric document prompting as universally worse.** Correct, and this is the better version of the fix I made two rounds ago: I replaced a wrong *name* heuristic with a narrower name heuristic, when the resolved values were available all along. It now compares the resolved prompt *values* — a custom name whose prefix equals the query prefix is symmetric and stays silent — and falls back to the name only where values are not exposed. The message no longer asserts asymmetry or a quality ranking; it states what is observable, that both sides carry the prefix, and lets the reader say which recipe they meant. Control test: a custom name resolving to a *different* prefix still warns, so "compare values" cannot degrade into "never warn".

**P2 — Update the experiment description for the unprompted hybrid arm.** Correct. The header still said all three approaches use instruction prompts — and still promised "Instruction prompts: Expected +1-5% quality improvement" from a run that measures no such thing, since no two arms share an index. Both corrected; the header now carries the confound the labels and saved JSON already did.

`ruff check`, `ruff format --check`, `mypy src` clean; 355 passed across the affected trees.

fxd24:
@codex review

Round 8 findings addressed on `4c945056fd4d15a76eab62007c7974faf9f63edc`:

- **P1 `LADDER_TRUST_EXISTING_CACHE`** — `${VAR:+...}` tests non-emptiness only, so `=0`/`=false`/a typo all expanded to `--trust-existing-cache`, adopting unverified vectors on exactly the run where the operator was turning adoption *off*. Now an exact-value `case`: `1` vouches, unset is off, anything else exits 2 naming the value. Verified: `LADDER_TRUST_EXISTING_CACHE=0` prints REJECTED.
- **P2 "offline"** — a cold run downloads weights and datasets. Reworded to "$0 and keyless", with a note that offline holds only once the `uv` and HF caches are warm.
- **P2 stale "worse than prompting neither side"** — narrowed in all three places (`vector.py`, `embeddings.py`, `CHANGELOG.md`) to a prefix that *differs* from the query-side one. Binding the query prefix on both sides is now described as the documented symmetric recipe that `_binds_the_query_prefix` deliberately accepts in silence.
- **P2 unknown licence** — split into a third bucket. Unknown still fails closed (excluded from the default candidates), but renders under "Unclassified licences — verify before shipping, in either direction" with no OSI verdict attached. New test falsified by removing the split.

The report is byte-identical after `--render-only`: no measured model is unclassified today, so the licence fix is on a latent path.

fxd24:
@codex review

Round 9 addressed on `88dd1c56a1be52c1cf8bd3b9ac9b3c0091603c57`:

- **P1 driver-level cache adoption** — correct, and the interesting part is *why* the Python guard did not catch it: each child invocation receives exactly one `--models`, so the single-model safeguard passed on all 14 of them while the driver blessed the whole ladder. A guard one level below the decision cannot see the decision. The check now lives where the model list is chosen: `LADDER_TRUST_EXISTING_CACHE=1` requires `LADDER_MODELS` to name exactly one model. Watched it fail (default 14 models → exit 2, message names the count) and pass (one model → "vouching for BAAI/bge-small-en-v1.5's cache").
- **P2 reranking-index migration advice** — correct. `QdrantHybridRerankingIndex.search_all()` refuses under the documented production reranker (`FastEmbedLateInteractionEmbedder`, `honours_prompt = False`), so the CHANGELOG was routing users from one exception to the next. Removed it from the alternatives, said plainly why it is not one, and gave the reranking refusal its own CHANGELOG entry. The same stale recommendation was still in `hybrid_vector_index.py`'s `search_all` docstring — two lines above the message that contradicts it — and is fixed too.
- **P2 prefetch depth** — correct, and it was under-counted in a way that made the list read as complete. `crossencoder_vs_hybrid` now names retrieval depth as a third confound, and **`crossencoder_vs_faiss` moved into `confounded_deltas` as well**: it was only ever matched on the prompt axis, and calling it "prompt-matched" beside two "confounded" rows implied the rest was controlled. Added `retrieval_depth_by_arm` (searched vs reported per arm, so `prefetch=100` cannot read as "returns 100"), and the console now prints a legend stating what "prompt-matched" does *not* control, with the rank 21–100 reachability spelled out.

`mypy src` clean, `ruff` clean, 368 passed across `tests/core/indexes` + `tests/examples`.

fxd24:
@codex review

Round 10 addressed on `15cdaf8c493561676d6c8fa09b7a4e3eee753739`. Both findings are the same defect shape as the licence one you caught in round 8 — **absence of evidence rendered as evidence** — which is worth naming, because it means the fix is a class of check, not two patches.

- **P2 recommendation gated on available intervals** — correct, and worse than it reads: `merge_rows()` clears every challenger's `vs_reference_*` *by design* whenever the reference is remeasured, so the empty-`wins()` state is routine, not exotic. The report then said "the measurement cannot tell them apart" and "therefore keep the current default" — a finding of equivalence **and a decision** read off a comparison that never ran. `has_comparison()` now separates "compared and did not win" from "never compared"; the second says it cannot rank at all, calls itself a missing measurement rather than a tie, and explicitly refuses to be read as a reason to keep the default. Partial coverage is named too ("covers N of the M OSI models"). Watched fail by forcing `compared = list(osi)`, with a control test for the genuine zero-win field so an always-"missing" check cannot pass.
- **P2 hard-coded legacy-cache claim** — correct. The supported `--trust-existing-cache` path falsifies "every namespace behind this table is in that state" on the very next run, and since `LadderRow` carries no cache-verification field the renderer cannot know either way. I took the second of your two options: it now describes the *policy* and points at where the live answer actually is (a namespace holding the canary row has been pinned; one without it has not, and a run meeting an unpinned namespace exits 3), instead of asserting a state it cannot observe. Persisting the state on rows would be the stronger fix but would render as "unknown" for all 300 existing rows, which is a bigger change than the finding warrants.

CI green on the previous head (8/8 real checks; `Sourcery` is rate-limited, a no-op rather than a pass). Local full suite at `4c94505`: 4646 passed, 2 skipped, 95.88% repo-wide. `ruff` and `mypy src` clean; 67 ladder tests pass.

fxd24:
@codex review

Round 11 addressed on `4b7da6988f8ba6cbd7684a893bc2b3eac8b31a2b` — **partly accepted, one half disputed with evidence**.

**Accepted (the reachable defects).** Zero wins had two causes rendering identically: beaten on every benchmark, or never compared. The table printed `0 of 5` for both, so a missing experiment read as five losses. And win counts from models compared on *different* benchmark subsets were ranked against each other as if they were the same experiment.

- `comparable_count(model)` gives each row its own denominator — `1 of 2` from a model compared twice is not the claim `1 of 5` makes.
- A model with no intervals renders `not compared` and is excluded from `ranked`, so it cannot pad the field.
- The best-candidate sentence now states its coverage whenever that is partial: *"best of N of the M OSI models, on P of the Q benchmarks — the best of what was measured, not of the field"*.

Falsified by forcing the table branch off; the new test fails. The real report changes by one clause (current coverage is full, 5 of 5).

**Disputed:** *"this branch can declare a 'Best OSI-licensed candidate' over models that were never compared"* — I could not reach that state. `ranked` is `(len(wins), best_delta, name)` and `max` orders on win count first. An uncompared model enters with `len(wins) == 0`, so it can only win `max` when **every** compared model also has zero wins — and that state routes to the preceding `best_count == 0` branch, which names no model at all. I tried to write a failing test for it before fixing and could not construct one; the falsification that *does* fail is the table-denominator branch. If there is an input shape I am missing, name it and I will fix it rather than argue.

I fixed the ranking-membership half anyway, because "cannot currently produce a wrong name" is a weaker property than "cannot rank an unmeasured model", and it costs one line.

CI green on the previous head (8/8 real; `Sourcery` is rate-limited — a no-op, not a pass). `mypy src` clean, `ruff` clean, 240 passed in `tests/examples`.

fxd24:
@codex review

Round 12 addressed on `c4a8343276a1c8a93030f73f4c0328f3bbe40a7f`. Both correct, and the second one is a defect **this PR introduced two rounds earlier** — worth calling out, because it is the shape where a fix creates the next bug.

- **P1 ranking across incomparable coverage.** Correct: 2 wins from 5 attempts outranking 1 from 1 ranks the *size of the experiment*, not the model. `comparable_benchmarks()` now returns a set, and the ranking runs on the intersection across all compared models; `common_wins()` counts only there. Disjoint coverage now **names no winner at all** rather than picking one from incomparable experiments. You also caught that the coverage caveat could vanish exactly when it mattered — the winner sentence now states the shared-set size *unconditionally* ("3 of the 5 benchmark(s) all 5 compared models share"), so the number is in the claim rather than in an optional footnote. Falsified by reverting `common_wins` to `wins`: the shared-set test fails; a second test covers the disjoint case.
- **P2 unusable recovery command.** Correct, and self-inflicted: round 9 added the single-model guard, and this message — printed *mid-sweep*, where `MODELS` is all 14 — kept suggesting `LADDER_TRUST_EXISTING_CACHE=1 run_ladder.sh` with `LADDER_MODELS` unset, so following the advice hit the new guard and exited 2. Verified end to end rather than by reading: a real exit-3 now prints `LADDER_MODELS='intfloat/e5-base-v2' LADDER_TRUST_EXISTING_CACHE=1 …`, and running that exact string clears the guard and proceeds. Cache state confirmed unchanged by the check (70,804 rows, canary still unpinned).

Real report moves by one sentence — current coverage is full (5 of 5 shared), so no ranking changed.

CI green on the previous head (8/8 real checks). `mypy src` clean, `ruff` clean, 242 passed in `tests/examples`.

fxd24:
@codex review

Round 13 addressed on `f5fa50ca0fd593fe8ad6d6b7f044aee07c2f8ed0`. Both correct, and both were **created by round 12's own shared-set fix** — narrowing the ranking narrowed `best_count`, but two sentences were still written for the old, wider count. Noting the pattern because it is now three rounds running where the previous fix produced the next finding.

- **P1 no-win verdict vs. exclusive wins.** Correct: with `best_count` taken on the shared set, a challenger winning a benchmark the others were never measured on scores zero, and the branch then claimed no model beats the reference "on any benchmark" — contradicting the CI-clear row directly above it. That case now makes **no default recommendation at all** and names the model: keeping the default would need those wins to be absent (they are not), and promoting on them would need them to be comparable (they are not either). Every remaining zero-win sentence is scoped to "the N shared benchmark(s)".
- **P2 tie resolved by name.** Correct. The name is in the sort tuple for byte-stable re-renders, not to adjudicate — but `max` does not know that, so an exact tie handed the title to the lexicographically greatest model, which reports the sort order as a measurement. Ties are now detected on the **measured** part of the key `(count, delta)` and rendered as co-winners with no unique winner, pointing at cost/licence/memory/latency as the tiebreak this table cannot supply.

Both falsified: disabling either branch fails exactly its own test and nothing else. The report is byte-identical — coverage is currently full and there is no tie, so neither new branch fires on the real data.

CI green on the previous head (8/8 real checks). `mypy src` clean, `ruff` clean, 244 passed in `tests/examples`.

fxd24:
@codex review

Round 14 addressed on `cc4de5ee5aad9e91e151c241114b27691d63247b`. Both correct.

- **P1 one probe is not proof of equivalence.** The `max_seq_length` example is exactly right, and it is the case a fixed short string can *never* see. Two changes, both reusing existing machinery rather than adding a parallel staleness check:
  - `max_seq_length` is folded into the canary **text**, so a changed value changes the cache *key*; the partition then holds vectors with no canary under the new key and the legacy gate already built refuses it. That was the cheapest correct place to put it — no schema change, no second thing to keep in sync.
  - a **second probe of ~1024 words**, past the 128/256/512-token limits real sentence-transformers checkpoints use, so it is truncated by all of them and moves when the cut moves — covering a checkpoint that re-cuts without saying so in config, which the key alone cannot catch.
  - `_canary_is_cached` now requires **all** probes, not any: a partition pinned before a probe existed has never been verified against what that probe covers, which is precisely what "legacy" already means here. In practice this costs nothing — every real namespace is already unpinned.

  On your "include the immutable revision" alternative: I did not take it, and I want to be explicit rather than quietly narrow the fix. Putting the revision in the namespace invalidates every cached vector to close the same hazard, and still only answers *is this the same Hub commit* — it cannot see a local runtime change like this one. The probe pair answers *does this checkpoint still produce these vectors under this configuration*, which is what the measurements actually depend on. If you think the revision belongs in the identity **as well**, say so and I will add it; I left it out to avoid a full re-encode for a strictly weaker signal.

- **P2 clear loss vs. inconclusive.** Correct, and the interesting part is that the *recommendation* was right while the *reason* was its opposite — which is how a wrong reason survives review. Clear losers (any shared interval entirely below zero) are now named as a **measured loss**; genuinely inconclusive comparisons keep the "cannot tell them apart" wording; and a mixture says which models are in which group.

Falsified: reverting `_canary_texts` to a single bare probe fails the `max_seq_length` test; emptying `clear_losers` fails the clear-loss test. Both have controls that fail if the new branch fires unconditionally. Report byte-identical. Re-verified end to end against the real cache: exit 3, 35,402 unverified vectors named, nothing written.

`mypy src` clean, `ruff` clean, 248 passed in `tests/examples`; CI green on the previous head (8/8 real checks).

fxd24:
@codex review

Round 15 — both P2 findings on `cc4de5e` fixed in `0639d3c`.

**1. `docs/research/20260727_embedder_ladder.md:19` — describe both required cache canaries.** Correct, and the consequence you named is the harmful one: the paragraph told an operator that a namespace holding "the canary row" is pinned, which is now exactly the state that gets refused as legacy. Fixed in the generator (`examples/research/embedder_ladder.py`), not in the `.md` — the report is generated, so a hand-edit would be overwritten by the next `--render-only`; regenerated in the same commit. The paragraph now says there are **two** probes and both are required, why the long (~1024-word) probe exists at all (a short probe cannot observe an input-selective change such as a shortened `max_seq_length`), that both probe texts carry `max_seq_length` in their cache key so retuning it reads as unpinned rather than silently matching, and that a namespace written by the earlier single-probe version is missing the long probe and is deliberately refused as legacy — distinct from a pre-check namespace with no canary at all. The live-answer sentence now reads "a namespace holding **all** the current probe rows has been pinned; one missing any of them has not."

**2. `examples/research/embedder_ladder.py:1676` — separate exact ties from inconclusive intervals.** Correct, and I verified the state you cited rather than taking it on the argument: querying the ladder rows for `vs_reference_ci_low == vs_reference_ci_high == 0.0` returns **6 rows, all on `fodors_zagat`**. `loses_clearly()` only recognised intervals strictly below zero, so a degenerate `[0, 0]` fell through to prose asserting the measurement "cannot tell them apart" — the opposite evidential claim to `_ci()`, which renders `[0, 0]` as `(exactly 0)`. A new `ties_exactly()` classifies a model whose every shared-set headline row is `[0, 0]` as an exact tie and reports it as a measured result; a genuinely wide interval straddling zero still reads as inconclusive.

**Falsification (each new branch watched failing its own test, with a control):** forcing `exact_ties: list[str] = []` fails exactly `test_an_exact_zero_interval_is_a_result_not_a_gap` while `test_a_wide_interval_around_zero_is_still_inconclusive` keeps passing — so the new branch is load-bearing and is not firing unconditionally.

`tests/examples/test_embedder_ladder.py` is now 78 tests; `mypy src` clean; `ruff check .` clean; `tests/examples` 250 passed, 1 skipped.

One note on bookkeeping: the PR's inline-thread list also still carries findings from earlier rounds (the `blocking_evaluation_with_instructions.py` console labelling, the `_build_embedder` integrity-error escape, the single-model trust guards in both the Python entry point and `run_ladder.sh`, the licence three-bucket split). Those threads reference earlier commits, not `cc4de5e`, and each was fixed in the round it was raised — they read as open only because the threads were never resolved. If any of them is still wrong *as landed*, please re-post it against `0639d3c` and I'll take it.

fxd24:
@codex review

Round 16 — all three findings on `0639d3c` fixed in `5b7ad70`. All three were real, and the third one had a live victim in the test suite exactly as you described.

**1. `examples/research/embedder_ladder.py:667` — preserve checkpoint load failures during cache fingerprinting.** Correct, and the blast radius is worse than a wrong error message: `_vector_affecting_runtime()` caught every exception from `_get_model()` and substituted `max_seq_length=None`, which is a *different canary key*, so a model with a populated cache was refused as stale — a missing checkpoint or a broken dependency reported as a corrupt cache, exiting with `EXIT_CACHE_INTEGRITY`, which `run_ladder.sh` reads as "abort the whole sweep" instead of "record one model failure". The load error now propagates and `evaluate_model_on_benchmark`'s handler records it as the model failure it is. An **absent** `max_seq_length` still yields `None` — that is the genuinely duck-typed case the long probe exists for, and every fake in the test file is in it, so I added a control for it.

**2. `examples/research/embedder_ladder.py:1689` — preserve exact ties when clear losses are also present.** Correct: `clear_losers` and `exact_ties` were mutually exclusive branches, so a single interval below zero swept every degenerate `[0, 0]` into "the rest are genuinely inconclusive, with intervals spanning zero" — asserting uncertainty about the one comparison the measurement actually resolved. They are now a flat three-way partition (clear loss / exact tie / inconclusive), each reported on its own terms. I kept the original inconclusive prose verbatim so the report does not lose the "'indistinguishable' is not a reason to move" nuance while fixing the classification.

**3. `src/langres/core/indexes/reranking_vector_index.py:413` — mirror the reranker prompt refusal in its fake index.** Correct. `FakeHybridRerankingVectorIndex` now takes `reranking_honours_prompt`, defaulting to **`False`** rather than `True`, because that is what actually ships: `FastEmbedLateInteractionEmbedder` declares `honours_prompt = False` and is the only late-interaction embedder in the package. I checked before choosing the default — no existing test passed a prompt to this double — so defaulting to the shipped configuration costs nothing and closes the gap for callers who never opt in.

That change immediately surfaced the victim you predicted: `tests/core/blockers/test_vector.py::test_stream_with_fake_hybrid_reranking_index_end_to_end` built the default double *with* `query_prompt` set and passed — green for a configuration that raises against the production index. It now names `reranking_honours_prompt=True` (that configuration is what the test is about), and a new counterpart test asserts the **default** double refuses that same `VectorBlocker` configuration. `CHANGELOG.md` records the new default, since the double is public.

**Falsification — each new branch neutered in turn, each failed exactly its own test with its control still passing:**

| neutered | fails | control still passes |
|---|---|---|
| `exact_ties` gated on `not clear_losers` | `test_an_exact_tie_survives_a_clear_loser_in_the_same_field` | `test_an_exact_zero_interval_is_a_result_not_a_gap` |
| the swallow restored in `_vector_affecting_runtime` | `test_a_dead_checkpoint_is_a_model_failure_not_a_cache_refusal` | `test_an_embedder_with_no_max_seq_length_still_yields_a_fingerprint` |
| the fake's guard forced to `if False` | `test_the_default_double_refuses_a_query_prompt` | the other 3 in that class |

`mypy src` clean; `ruff check` + `ruff format --check` clean; `tests/core` + `tests/examples` 1842 passed, 1 skipped.

fxd24:
@codex review

Round 17 — all three findings on `5b7ad70` fixed in `9c2b17f`. All three were real.

**1. `examples/research/embedder_ladder.py:1793` — gate opt-in recommendations on measured evidence.** This is the sharpest of the three, and it is the same defect class this PR has been chasing throughout: absence of evidence rendered as evidence. "Recommended as a documented opt-in" was emitted off the **licence bucket alone**, so a checkpoint compared and beating the reference nowhere — or never compared at all — still read as recommended. The phrase was two statements fused: *documented opt-in* is the exposure **mechanism** the licence requires, while *recommended* is a claim only the measurement can make. Now three states — a win recommends it; a comparison with no win says the sweep measured none and that opt-in is the required mechanism if you use it anyway; no interval says so explicitly as a **missing measurement, not a verdict**.

**2. `src/langres/core/blockers/vector.py:485` — qualify reranking prompt support conditionally.** Correct, and it was actively harmful because the docstring is the constructor's own documentation: it said `QdrantHybridRerankingIndex` "honours it at the stage that decides the final order", which is true of the reranking pass and false of the configuration that ships — `FastEmbedLateInteractionEmbedder` declares `honours_prompt = False`, and `search_all` now raises for exactly that pairing. A caller following the docs was walked into a runtime failure. It now says support requires a prompt-honouring reranker, that the only late-interaction embedder in this package is not one, and names both ways out.

**3. `examples/research/blocking_evaluation_with_instructions.py:341` — stop attributing the confounded result to reranking.** Correct: the console credited "full cross-attention" and "reranking adds value" while the CrossEncoder arm differs from the hybrid arm in **three** ways simultaneously — reranker, prompt regime (hybrid is unprompted), retrieval depth (`CROSSENCODER_PREFETCH=100` vs `PREFETCH_LIMIT=20` before reranking). The saved JSON already recorded all three under `confounded_deltas`, so the console was claiming more than the file did. Both branches now report an **arm-level** result naming the three confounds — the same treatment the hybrid-wins branch already got.

**Falsification:** forcing the licence-only verdict back (`if True` in place of `if won`) fails exactly `test_a_restricted_model_with_no_win_is_not_recommended` and `test_a_restricted_model_that_was_never_compared_says_so`, while the control `test_a_restricted_model_that_did_win_is_still_recommended` keeps passing — so the gate is load-bearing and does not fire unconditionally.

`mypy src` clean; `ruff check` + `ruff format --check` clean; `tests/core` + `tests/examples` 1845 passed, 1 skipped.

fxd24:
@codex review

Round 18 — the finding on `9c2b17f` is fixed in `05ff6b8`, and you caught a good one: I fixed the restricted section and left its sibling with the identical defect, in the same commit.

**`examples/research/embedder_ladder.py:1842` — distinguish unclassified models that were never compared.** Correct. The unclassified-licence bullet still read "Measured ahead of `<ref>` on: no benchmark, with an interval clear of zero" whether the model was compared and won nothing or was never compared at all — absent evidence rendered as a result, which is the thing the previous round had just fixed one section above.

Rather than write the three-way branch a second time, I extracted it: `evidence_phrase()` now owns the won / compared-and-did-not-win / never-compared distinction and **both** sections call it. Each caller appends only its own consequence — the restricted section its opt-in exposure mechanism, the unclassified section its "read the model card, then record the identifier" instruction — so the licence claim and the measurement claim stay separate at both sites, and the measurement wording has exactly one place left to be wrong. That the branch existed twice is precisely why one copy went unfixed.

**Falsification:** collapsing the two no-win states back into one (`if True` in place of `if has_comparison(model)`) fails **both** `test_a_restricted_model_that_was_never_compared_says_so` and `test_an_uncompared_unclassified_model_is_not_reported_as_measured` from a single edit — which is the property sharing the helper buys — while the two middle-state controls (`..._with_no_win_is_not_recommended`, `test_a_compared_unclassified_model_that_never_wins_says_so`) keep passing.

`mypy src` clean; `ruff check` + `ruff format --check` clean; `tests/core` + `tests/examples` 1847 passed, 1 skipped.

fxd24:
@codex review

Round 19 — the finding on `05ff6b8` is fixed in `d7e0185`.

**`examples/research/embedder_ladder.py:1538` — include custom checkpoints in the missing-grid report.** Correct, and the sharpest part is that the comment sitting at the denominator site *described this exact hazard* ("point at a 'What did not run' section that only iterates `MODELS` and so cannot account for the difference") — and then only the denominator was fixed. A failed custom checkpoint printed "14 of the 15 models in the ladder" above a table that could not name the fifteenth; a partially measured one escaped the grid check entirely, leaving the report free to end with "Every model, benchmark and prompt arm in the ladder was measured."

Fixed with a single `_ladder_specs(rows)` helper — `MODELS` plus every model appearing in the rows, **including stale and failed ones**, so a run that went badly cannot shrink the ladder it was measured against. All three readers now share it: the recommendation denominator, the "What did not run" enumeration, and the per-model grid check. `_render_recommendation` takes that ladder rather than deriving its own from the current rows, which also removed a **second, subtly different denominator** I had not noticed: the old one was built from current-revision rows only, so a stale-only custom model was counted by one section and not the other — two totals for the same document.

**One existing test asserted the old behaviour** — that a stale custom model never appears anywhere below the header. That was the bug, not the contract: the property is that no stale *number* reaches a measurement table, not that the name is unmentionable. It now asserts exactly that, and additionally that the model **is** named under "What did not run" with its reason, which is the accounting that keeps the exclusion visible.

**Falsification:** pointing `never` and `gaps` back at `MODELS` fails all three of `test_a_failed_custom_checkpoint_is_counted_and_then_named`, `test_a_partially_measured_custom_checkpoint_is_grid_checked`, and `test_rows_from_an_older_metric_revision_are_excluded_and_named`.

Report re-rendered: the only content change is OSI-table row order (now the ladder's own size ordering). No measured number moved.

`mypy src` clean; `ruff check` + `ruff format --check` clean; `tests/core` + `tests/examples` 1849 passed, 1 skipped.

PR #244, opened 2026-07-27
Make main green honestly: fix the contract coverage gate, and the blindness that hid it for six days

fxd24:
`main` has been red for six days. This makes it green without moving the floor, and fixes the reason nobody saw it.

## Headline numbers

| measurement | value | floor | |
|---|---|---|---|
| **before** (CI, `82a222f`) | **97.28%** | 98 | ✗ |
| **after — tests only, this PR's pragmas reverted** | **98.16%** | 98 | ✓ |
| **after — as shipped** | **98.26%** | 98 | ✓ |
| after — fast suite (new per-PR gate) | 97.93% | 97.5 | ✓ |

**The `# pragma: no cover` change is not load-bearing: the gate clears 98 on the new tests alone.** Split of the +0.98pp, measured (see "Earned vs. arithmetic" below): **+0.88pp from new tests, +0.10pp from removing 12 shim statements from the denominator.**

`--fail-under=98` is untouched.

## Root cause: a code regression in #229 — not the include set widening

Two candidates had to be separated: (a) #229 added code that genuinely dropped coverage, or (b) the include set was widened without the floor being re-derived, so the floor had been nominally-98/actually-unmet for a while. The workflow's own `STALE-NUMBER NOTE` explicitly raised (b).

**It is (a), entirely.** `git diff 3dbd13d HEAD -- .github/workflows/test.yml` shows the `--include` list is **byte-identical** between the last green run and `HEAD`. The widening had already happened *and been measured green at 98.28%*. Nothing about the list changed when the gate went red.

The per-merge series, read out of the `Coverage gate` step of each merge commit's own CI run:

| commit | PR | gate TOTAL | stmts | miss |
|---|---|---|---|---|
| `3dbd13d` | #217 | **98.28%** | 8024 | 105 |
| `c270522` | **#229** | **97.27%** | 8883 | 178 |
| `94df55e` | #230 | 97.16% | 8968 | 187 |
| `82a222f` | tonight's 5 | 97.28% | 9023 | 178 |

**#229 alone accounts for the whole drop** (−1.01pp of a −1.00pp total). It merged ~2,000 net new lines into gated `core/` paths at roughly 90% coverage:

```
core/_model_run.py    +241 stmts  100%    -> 89.85%
core/_artifacts.py     +86 stmts  100%    -> 89.10%
core/registry.py       +76 stmts  97.44%  -> 84.09%
tracking/runs.py      +141 stmts  100%    -> 95.99%
core/op_adapters.py    +84 stmts  100%    -> 97.47%
```

#230 added `core/indexes/qdrant_dense_index.py` (−0.11pp). Tonight's five merges moved it **up** 0.12pp — they inherited the red, they did not cause it.

The `STALE-NUMBER NOTE` at the gate step promised the TOTAL "is recomputed once this lands". It never was. It has now been, and is replaced with the resolved finding.

## Why six days passed

`test-full` — the only job that runs this gate — is gated `if: github.event_name != 'pull_request'`. On PR #229 the check list read:

```
lint pass · test pass · test-core-only pass · test-finetune pass
test-full  SKIPPING        <- the only job running the contract gate
```

It merged fully green with the gate never having executed once, and `main` then absorbed six more merges. Step 7 of `test-full` is the *sole* failure on `main`; steps 1–6 and every other workflow (Lint, CodeQL, Docs, Scorecard) are green.

### 1. Real tests for contract code #229 shipped untested (the bulk)

- **`core/registry.py` — the `@register_op` fail-closed contract.** Every validation branch was untested: missing `from_config`/`config`, a non-Pydantic `config_model`, a `config_model` without `extra="forbid"` or `strict=True`, duplicate role, duplicate `op_type`, a trusted lazy role whose module registers nothing, a component handed to a component-free Op. These are what stop a saved artifact smuggling unvalidated keys into `from_config`. **84.09% → 97.73%**
- **`core/_artifacts.py` — the OpSpec guards.** Five built-in loaders' missing-component refusals, the missing-param-schema refusal, the undeclared-component refusal on save, the not-a-Stage refusal on load, and the `calibrator_score` round-trip. **89.10% → ~99%**
- **`tracking/runs.py` — the run-snapshot contract.** Immutability (`popitem`, `|=`, nested freezing) and JSON-safety (sets and non-finite numbers refused), plus the fingerprint's deliberate asymmetry: cluster order is traversal noise and normalises away, token order is content and does not. **95.99% → 99.20%**
- **`core/method_registry.py` — every optional builder's missing-extra `ImportError`.** CI installs `--all-extras`, so all five "you need the `[llm]`/`[trained]` extra" branches are unreachable there and all five were untested. They are the point of the extras split. **92.76% → 100%**
- **`ArtifactSource`** had **no test anywhere**: a Hub artifact source must be pinned to an immutable 40-character commit SHA, never a branch or tag that re-points later.
- **`SpendCappedMatcher`'s group drain** — a group is one paid call, so the cap must return the whole group in `partial_judgements`, and must not pull past `group_end` (which would fire the *next* group's paid call for output nobody keeps).
- **`LinkVerdict.__bool__`** — without it every verdict is truthy, so the documented `if model.compare(a, b):` reports a match for a **non-match**. A wrong answer, not a missing feature.
- **`Blocker.schema`** returns `None` when the registry no longer knows the name (the normal state for an inferred schema in a fresh process) instead of raising from a property callers treat as cheap.

### 2. Six of the last seven W2 shims get the pragma their five siblings already carry

`core/{debugging,diagnostics,runs}.py` and the three `core/trackers/*` adapters are `# TEMPORARY: deleted by the W2 sweep` re-export shims. A re-export owns no contract, and the module each redirects to is *itself* inside this gate at 100%. Five sibling shims (`harvest`, `anchor_store`, `benchmark`, `canonicalizer`, `review`) already carry `# pragma: no cover`, and the gate's own comment already documents this as the intended treatment. Each is commented with why and when it goes away.

#### Earned vs. arithmetic — the numbers, because pragma raises a percentage without testing anything

Re-measured on the shipped tree from **one** full-suite `.coverage` (4657 passed, 2 skipped, 18m43s), reporting three times against that same data file with the pragma tokens mangled and restored, so nothing but the pragma differs. `coverage.py` applies exclusions at *report* time, so this needs no re-running:

- **A — as shipped: 98.26%** (9011 stmts, 98 miss).
- **B — this PR's six shim pragmas neutralised: 98.16%** (9023 stmts, 110 miss) — **still passes 98**. Cost of the exclusion: **0.10pp, 12 statements.**
- **12 of 12** of those statements were uncovered beforehand (all six shims sat at 0.00%), so this is pure denominator reduction — no already-covered line was pragma'd.
- **Attribution of the +0.98pp: tests +0.88pp (~90%), pragma +0.10pp (~10%).** The test figure is approximate — it compares two different trees (`82a222f` vs here), not a controlled A/B.

So the pragma is hygiene, not the fix. Revert it and this PR still turns `main` green.

**The honest count-everything figure, for completeness: 97.20%** (9141 stmts, 204 miss) — every `# pragma: no cover` in `src/` neutralised, which is 130 statements across 34 files, and it sits **below the 98 floor**. That is not this PR's doing and is not proposed as the gate: **118 of those 130 pre-date this PR**, and most are the ordinary kind — `Protocol` method bodies (`core/op.py` 7, `core/embeddings.py` 7, `core/indexes/vector_index.py` 6, `core/fit.py` 5), the real QLoRA `train` that only runs in `test-finetune`/on GPU (`training/finetune.py` 38), and platform- or extra-specific `ImportError` branches. It is recorded so the floor is read for what it is — a ratchet over a denominator that excludes those — rather than as a claim that ~98% of every line in the contract is executed.

### 3. The positioning fix — the gate now runs per-PR

Fixing the number without fixing the blindness leaves the next regression equally invisible.

**Measured first, because the existing justification was stale by 4×.** The workflow said slow tests "cost ~10min on every PR"; that number is what justified keeping the gate off PRs:

| | measured |
|---|---|
| `test-full` | **~38–41 min** (green run `3dbd13d` 37m39s; run `30307689766` 41m) |
| per-PR `test` job | **~6 min** (PR #229's own run: 5m53s) |
| the gate step itself | **~12 s** |

The cost is the *suite*, never the gate. So running `test-full` on every PR is genuinely too slow (6× PR wall-clock), but the gate itself is nearly free — and the per-PR `test` job already writes `.coverage` (pyproject addopts carry `--cov=langres`).

**So the gate now runs in both places, over one include list, at two floors:**

| where | suite | floor | role |
|---|---|---|---|
| `test` (**every PR**) | fast | **97.5** | early warning, ~12s |
| `test-full` (merge to main) | full | **98** | authoritative |

**Why two floors is not the "two numbers drift" disease.** The thing that rots is the include *list*, and there is now only **one** of it — hoisted to `env.CONTRACT_COVERAGE_INCLUDE`, read by both gates, verified byte-identical to the previous inline list. The floors differ because the two gates measure genuinely different data, and that difference is small and measured: full **98.26%** vs fast **97.93%** on the same tree — the slow tests are worth **0.33pp** of this include list, so the fast number is a close proxy, not a different quantity wearing the same name. 97.5 is that value rounded down the same way the 98 floor was.

`.claude/rules/testing.md` is updated to match — its copy of the include list had *already* gone stale (it omitted `metrics/*`, `tracking/*` and the four `curation/` files), which is the duplication this PR removes.

## What I deliberately did not do

- **Did not lower `--fail-under`.** That is the exact disease this repo has already shipped once (a step named "95% coverage gate" enforcing 90).
- **Did not touch `core/_model_run.py` (29 uncovered), `core/_model_state.py` (8), `training/fit_report.py` (9)** — another agent holds those files. They are now the largest uncovered blocks in the gate and the natural next PR.
- **Did not touch `core/indexes/*`** (T2's files) or `core/op.py` (T1's).
- **Did not add tests that assert nothing** to reach the number.

## Verification

```
uv run pytest -m "not integration and not finetune" --cov-report=xml   # 4657 passed, 2 skipped, 18m43s
uv run coverage report --include="$CONTRACT_COVERAGE_INCLUDE" --precision=2 --fail-under=98
                                                                       # TOTAL 98.26%
uv run ruff format --check . && uv run ruff check . && uv run mypy src # clean, 215 files
```

*Isolation, since it is the whole basis of these numbers: the full-suite run above had **no** concurrent pytest in the same worktree, verified by sampling every 30s for its full 18m43s — 38 samples, zero foreign pids, zero unobservable probes. An earlier attempt overlapped ~9 minutes with an orphaned run left by a previous agent in this worktree and was discarded unread: two `uv run` in one worktree contend on `.venv`/`.coverage` and yield a plausible but falsely LOW number. Two cross-checks that the tree matches CI: this run's include-list denominator is **9011 statements**, identical to the 9011 CI reports for the fast suite on the same commit; and the full suite's **98** misses are strictly fewer than the fast suite's **126**, the direction coverage monotonicity requires.*

## Summary by Sourcery

Add regression tests around core contract behaviors and make the coverage gate visible and reliable both per-PR and on merge while keeping the existing 98% floor.

New Features:
- Cover the core op registration and serialization contracts, including @register_op validation, OpSpec guards, and serializer registration invariants.
- Add tests for tracking run snapshot immutability, JSON-safety, and dataset fingerprint normalization.
- Add tests for artifact provenance pinning, blocker schema resolution behavior, matcher spend capping semantics, and LinkVerdict truthiness and repr.
- Verify method registry optional builders surface actionable extras installation errors.

Enhancements:
- Mark temporary core re-export shims with no-cover pragmas to exclude throwaway compatibility layers from the contract coverage denominator.
- Document the dual contract coverage gates and their maintenance rules in the testing guidelines, including measured timings and rationale for separate floors.

CI:
- Run the contract coverage gate against a shared include list from both the fast per-PR test job (at a 97.5% early-warning floor) and the full test-full job (at the authoritative 98% floor), and centralize the include configuration in workflow env to prevent drift.

Tests:
- Extend core and tracking test suites to exercise previously untested validation, error handling, and contract paths across registry, serialization, method registry, spend caps, blockers, run snapshots, and result models.

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -81,8 +81,8 @@ code. Verify as you go.** Read before writing tests or running the suite.
   the same include list (same denominator), so **full-suite coverage ≥ fast-suite
   coverage, always** — the per-PR gate can never demand what the authoritative one
   does not already meet. The observed gap is small, which is why 97.5 is close
-  rather than slack: measured on one tree, 2026-07-28, full suite **98.05%** vs
-  fast suite **97.87%** — **0.18pp**.
+  rather than slack: measured at `1b24a2a`, 2026-07-28, full suite **98.26%** vs
+  fast suite **97.93%** — **0.33pp**.
 - **Timings, measured 2026-07-28** (the previous "~10min" claim was stale by 4x and
   was the justification for keeping the contract gate off PRs entirely):
   `test-full` **~38–41 min**, the per-PR `test` job **~6 min**, the gate step
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -34,12 +34,15 @@ code. Verify as you go.** Read before writing tests or running the suite.
 - Mark slow tests with `@pytest.mark.slow`, integration tests with `@pytest.mark.integration`
   - **Mark any heavy test `@pytest.mark.slow`** (loads embedding/ML models, runs
     torch inference, etc.). CI runs the **fast** subset (`not slow`) on every
-    PR; the **slow** tests + the coverage gates run on **every merge to main
-    (push)** + **on demand** (`workflow_dispatch`) via the `test-full` job. So
-    per-PR CI does not gate coverage or exercise slow ML paths — mislabeling a
-    heavy test as fast slows every PR, and the coverage floors are verified on
-    each merge to main, not per-PR. Run the full suite locally
-    (`uv run pytest`) before merging a change to ML/embedding code.
+    PR; the **slow** tests run on **every merge to main (push)** + **on demand**
+    (`workflow_dispatch`) via the `test-full` job. So per-PR CI does not exercise
+    slow ML paths — mislabeling a heavy test as fast slows every PR. Run the full
+    suite locally (`uv run pytest`) before merging a change to ML/embedding code.
+  - **Coverage is gated in both places, but not on the same numbers** — see the
+    two-gate table below. The *contract* floor runs on **every PR** (early
+    warning, 97.5) and authoritatively on **each merge** (98). The *repo-wide*
+    90% floor is enforced only on `test-full`, because the fast suite omits the
+    slow tests that are the only cover for some ML paths.
 - Run tests: `uv run pytest` (pre-push hook runs non-slow, non-integration tests automatically)
 - Check coverage: `uv run pytest --cov`; keep `core/**` in the 95–100% tier
   (the repo-wide gate is a relaxed 90% floor — see `pyproject.toml`)
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -73,10 +73,16 @@ code. Verify as you go.** Read before writing tests or running the suite.
 - **Why two floors is not the "two numbers drift" disease.** The thing that rots
   is the include *list*, and there is only one of it. The floors differ because the
   two gates measure genuinely different data: the per-PR job excludes the slow ML
-  tests, so paths covered only by them read as uncovered there. Measured on one
-  tree, 2026-07-28: full suite **98.05%**, fast suite **97.87%** — the slow tests
-  are worth only **0.18pp** of this include list, so the fast measurement is a close
-  proxy, not a different quantity wearing the same name.
+  tests, so paths covered only by them read as uncovered there.
+  The lower floor is **structural, not tuned**: the per-PR selection
+  (`not integration and not slow and not finetune`) differs from `test-full`'s
+  (`not integration and not finetune`) *only* by `not slow`, so its tests are a
+  strict **subset**. Coverage is monotone in tests executed and both gates score
+  the same include list (same denominator), so **full-suite coverage ≥ fast-suite
+  coverage, always** — the per-PR gate can never demand what the authoritative one
+  does not already meet. The observed gap is small, which is why 97.5 is close
+  rather than slack: measured on one tree, 2026-07-28, full suite **98.05%** vs
+  fast suite **97.87%** — **0.18pp**.
 - **Timings, measured 2026-07-28** (the previous "~10min" claim was stale by 4x and
   was the justification for keeping the contract gate off PRs entirely):
   `test-full` **~38–41 min**, the per-PR `test` job **~6 min**, the gate step
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -43,42 +43,45 @@ code. Verify as you go.** Read before writing tests or running the suite.
 - Run tests: `uv run pytest` (pre-push hook runs non-slow, non-integration tests automatically)
 - Check coverage: `uv run pytest --cov`; keep `core/**` in the 95–100% tier
   (the repo-wide gate is a relaxed 90% floor — see `pyproject.toml`)
-- **Two gates run on `test-full`, and they are not the same number as this
-  policy.** The repo-wide floor is 90% (`--cov-fail-under` in `pyproject.toml`).
-  The contract additionally has its own path-scoped gate
-  (`coverage report --include="src/langres/core/*,src/langres/report/*,src/langres/autoresearch/*,src/langres/training/*,src/langres/optimize.py"
-  --precision=2 --fail-under=98` in `.github/workflows/test.yml`). That 98 is a
-  **regression ratchet** pinned just under the measured value (98.84% at
-  `ba4b1b7`), not the policy — the *target* remains 95–100%. It exists because
-  the repo-wide 90% floor sits ~8 points below actual coverage, so the contract
-  could be quietly declassified with CI green throughout. Raise it as the real
-  number climbs; if it blocks legitimate work, lower it deliberately rather than
-  deleting it.
-  **When the contract moves to a new package, extend that `--include` glob** or
-  the gate silently stops covering it. (`report/`, `autoresearch/`, `training/`
-  and the `optimize.py` facade are in the glob for exactly that reason — they
-  carry public surface (`EvalReport`, `optimize()`/`score_blocking`,
-  `derive_threshold`/`Calibrator`/`FitReport`/`QLoRA`), so letting them fall out
-  would un-gate contract code while making the remaining number look *better*.
-  `training/*` was added when the fit family moved there out of `core/` — a
-  coverage-neutral relocation of already-gated code.)
-  **`src/langres/optimize.py` is listed as a file, not folded into a `*` glob**:
-  the facade is a module and the engine beside it is the `autoresearch/` package,
-  and coverage compiles a trailing `/*` to `optimize[/\\].*` — which matches a
-  directory's contents and can never match `optimize.py`. An `optimize/*` entry
-  here matches **nothing** and drops the facade silently. (Trailing `/*` *is*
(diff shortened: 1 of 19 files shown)


fxd24:
@codex review

fxd24:
@codex review

fxd24:
## Disposition of the four Codex findings, and an independent re-measurement

### The numbers, re-verified from a second isolated run

The contract figure was re-measured from a **fresh** full-suite run (contaminated
inputs deleted first; the run was the only pytest in this worktree for its whole
duration, checked by resolving every live pytest's `cwd`):

| | statements | branches | contract coverage |
|---|---:|---:|---:|
| **A** — as shipped | 9011 | 2616 | **98.26%** |
| **B** — this PR's six shim pragmas neutralised | 9023 | 2616 | **98.16%** |

`A` reproduces the committed figure **exactly**, so the number in the workflow
comment is confirmed rather than merely asserted.

The six shims measure **21 statements, 12 uncovered** with the pragma off, which
is where `9011 + 12 = 9023` comes from — the "12 statements across 6 shims" claim
checks out. Separately, `core/trackers/__init__.py` measures **9 statements at
100%**: it is genuinely *tested*, not excluded.

**Conclusion unchanged and now independently supported: exclusion is worth
0.10pp, and the gate clears 98 without it.**

### P2 — "Cover the lazy tracker shim instead of excluding it" — already fixed

Raised against `07498e75`. That shim was subsequently un-pragma'd and covered by
a real test. Verified above by measurement: 9 statements, 0 missed, 100%.

### P2 — "Reconcile the per-PR coverage guidance" — respectfully, a misreading

The comment reads lines 37–41 of `.claude/rules/testing.md` as saying per-PR CI
does not gate coverage. Those lines say per-PR CI *"does not exercise **slow ML
paths**"* — a statement about which **tests run**, not about gating. The bullet
immediately below opens: *"**Coverage is gated in both places**, but not on the
same numbers"*, then gives the two-gate table. No change made.

### P2 — "Update the fast-test job description" — same, a truncated read

The quoted range 79–83 ends mid-argument. Line 83 says the **repo-wide** floor is
NOT enforced — correct, `--cov-fail-under=0` does exactly that. Line 86 continues:
*"the `langres.core` **CONTRACT** floor **IS** enforced, at a lower early-warning
97.5"*, and line 89 adds that the authoritative 98 floor still runs on
`test-full`. The header already states the semantics the YAML implements. No
change made.

### P2 — "Avoid gating slow-only changes with fast coverage" — real, and bounded

This one stands, and deserves a number rather than a reassurance.

Contract code covered *only* by a `slow`-marked test does read as uncovered in
the fast job. Measured, that gap is **0.33pp** (full 98.26% vs fast 97.93%), and
the fast floor sits at **97.5** — about **0.43pp** of margin. Against ~11.6k
measurable units, roughly **50 additional statements/branches** reachable *only*
from slow tests would have to land in one change to trip the early warning.

Two things keep that acceptable rather than merely unlikely: this gate is an
**early warning**, not the contract floor — `test-full`'s 98 remains
authoritative — and a trip is a *loud, cheap, per-PR* signal at the point where
it is diagnosable, versus discovering it after merge, which is the failure this
PR exists to end. The floor is deliberately **not** being lowered to widen that
margin; that would trade a real signal for a quiet one.

If a slow-only change ever does trip it, the fix is to mark the covering test
fast where that is honest, or accept the red and confirm against `test-full` —
not to relax the number.

PR #246, opened 2026-07-28
feat(embeddings): default embedder -> intfloat/e5-base-v2 (measured, OSI-licensed)

fxd24:
## What

`DEFAULT_EMBEDDING_MODEL` moves from `all-MiniLM-L6-v2` to **`intfloat/e5-base-v2`** (MIT, 109.5M params, 768-dim), chosen from the committed embedder ladder over the OSI-licensed field only.

## The measured numbers

From `docs/research/20260727_embedder_ladder.md` — mean **per-record candidate recall** at k=20, `prompt_arm="none"`, paired per record and bootstrapped by gold cluster. Ahead of the incumbent on **3 of 5**, behind on **none**:

| benchmark | Δ per-record recall vs `all-MiniLM-L6-v2` | 95% CI | clusters |
|---|---:|---|---:|
| `wdc_computers` | **+0.1528** | [+0.1254, +0.1797] | 877 |
| `abt_buy` | **+0.0324** | [+0.0216, +0.0451] | 1,012 |
| `walmart_amazon` | **+0.0159** | [+0.0078, +0.0246] | 846 |
| `amazon_google` | +0.0056 | [-0.0002, +0.0116] (spans 0) | 995 |
| `fodors_zagat` | +0.0000 | [+0.0000, +0.0000] (exactly 0) | 112 |

Absolute recall at k=20, and the cost of that ceiling (`cands/recall`, lower is cheaper) — e5 is cheaper on **all five**:

| benchmark | MiniLM recall | e5 recall | MiniLM cands/recall | e5 cands/recall |
|---|---:|---:|---:|---:|
| `abt_buy` | 0.9349 | 0.9674 | 11,985 | **9,095** |
| `amazon_google` | 0.8266 | 0.8324 | 27,853 | **25,749** |
| `fodors_zagat` | 1.0000 | 1.0000 | 4,880 | **4,431** |
| `walmart_amazon` | 0.8654 | 0.8810 | 53,658 | **52,776** |
| `wdc_computers` | 0.5770 | 0.7264 | 56,103 | **43,834** |

### How partial the field is

**7 of the 14 models in the ladder have rows at `metric_revision` 1.** The other 7 (`Alibaba-NLP/gte-base-en-v1.5`, `nomic-ai/nomic-embed-text-v1.5`, `BAAI/bge-large-en-v1.5`, `mixedbread-ai/mxbai-embed-large-v1`, and the three `Qwen/Qwen3-Embedding-*`) are recorded "not run". **This is the best *measured* model, not a survey** — absence from the table is not evidence against a model.

Two more scope limits, from the ladder's own caveats: the metric is **blocking candidate recall, not end-to-end F1**; and `fodors_zagat` is saturated (every model scores 1.0000), so it settles nothing. Separability AUC is higher for e5 on all five, but the ladder explicitly calls AUC a floor check near its ceiling, not a fine-grained ranking — so it is not offered as evidence here.

### Why not `google/embeddinggemma-300m`

It measured better overall (`wdc_computers` +0.2079, `abt_buy` +0.0392) but ships under the **Gemma Terms of Use — not OSI-approved**, carrying a prohibited-use policy that survives redistribution. langres is Apache-2.0; a default must not push terms onto users who never chose them. It remains a documented opt-in. (It is also **worse** than the incumbent on `amazon_google`, -0.0294 [-0.0425, -0.0163] — there is no single winner.)

## The prompt recipe: unmeasured, so not wired — and it's `query: ` on both sides

Two corrections here, both to premises I was handed. I read the checkpoint's own model card and the committed rows rather than relying on either.

**1. The ER-relevant recipe is the *symmetric* one.** The card (FAQ 1, L2685-2694) splits by task shape: `"query: "`/`"passage: "` for **asymmetric** retrieval (L2690), but `"query: "` on **both sides** for **symmetric** tasks — semantic similarity, paraphrase retrieval (L2692). Entity resolution is symmetric — records against records, no "passage" side — so the retrieval pair is the wrong recipe to reach for. `core/blockers/vector.py:36-48` (from #242) already documents this shape, names e5-base-v2, and accepts it *without* warning.

**2. It is still unmeasured, so it does not ship.** Verified three independent ways:

1. **The winning rows are `prompt_arm: "none"`** — neither side prefixed. The `+0.1528` on `wdc_computers` is `vs_reference_delta` on the `none` row.
2. **e5-base-v2 has no `documented` arm at all.** Row counts by arm: `{none: 20, instruct: 20}`. Only `google/embeddinggemma-300m` has one. The harness's `ModelSpec("intfloat/e5-base-v2", license="mit")` passes no `documented_arm`, and `arms_for()` only adds that arm when one is declared.
3. **The checkpoint registers no prompts.** Every e5 row records `registered_prompts: []`, and `config_sentence_transformers.json` is a `.no_exist` entry in the Hub cache — the checkpoint ships none, so nothing was applied implicitly either.

### The card and the ladder don't conflict — together they set a floor

The card warns that running e5 with **no** prefix degrades it. The ladder ran it with no prefix. So:

> **e5-base-v2 beat the previous default by +0.1528 / +0.0324 / +0.0159 while running in the configuration its own card calls degraded.**

That makes the measured margin a **lower bound**, and the prefix *uncollected upside* — not a correctness problem. Shipping the symmetric recipe now would instead put a configuration nobody measured behind the numbers above, which is the specific way this task fails silently.

The caution isn't theoretical: the ladder's boxed warning records that a query-side prefix **hurt** several models, and e5's measured `instruct` arm (a generic query-side instruction — neither recipe) moved nothing measurable on any of the five benchmarks (four intervals span 0; `fodors_zagat` exactly 0).

**Follow-up, not a blocker:** add a symmetric `"query: "` arm for this model to `examples/research/embedder_ladder.py`, re-run, and move the constant's configuration and its numbers together. Pinned meanwhile by `test_the_default_embedder_ships_with_no_prompt_recipe`, which asserts the default binds no `prompt_name` and no `prompts`, in the object and in its serialized config.

Worth stating explicitly: shipping the symmetric recipe **would** be expressible today without new API (`VectorBlocker` + `SentenceTransformerEmbedder(prompts=..., prompt_name="query")`). The blocker is the missing measurement, not a missing seam.

## Migration, and a latent cache bug this change would have activated

**A normally saved artifact keeps working — do not discard it.** `FAISSIndex.config()` serializes the embedder's own `ComponentSpec`, `from_config()` reconstructs *that* embedder, and `load_state()` restores stored vectors verbatim without invoking it (`vector_index.py:342-392`). (My first draft claimed such indexes "will not load". That was wrong — caught in review, corrected.)

What actually changes:

- **Rebuilding an index against the new default** gives 768-dim vectors from a different model; it can't be mixed with or compared against old 384-dim ones.
- **`DiskCachedEmbedder` caches re-encode once** — because of the fix below.

**Fixed: `DiskCachedEmbedder` could silently serve a different model's vectors.** `_embedder_discriminator()` covered `prompt_name` / `truncate_dim` / `prompts` but **not the model**, and `namespace` defaults to `"default"` — so the ordinary `DiskCachedEmbedder(embedder=SentenceTransformerEmbedder(), cache_dir=...)` wrote every model into one `default.db` keyed on the bare text. Swapping the wrapped model returned the *previous* model's vectors at the previous dimensionality: silently on a warm cache, as a ragged 384/768 stack on a partial one.

This PR is what would have made that latent bug live — every user of the default keeps reading MiniLM vectors without touching a line of their own code. The model is now part of the key (`ModelRef` where present, so a `revision` pin counts). Caches from an identity-less embedder (`FakeEmbedder`) still hit; caches from a named model re-encode once.

To keep the old behaviour exactly, name the old model:

```python
SentenceTransformerEmbedder("all-MiniLM-L6-v2")   # previous default
```

Cost: 109.5M params vs 22.7M — bigger download, more memory, slower encoding. The ladder's `index build (s)` column **cannot** price this and I have not quoted it: every `all-MiniLM-L6-v2` row read from the embedding cache (`enc 0`) while the e5 rows encoded for real, so the two columns measure different things.

## Changed

- `core/model_ref.py` — the constant, the evidence table, and the prompt-recipe reasoning with its reproduction path.
- `core/embeddings.py` — the cache-key fix above; docstrings that stated the default; plus `FakeEmbedder`'s "(matches all-MiniLM-L6-v2)" parenthetical, which was literally true but implied the fake tracks the default (value left at 384).
- `core/matchers/cascade.py` — `embedding_model_name` was a **fourth hard-coded copy** of the old literal that the `model_ref` consolidation missed; now reads the shared constant.
- `docs/research/20260727_embedder_ladder.md` + `examples/research/embedder_ladder.py` — the ladder called MiniLM "langres's current default" in two places, with the wording hard-coded in the renderer so re-rendering regenerated the stale claim. Now described as "this ladder's reference model, and langres's default when these rows were measured" — true permanently, since it's a fact about the measurement rather than the current config. Fixed in the renderer, then regenerated with `--render-only`: that re-render changed exactly those 3 lines and reproduced every other line byte-identically, which also confirms the committed doc still matches the committed rows.
- `tests/` — the assertion pinning the literal, the no-prompt guard, and `TestCacheKeyCoversTheModelItself` for the cache fix.
- `CHANGELOG.md`.

**Deliberately not changed** (flagged rather than silently swept):

- `langres/data/*` benchmark loaders (`_EMBED_MODEL`, `_DEFAULT_EMBED_MODEL`, per-dataset factories) — these pin the model behind **published baselines**; retargeting them would silently invalidate committed results.
- `SearchSpace.embedding_model` / `BlockerOptimizer` — explicitly enumerated **search axes**, not the library default. Changing them would alter every default `optimize()` run and the documented `$0/offline` example in `docs/EXPERIMENTS.md`.

Both are defensible either way; say the word and I'll move them.

No non-research doc claimed the default by name — every `all-MiniLM-L6-v2` hit under `docs/` is an explicit argument in an example, or the `SearchSpace` field default, both still accurate.

## Verification

`ruff format --check` (572 files) pass · `ruff check` pass · `mypy src` (215 files) pass · fast suite **4711 passed, 2 skipped**

I did **not** touch `src/langres/resources/sentence_transformers.py` (owned by another stream this round), and did not need to.

**Note on the check list: Sourcery did not review this PR.** It posted "you have reached your weekly rate limit of 500000 diff characters" and its check reads `skipping`; its "Reviewer's Guide" comment is not a review. Codex did review, and its three findings are addressed above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,17 +48,47 @@ above was measured bare** — the ladder's `prompt_arm="none"`, with no
 ship a configuration nobody measured under the measured numbers' banner, so it is
 left off and pinned by test. Measure it in the ladder first if you want it.
 
-**Migration — if you have a saved vector index, it will not load.** The
-dimension changes 384 → 768, so an index built with the old default is
-incompatible with a freshly-constructed default embedder, and even at equal
-dimension the vectors would be from a different model. Either re-embed your
-corpus, or pin the old model explicitly and change nothing:
+**Migration.** A **normally saved artifact keeps working** — `Resolver.save` /
+`FAISSIndex.config()` serialize the embedder you built with, so `load` rebuilds
+*that* embedder (still `all-MiniLM-L6-v2`) and `load_state` restores the stored
+vectors verbatim without re-embedding. Do **not** discard those.
+
+What does change:
+
+- **Anything that rebuilds an index against the new default** now produces
+  768-dim vectors from a different model. Recall changes as measured above, and
+  such an index cannot be compared against, or mixed with, 384-dim vectors from
+  the old default.
+- **A `DiskCachedEmbedder` cache is re-encoded once** — see the entry below;
+  before that fix the same cache would have silently served the old model's
+  vectors.
+
+To keep the previous behaviour exactly, name the old model:
 
 ```python
 from langres.core.embeddings import SentenceTransformerEmbedder
 
 embedder = SentenceTransformerEmbedder("all-MiniLM-L6-v2")   # previous default
 ```
+
+### Fixed: `DiskCachedEmbedder` could serve a different model's vectors
+
+`_embedder_discriminator()` covered `prompt_name` / `truncate_dim` / `prompts`
+but **not the model**, and `namespace` defaults to `"default"` — so the ordinary
+`DiskCachedEmbedder(embedder=SentenceTransformerEmbedder(), cache_dir=...)` wrote
+every model's vectors into one `default.db` keyed on the bare text. Swapping the
+wrapped model then hit those entries and returned the **previous** model's
+vectors, at the previous model's dimensionality: silently on a fully warm cache,
+and as a ragged 384/768 stack on a partially warm one.
+
+The model now forms part of the key — the `ModelRef` where there is one, so a
+`revision` pin counts too. **Existing caches written by a named model are
+re-encoded once**; caches from an embedder with no model identity (e.g.
+`FakeEmbedder`) still hit, as before.
+
+Changing `DEFAULT_EMBEDDING_MODEL` in the same release is what turned this latent
+hazard into a live one — every user of the default would have kept reading
+MiniLM vectors without touching a line of their own code.
 
 Affected defaults: `SentenceTransformerEmbedder()`,
 `Resolver.from_schema(..., matcher="embedding")`, `VectorLLMCascade(embedder=...)`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,72 @@
 
 ## [Unreleased]
 
+### The default embedding model is now `intfloat/e5-base-v2`
+
+**This changes results, and it invalidates saved vector indexes.** Read the
+migration note below before upgrading if you have one on disk.
+
+`DEFAULT_EMBEDDING_MODEL` moves from `all-MiniLM-L6-v2` to `intfloat/e5-base-v2`
+(MIT). The embedder ladder (`docs/research/20260727_embedder_ladder.md`) measured
+it ahead of the incumbent on **3 of the 5** benchmarks and behind on **none** —
+mean per-record candidate recall at k=20, bootstrapped by gold cluster:
+
+| benchmark | Δ per-record recall | 95% CI | clusters |
+|---|---:|---|---:|
+| `wdc_computers` | **+0.1528** | [+0.1254, +0.1797] | 877 |
+| `abt_buy` | **+0.0324** | [+0.0216, +0.0451] | 1,012 |
+| `walmart_amazon` | **+0.0159** | [+0.0078, +0.0246] | 846 |
+| `amazon_google` | +0.0056 | [-0.0002, +0.0116] (spans 0) | 995 |
+| `fodors_zagat` | +0.0000 | [+0.0000, +0.0000] (exactly 0) | 112 |
+
+It is also a *cheaper* ceiling at that operating point, not just a higher one:
+candidates per unit recall fall on all five (`wdc_computers` 56,103 → 43,834;
+`abt_buy` 11,985 → 9,095).
+
+**What this is not.** It is a **blocking** result — candidate recall, not
+end-to-end F1 — and the field is the **7 of 14** ladder models with rows at
+`metric_revision` 1, so this is the best *measured* model, not a survey.
+`fodors_zagat` is saturated and settles nothing. The ladder's `index build (s)`
+column cannot be used to price the switch either: every `all-MiniLM-L6-v2` row
+read from the embedding cache (`enc 0`) while the e5 rows encoded for real, so
+those two columns measure different things. What is certain is that e5-base-v2
+is **109.5M parameters against 22.7M** and **768-dim against 384** — a bigger
+download, more memory, and slower encoding.
+
+**`google/embeddinggemma-300m` measured better overall and is deliberately not
+the default.** It ships under the Gemma Terms of Use, which is *not* OSI-approved
+and carries a prohibited-use policy that survives redistribution. langres is
+Apache-2.0; a default may not push terms onto users who never chose them. It
+stays a documented opt-in: `SentenceTransformerEmbedder("google/embeddinggemma-300m")`.
+
+**The new default ships with no prompt, on purpose.** E5's model card documents
+an asymmetric `"query: "` / `"passage: "` recipe, but the checkpoint registers no
+prompts (it ships no `config_sentence_transformers.json`) and **every number
+above was measured bare** — the ladder's `prompt_arm="none"`, with no
+`documented` arm for this model at all. Wiring the model-card prefixes in would
+ship a configuration nobody measured under the measured numbers' banner, so it is
+left off and pinned by test. Measure it in the ladder first if you want it.
+
+**Migration — if you have a saved vector index, it will not load.** The
+dimension changes 384 → 768, so an index built with the old default is
+incompatible with a freshly-constructed default embedder, and even at equal
+dimension the vectors would be from a different model. Either re-embed your
+corpus, or pin the old model explicitly and change nothing:
+
+```python
+from langres.core.embeddings import SentenceTransformerEmbedder
+
+embedder = SentenceTransformerEmbedder("all-MiniLM-L6-v2")   # previous default
+```
+
+Affected defaults: `SentenceTransformerEmbedder()`,
+`Resolver.from_schema(..., matcher="embedding")`, `VectorLLMCascade(embedder=...)`,
+the `embedding` method's `default_model`, and `CascadeChainMatcher` — whose
+`embedding_model_name` was a fourth hard-coded copy of the old literal and now
+reads the shared constant. **Not** changed: the pinned `all-MiniLM-L6-v2` in the
+benchmark loaders under `langres.data` (they back published baselines) and in
+`SearchSpace` / `BlockerOptimizer` (explicitly enumerated search axes).
+
 ### The better clusterer is now reachable (`clusterer=` opt-in)
 
 `CorrelationClusterer` measured better than the default transitive-closure
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@
 
 ### The default embedding model is now `intfloat/e5-base-v2`
 
-**This changes results, and it invalidates saved vector indexes.** Read the
-migration note below before upgrading if you have one on disk.
+**This changes results for any index built against the default.** Saved
+artifacts are *not* invalidated — they reload on the model they were saved with —
+so read the migration note below before re-embedding anything.
 
 `DEFAULT_EMBEDDING_MODEL` moves from `all-MiniLM-L6-v2` to `intfloat/e5-base-v2`
 (MIT). The embedder ladder (`docs/research/20260727_embedder_ladder.md`) measured
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,13 +40,24 @@ and carries a prohibited-use policy that survives redistribution. langres is
 Apache-2.0; a default may not push terms onto users who never chose them. It
 stays a documented opt-in: `SentenceTransformerEmbedder("google/embeddinggemma-300m")`.
 
-**The new default ships with no prompt, on purpose.** E5's model card documents
-an asymmetric `"query: "` / `"passage: "` recipe, but the checkpoint registers no
-prompts (it ships no `config_sentence_transformers.json`) and **every number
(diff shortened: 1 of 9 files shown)


fxd24:
Thanks — I verified all three against the source rather than taking them on faith, and **all three were right**. Fixed in `2a34435`.

### P1 — disk cache served the old model's vectors: confirmed, fixed

Verified: `_embedder_discriminator()` returned `""` whenever `prompt_name` / `truncate_dim` / `prompts` were unset, and `DiskCachedEmbedder.namespace` defaults to `"default"` (`embeddings.py:1320`). So the ordinary wrap wrote every model into one `default.db` keyed on the bare text.

This one mattered most, because **this PR is what makes the latent hazard live**: a user of the default changes no code, upgrades, and silently keeps reading 384-dim MiniLM vectors out of a warm cache.

The model is now part of the key — `model_ref` where present (so a `revision` pin counts, since the same name at a different Hub revision is a different checkpoint), falling back to `model_name`. Embedders with no model identity (`FakeEmbedder`) still return `""`, so their caches keep hitting and the two existing back-compat tests stay green unmodified. Caches written by a *named* model are re-encoded once — the price of no longer being silently wrong, and called out in the CHANGELOG.

Pinned by `TestCacheKeyCoversTheModelItself`: two models in one namespace don't share entries, the same model still does, a revision pin changes the key, and an identity-less embedder keeps its entries.

### P2 — saved-index warning was wrong: confirmed, rewritten

You're right and I was wrong. `FAISSIndex.config()` serializes the embedder's own `ComponentSpec`, `from_config()` reconstructs *that* embedder, and `load_state()` restores stored vectors verbatim without invoking it (`vector_index.py:342-392`). A normally saved artifact loads fine on the old model. I asserted a persistence claim without reading the persistence code.

The CHANGELOG now says plainly that saved artifacts keep working and should **not** be discarded, and narrows the real exposure to (a) rebuilding an index against the new default, and (b) the `DiskCachedEmbedder` case above.

### P2 — stale "current default" labels in the ladder: confirmed, de-rotted

Both sites (`:25`, `:244`) plus the two hard-coded renderer strings (`embedder_ladder.py:2025`, `:2435`) and an "incumbent" at `:1568`.

Rather than swap one stale label for another that rots the next time the default moves, I made the wording **independent of what ships**: `REFERENCE_MODEL` is now described as "this ladder's reference model, and langres's default when these rows were measured" — true permanently, since it is a fact about the measurement rather than about the current config.

Fixed in the renderer first, then regenerated the doc with `--render-only`. That re-render changed exactly those 3 lines and reproduced every other line byte-identically, which also confirms the committed doc still matches the committed rows.

---

`ruff format --check` (572) · `ruff check` · `mypy src` (215) · fast suite **4711 passed, 2 skipped** — all green locally.

Note for anyone reading the check list: **Sourcery did not review this PR** — it posted "you have reached your weekly rate limit of 500000 diff characters" and its check reads `skipping`. Its "Reviewer's Guide" comment is not a review.

fxd24:
## Correction to my own PR body: the untaken recipe is **symmetric `query: `**, not `query:`/`passage:`

I read the checkpoint's model card directly (`~/.cache/huggingface/hub/models--intfloat--e5-base-v2/.../README.md`) rather than relying on the framing I was given, and my earlier description of the recipe was wrong. Corrected in `4f68c27`.

**What the card actually says** (FAQ 1, L2685-2694):

> **Do I need to add the prefix "query: " and "passage: " to input texts?**
> Yes, this is how the model is trained, otherwise you will see a performance degradation.
> - Use "query: " and "passage: " correspondingly for **asymmetric** tasks such as passage retrieval in open QA, ad-hoc information retrieval. *(L2690)*
> - Use "query: " prefix for **symmetric** tasks such as semantic similarity, paraphrase retrieval. *(L2692)*

Entity resolution is **symmetric** — records against records, no "passage" side. So the recipe this model would want here is **`query: ` on both sides**, not the retrieval pair I named. `core/blockers/vector.py:36-48` (from #242) already documents exactly this shape, names e5-base-v2, and accepts it *without* warning (`prompt_name="query"`, `query_prompt` unset).

**The decision does not change, but the reasoning is now correct — and stronger.**

The ladder measured e5 **bare**: `prompt_arm="none"`, `registered_prompts: []` on every row, no `config_sentence_transformers.json` in the checkpoint (so nothing was applied implicitly either), and no `documented` arm for this model. The symmetric recipe is therefore **unmeasured in this repo** — switching it on would put a configuration nobody measured behind the numbers this PR cites.

The card and the ladder don't conflict; together they set a **floor**:

> e5-base-v2 beat the previous default by +0.1528 / +0.0324 / +0.0159 **while running in the configuration its own card calls degraded.**

That makes the measured margin a lower bound and the prefix *uncollected upside*, not a correctness problem. Collecting it is a measurement job: add a symmetric arm to `examples/research/embedder_ladder.py`, re-run, and move the constant's configuration and its numbers together.

Worth noting the caution is not theoretical — the ladder's own boxed warning records that a query-side prefix **hurt** several models, and e5's measured `instruct` arm (a generic query-side instruction, which is neither recipe) moved nothing measurable on any of the five benchmarks: four intervals span 0, `fodors_zagat` is exactly 0.

Updated in the constant's comment (`core/model_ref.py`), the guard test's docstring, and the CHANGELOG. **No source behaviour changed in this commit** — it is the rationale that was wrong, not the shipped configuration.

### Scope note

I did **not** touch `src/langres/resources/sentence_transformers.py` (owned by another stream this round) — and did not need to. Files changed on this branch:

```
CHANGELOG.md
docs/research/20260727_embedder_ladder.md
examples/research/embedder_ladder.py
src/langres/core/embeddings.py
src/langres/core/matchers/cascade.py
src/langres/core/model_ref.py
tests/core/test_cached_embedder.py
tests/core/test_embedding_model_ref.py
tests/core/test_method_registry.py
```

For the record: shipping the symmetric recipe *would* be expressible without that file (`VectorBlocker` + `SentenceTransformerEmbedder(prompts=..., prompt_name="query")`), so the blocker here is the **missing measurement**, not a missing API.

fxd24:
@codex review

Three findings from the previous pass are addressed in `2a34435` and the e5 prompt rationale corrected in `4f68c27`. Please re-review at HEAD — particularly the new cache-key discriminator in `src/langres/core/embeddings.py` (it now includes model identity, which invalidates existing named-model caches by design) and whether the CHANGELOG's saved-artifact migration text is now accurate.

fxd24:
Re-review triage at `4f68c27` — **1 of 3 was new and real; 2 are stale re-reports of findings already fixed in the commits under review.** I verified each rather than assuming.

### ✅ New and correct: CHANGELOG headline contradicted its own migration note

Right, and it was my own inconsistency: I corrected the migration body last round after your first P2, but left the headline saying *"it invalidates saved vector indexes"* — which the section below then contradicts. Exactly the kind of thing that makes someone discard a working artifact after reading only the bold line.

Fixed in `9d2a400`:

> **This changes results for any index built against the default.** Saved artifacts are *not* invalidated — they reload on the model they were saved with — so read the migration note below before re-embedding anything.

### ⏪ Stale: "Invalidate disk caches when changing the default model" (P1)

Already fixed in `87287fc`, which is in the reviewed range. The finding cites `embeddings.py:1395-1453` and the pre-fix behaviour ("omits the model identity and returns an empty discriminator"); the anchor has moved to `model_ref.py:144`.

Verified by running it, not by reading the diff:

```
same namespace?   default default True     # same db_path, as in the report
old discriminator: |||MODEL|||ModelRef(base='all-MiniLM-L6-v2', ...)|||NAME|||None|||DIM|||None|||PROMPTS|||[]
new discriminator: |||MODEL|||ModelRef(base='intfloat/e5-base-v2', ...)|||NAME|||None|||DIM|||None|||PROMPTS|||[]
KEYS COLLIDE?     False
```

Two `DiskCachedEmbedder`s sharing one `default.db` — the exact scenario described — no longer collide. Also pinned by `TestCacheKeyCoversTheModelItself` (4 cases: two models don't share, the same model still does, a `revision` pin changes the key, an identity-less embedder keeps its entries).

### ⏪ Stale: "Update the ladder's current-default labels" (P2)

Already fixed in `2a34435`, also in the reviewed range:

```
$ grep -n "current default\|current \`DEFAULT_EMBEDDING_MODEL\`" \
      docs/research/20260727_embedder_ladder.md examples/research/embedder_ladder.py
(no matches)
```

Both doc sites and both hard-coded renderer strings now read *"this ladder's reference model, and langres's default when these rows were measured"* — phrased as a fact about the measurement rather than the current config, so it can't rot again the next time the default moves. Fixed in the renderer first, then regenerated with `--render-only`.

---

**CI at `4f68c27`: all checks green** — `test (3.13)` 6m47s, `test-core-only`, `test-finetune`, `lint`, `build`, `docs-clean-install`, `CodeQL`, `Analyze (python)` all pass; `deploy` / `test-full` / `Sourcery review` skipped. `9d2a400` is a CHANGELOG-only commit on top.

PR #248, opened 2026-07-28
feat(data): register febrl_dedup (FEBRL3) — the portfolio's first task=dedup benchmark

fxd24:
## The gap

`BenchmarkTask = Literal["linkage", "dedup"]` was declared — and **all 10 registered benchmarks were `task="linkage"`**. So every number langres publishes is about cross-source linkage, while `dedupe()`, the primary shipped verb, was unmeasured. The `dedup` half of the Literal existed and nothing used it.

## Why not just relabel `febrl_person`?

That was the first thing I checked. FEBRL4 (what `febrl_person` vendors) *is* a dedup-shaped generation split into two files, and combining them back is defensible — but it makes a poor dedup benchmark:

- **every gold cluster is exactly size 2** (500 + 500 records, 500 links, zero singletons), so BCubed and transitive-closure behaviour are never exercised. A benchmark whose clusters are all pairs cannot tell a correct merge from an over-merge.
- its duplicates are all cross-source by construction, so the "no cross-source filter" property that distinguishes dedup would be cosmetic.

**FEBRL3** is the FEBRL family's actual deduplication dataset and is strictly better here: one 5000-record table, 2000 entities, cluster sizes **1–6** (835 singletons), 6538 gold pairs. Same generator, same tool, same licence clearance — no new dependency, no new licence question.

## Licence

`recordlinkage` **BSD-3-Clause** (the toolkit that redistributes FEBRL); upstream data under **ANUOS License 1.1** (MPL-style, redistribute-with-attribution). **Neither carries a NonCommercial term**, so it is compatible with Apache-2.0 — the same clearance recorded for FEBRL4 in `datasets/febrl_person/SOURCE.md`, and the same test that excluded CC-BY-NC OpenSanctions. Verified against the `recordlinkage` repository (https://github.com/J535D165/recordlinkage) and restated with attribution in `src/langres/data/datasets/febrl_dedup/SOURCE.md`. Data is fully synthetic (ANU frequency tables + injected corruptions) — no PII.

## Gold labels are NOT a transitive closure

Called out because the repo has been bitten by this (DBLP-Scholar's 37-record component). `gold_clusters.csv` gives `record_id,cluster_id` **membership** — the generator's own entity — so nothing is reconstructed by taking connected components of a noisy link file.

This is asserted rather than assumed. `recordlinkage` also ships a pairwise `links` index; the generation script verifies it is **exactly** the set of within-entity pairs implied by the record ids:

```
derived pairs 6538  shipped 6538  equal True
```

If a future `recordlinkage` release changed either, regenerating fails loudly instead of silently vendoring a different task.

## Ids are opaque on purpose

Upstream ids are `rec-<N>-org` / `rec-<N>-dup-<K>` — `N` **is** the entity number, shared by co-referent records. Keeping them would leak the label into the record dict. The fixture assigns `r0000`..`r4999` in shipped (already-shuffled) row order, so neighbouring ids are not co-referent either. Verified separately that `StringComparator.from_schema` selects exactly the 10 attribute fields — no `id`, no `embed_text` — so the headline numbers below are not a leak.

## The metric path is honest

`langres/benchmarks/runner.py` and `judge_eval.py` contain zero references to `source` / `cross_source` / bipartite structure, so nothing had to be forked. The one linkage assumption that *did* sit in the path was `_benchmark_utils.sweep_blocking_k`, which hardcoded `cross_source(...)` — a single-source corpus has no `.source` to compare. It now takes `cross_source_only: bool = True` (default **behaviour-preserving**), and `febrl_dedup` passes `False`. A test pins that `_bu.cross_source` raises on this corpus, so a future edit routing it through the linkage path fails loudly instead of returning a quiet 0.0 recall.

Scoring reads `bcubed_f1` / `cluster_pairwise_f1`. No pair-level metric is relabelled as a dedup score.

## Measured end-to-end

Blocking sweep (MiniLM + FAISS-cosine over `embed_text`, whole corpus, no cross-source filter), macOS/MPS:

| k | 5 | 10 | 20 | 30 | **50** | 80 |
|---|---|---|---|---|---|---|
| Pair-Completeness | 0.8790 | 0.9137 | 0.9345 | 0.9440 | **0.9552** | 0.9618 |

`DEFAULT_DEDUP_BLOCKING_K = 50` — smallest swept k clearing the 0.95 gate, and the knee (30 more neighbours buy 0.7pp). Dedup needs a larger k than the FEBRL4 linkage task (k=20 → 0.978) structurally, not by tuning: a record here can owe up to 5 gold pairs and must retrieve **all** its co-referents.

`run_methods(get_benchmark("febrl_dedup"), [...], seed=0)` on the held-out test split (1490 records / 597 gold clusters), $0 and offline, clusterer threshold tuned on **train** only:

| method | threshold | BCubed P | BCubed R | BCubed F1 | cluster-pairwise F1 | pair F1 |
| --- | --- | --- | --- | --- | --- | --- |
| `rapidfuzz` | 0.6 | 0.9977 | 0.9935 | **0.9956** | 0.9923 | 0.9777 |
| `embedding_cosine` | 0.8 | 0.5551 | 0.9415 | 0.6984 | **0.0247** | 0.7240 |

Two readings, both of which a size-2-clusters linkage benchmark cannot give you:

1. **The all-singletons sanity floor is BCubed F1 0.5721.** All-singletons BCubed precision is 1.0 and its recall is exactly `n_clusters / n_records`, so the floor is `2r/(1+r)` with `r = 597/1490 = 0.4007`. `rapidfuzz` is +0.4235 above it; `embedding_cosine` only +0.1263. The raw BCubed number flatters both. (The ratio is high because 835 of the 2000 entities — 42% of *entities*, 17% of *records* — are genuinely non-duplicated.)
2. **`embedding_cosine` collapses under transitive closure** — BCubed 0.6984 but cluster-pairwise **0.0247**. Its pair-level F1 (0.7240) reads as merely mediocre; the cluster view exposes that its false positives *chain* multi-record clusters into giant components. Structurally invisible on every existing benchmark.

`rapidfuzz` at 0.9956 is a property of the dataset (synthetic; `date_of_birth` + `soc_sec_id` are near-unique and lightly corrupted), documented as a regression **floor**, not a target.

## Caveats, stated in the loader and the docs

- Synthetic: corruptions are typo/OCR/field-swap shaped and do not stand in for real-world messiness.
- The 0.9552 measurement clears the 0.95 gate by only 0.5pp, which is **smaller than the known cross-platform MiniLM spread** (`febrl_person` measured that delta at over 1.6pp — its k=5 reads 0.966 on macOS and below 0.95 on Linux CI). So the slow test asserts a `PC_REGRESSION_FLOOR = 0.93` regression floor rather than re-asserting the gate: a gate that fails on a runner difference checks the runner, not the data. The floor still sits above the k=30 measurement (0.9440), so a real blocking regression — or a silently reintroduced cross-source filter — trips it.

## Also updated (same change, not a follow-up)

- `docs/BENCHMARKS.md` — the line "Every entry is a **cross-source linkage** benchmark" was **falsified** by this change. Rewritten; the table gains a `Task` column; new §2a documents the dedup task shape and the measured run.
- `tests/test_wheel_contents.py` `SHIPPED_NON_PY_FILES` + the `[tool.hatch.build]` comment + `BenchmarkDataNotFoundError`'s bundled-datasets hint — the fixture ships in the wheel (488K, permissively licensed) because excluding it would leave a pip install with no runnable dedup target.
- `examples/research/portfolio_profile.py` — its `large-component` caveat was already scoped to `task == "linkage"` with the comment *"No registered entry is a dedup set today… this keeps the rule honest for the one that eventually is."* That entry now exists; comment updated.
- `CHANGELOG.md`.

## Verification

- `ruff format --check` / `ruff check` / `mypy src` — clean.
- Fast suite: **4723 passed, 2 skipped** (`not integration and not slow and not finetune`).
- New slow tests pass locally: blocking regression floor (17s), the end-to-end `run_methods` BCubed run (111s), and the literal-`dedupe()` test (15s).
- **The `dedupe()` front door itself is asserted**, not only the harness's `resolve()`: `ERModel.dedupe` on the held-out split returns 344 clusters at BCubed P 0.9957 / R 0.9935 / **F1 0.9946**, cluster-pairwise 0.9900 — pinned by `test_the_dedupe_front_door_itself_scores_on_this_benchmark`.
- Wheel packaging gate: 4 passed.


fxd24:
@codex review

All four findings addressed in 89ebeca. Two of them were false claims I had made by assuming a mechanism instead of tracing it — thanks:

- **P1 portfolio artifacts** — regenerated `examples/research/results/portfolio_profile.json` (11 entries, tasks `['dedup','linkage']`) and updated `docs/research/20260727_portfolio_annotation.md`, whose §7 named "no registered dedup benchmark" as the portfolio's largest gap.
- **P2 harness cross_source** — confirmed correct. `grep -rn cross_source src/` returns nothing under `langres/benchmarks/`; `run_methods` scores whatever the blocker emits on every dataset. The filter is a blocking-recall *diagnostic* only. Docs and module docstring rewritten to say that.
- **P2 uniqueness claim** — confirmed false and self-contradicting (`dblp_scholar` 37, `amazon_google` 6, per the regenerated profile). Restated as single-source + membership gold.
- **P2 circular k pin** — the sharpest one. `blocking_k == DEFAULT_DEDUP_BLOCKING_K` moved with the constant, and the slow test swept at the constant, so editing it to 30 kept everything green. Now pinned to the literal `50` with a message telling the editor to re-measure. The floor itself stays at 0.93 deliberately (the measured 0.9552 clears the 0.95 gate by less than the known cross-platform MiniLM spread — `febrl_person` measured >1.6pp), and its docstring now states what it can actually catch rather than overclaiming.

fxd24:
Round 2 addressed in 245ffb3. Three of four were real; one I am consciously **not** fully adopting, with reasoning.

**Fixed — the false metric-visibility claim (P2, BENCHMARKS.md:139).** You are right and this was my error. Over-merging is *not* hidden by pair-sized gold: a false edge joining two size-2 entities already produces an oversized predicted component that BCubed and cluster-pairwise both penalise. What multi-record *gold* actually buys is the opposite direction — charging a matcher for **under**-merging a 6-record entity, which requires such an entity to exist. Corrected in all three places you named (BENCHMARKS.md, CHANGELOG.md, SOURCE.md).

**Fixed — task-as-dispatch (P2, registry.py:45).** Also correct, and worse than you flagged: that comment reintroduced the exact myth I had removed elsewhere in this same PR. Nothing in `benchmarks/runner` reads the field; both values get the same pair and cluster tracks. Restated as descriptive metadata naming its real consumers (discovery + the research profiler's linkage-scoped `large-component` rule).

**Fixed — annotation table (P2).** Updating only §7 was insufficient. §2 now has a `febrl_dedup` row and the surrounding denominators are corrected (nine loadable → ten, six of nine → six of ten, four unpublished → five). Also recorded that this dataset makes `optimize.py`'s 'all gold matches are inter-source' assertion false for **100%** of its pairs, not just a majority.

**Already fixed — P1 portfolio artifacts.** Re-flagged on 015a419, but done in 89ebeca; verified on HEAD: the tracked JSON has 11 entries including `febrl_dedup`/`dedup`. I also added a CI gate (`test_canonical_artifact_covers_every_registered_benchmark`) so this cannot go stale silently again — proven to fail against the pre-regeneration artifact (10 profiled vs 11 registered).

**Partially adopted — the 0.93 blocking floor (P2).** Your second half is fixed: the pin was circular (`blocking_k == DEFAULT_DEDUP_BLOCKING_K` moved with the constant), so it is now pinned to the **literal 50** with a message telling the editor to re-measure. Changing the constant to 30 now fails.

I am not raising the floor above the k=30 value (0.9440), and I would rather say why than quietly comply. The measurement is 0.9552 on macOS/MPS; the Linux value is **unmeasured**. `febrl_person` documents a cross-platform MiniLM spread of over 1.6pp (its k=5 reads 0.966 on macOS, below 0.95 on Linux CI). A floor at 0.945 is inside that spread, so it would risk failing on a runner difference rather than on a regression — and `test-full` runs only post-merge, where this repo has already had `main` sit red for six days. I have instead documented precisely what 0.93 can and cannot catch (gross failure: broken `embed_text`, corrupted fixture, a reintroduced cross-source filter reading ≈0.0) and what pins `k` separately. Once there is a Linux measurement, tightening it is a one-line follow-up.

PR #249, opened 2026-07-28
research: reproduce published blocking results (DeepBlocker, UniBlocker) with langres components

fxd24:
## Why

langres has produced plenty of its own blocking numbers and had never re-derived
anyone else's. (The one prior "replication", `langres.data.peeters`, replays an
archived set of model answers — it does not re-derive a metric from our own
models.) So there was no external evidence that the harness measures what it
claims to measure.

This is agenda item **E1** (`docs/research/20260717_research_agenda.md`): take
published blocking numbers, match the protocol as closely as our data allows,
re-derive them with langres components, and report the gap honestly.

The deliverable is confidence, not a matching number. Nothing here was tuned.
Two off-the-shelf sentence encoders, one serialization
(`concat_comparable_fields`), one exact-cosine index, one sweep — and the
harness was committed *before* it ran.

## The verdict splits in two, and they must not be run together

**1. Does the harness compute the literature's quantity? Yes, and this part is
unconditional** — it needs no assumption about anyone's gold pairs:

- The candidate set is theirs **to the pair**: `C = ∪_a topK(a → B)` gives
  `150 × 2,616 = 392,400` on `dblp_scholar` against DeepBlocker's printed
  `392.4k` (Table 6).
- `PQ = hits/(k·|A|)` is their formula: recomputing UniBlocker's own printed
  PC/PQ reproduces their PQ to 2 dp on five of six rows — and the sixth
  resolves only if `|A| = 533` rather than the 553 their Table 2 prints, i.e.
  the check is sharp enough to find a typo in their table.
- The internal `score_blocking` crosscheck reproduces the embedder ladder's
  published recalls **and** exact candidate counts (abt_buy 0.9368 / 10,897;
  amazon_google 0.8108 / 22,886).

**2. Do we reproduce their recall? Mostly, not uniformly, and never
unconditionally.**

- `dblp_scholar` **98.82%** vs their published **98.1%** (+0.72 pp) — untrained
  embedder against their trained Autoencoder, at their exact K and gold count.
- `dblp_acm` **96.76 / PQ 82.11** vs UniBlocker's STransformer **95.28 / 81.00**
  (+1.48 / +1.11 pp).
- `fodors_zagat` is **5.4 pp above** their STransformer row and unexplained — we
  land on the *other three* methods in their own table instead.
- The three product benchmarks are **4–7 pp apart as points** and only *overlap*
  once the gold-set interval is applied, which is not the same as agreeing.

**All of part 2 is conditional on our gold pairs corresponding to theirs.**
Neither paper publishes a pair list, so the condition is argued from provenance
and **verified nowhere** — including on the benchmarks whose pair *counts*
match, since an equal count is not an identical set. The report says this in
every place the numbers appear, and states plainly what the unconditional checks
*cannot* see: a wrong gold split — a different pair set with the same table sizes
and match count — passes all three of them unchanged.

## What is in here

| file | what it is |
|---|---|
| `docs/research/20260728_external_reproduction.md` | the report: papers, exact reported numbers with table citations, our measurements, the gap, the explanation |
| `docs/research/20260728_external_reproduction_reference.json` | every published number, hand-transcribed, each tagged with the table it came from |
| `docs/research/20260728_external_reproduction_rows.jsonl` | our 16 measured rows (full PC/PQ curves to k=150) |
| `docs/research/20260728_external_reproduction_crosscheck.json` | the 3-way protocol crosscheck |
| `examples/research/external_reproduction.py` | the harness (`--render-only` re-renders the report from the committed rows; needs no extras) |
| `tests/examples/test_external_reproduction.py` | 22 offline tests, all about *which rows a report is allowed to describe* |

## Papers actually retrieved

DeepBlocker (PVLDB 14(11), plus the authors' tech report), UniBlocker, SC-Block,
Sudowoodo, Zeakis et al.

**Not extractable:** DeepBlocker's per-solution R–C curves (Figures 6–8) and
Zeakis Figure 3 are plots only — so DeepBlocker publishes *no* numeric
off-the-shelf sentence-encoder baseline, which is why UniBlocker's Table 3
(which does) carries the like-for-like comparison.

Nothing was vendored. qcri/DeepBlocker is BSD-3-Clause; the PDFs stayed in
gitignored `tmp/` and only transcribed numbers were committed.

## Where the gaps come from

1. **Dataset version.** langres ships the DeepMatcher labelled splits, whose
   positives already survived somebody else's blocker, so our gold holds fewer
   pairs than the papers report on most datasets (amazon_google 1167 vs 1300;
   walmart_amazon 962 vs 1154) and the same count on two (dblp_scholar 5347,
   fodors_zagat 112). Every affected row is reported as a two-sided
   **conditional interval** — explicitly not a mathematical bound — with both
   counts printed.
2. **Gold-set definition, not model quality.** `score_blocking` scores the
   transitive closure of the gold clusters; the papers score the raw positive
   list. On amazon_google at k=20 the same embeddings give 95.80 (paper
   protocol) → 80.43 (closure gold) → 81.08 (`score_blocking`). That ablation is
   entirely langres-internal, so it establishes the closure explains the whole
   gap **between our two protocols** — it is a sufficient-looking explanation for
   the distance to the literature, not a demonstrated one.

## Reproducibility of the study itself

Both checkpoints are pinned to explicit Hub commits, the revision is part of the
disk-cache namespace and of every row, and an unpinned model runs uncached
rather than risking stale vectors. The rendered report is gated: it will not
emit its verdict unless every canonical cell is present at full k, the specific
crosschecks the prose cites survived admission, and **every one of the 14 figures
the prose quotes still matches the artifacts to 2 dp**. A partial or re-measured
render downgrades itself and says so.

## Review

Seven Codex rounds; **32 findings, all accepted and fixed** (Sourcery was
rate-limited for the week and never reviewed). The valuable half were prose: the
numbers were never wrong, but the sentences describing what they support kept
being stronger than the evidence — "strict subset" where only counts were known,
"the same blocker measured two ways" for a local-only ablation, "a wrong split
could not survive those checks" when it plainly could. Each round is a separate
commit with the finding and the reasoning; per-round dispositions are in the
thread. Round 7 came back clean.

## Cost

$0. No paid API calls. Two local sentence-transformer checkpoints, exact cosine
via `IndexFlatIP` over L2-normalized vectors (no ANN approximation loss).


fxd24:
@codex review

fxd24:
## Disposition of the four Codex findings

All four accepted and fixed in `a65348d`. Two of them were the repo's recurring
"a claim with no gate behind it" shape, so each fix ships with the check that
would have caught it.

### P1 — `uv run --extra semantic` (accepted, real)

Confirmed rather than assumed: `[semantic]` in `pyproject.toml` carries both
`faiss-cpu` and `sentence-transformers`, and I hit this exact
`ModuleNotFoundError: No module named 'faiss'` on a bare `uv sync` while
building the harness. The docstring now says `uv run --extra semantic python`,
names *why* (which class needs which package), and separately documents that
`--render-only` genuinely needs no extras — stdlib + numpy only.

That last sentence is itself a claim, so the harness is now in
`test_example_modules_are_import_light`'s list; if anything heavy ever moves to
module scope, the gate fails.

### P1 — pin and record the checkpoint revisions (accepted, real)

Both defaults are now loaded through
`ModelRef(base=…, kind="hf", revision=…)`, the revision is part of the
`DiskCachedEmbedder` namespace (so a rerun on a new commit cannot be served the
old commit's vectors under the same name), and `measure()` records it in the
row. An unpinned model still runs but logs a warning naming the risk.

**The SHAs are not invented.** They are what this machine's Hugging Face cache
has in `refs/main` for each repo, read after the measuring run:

```
sentence-transformers/all-mpnet-base-v2  e8c3b32edf5434bc2275fc9bab85f82640a19130
sentence-transformers/all-MiniLM-L6-v2   1110a243fdf4706b3f48f1d95db1a4f5529b4d41
```

`refs/main` is by definition what an *unpinned* load resolves to, so this pin
changes nothing about which weights produced the committed rows — it only stops
a future run from resolving elsewhere. The report's "What could still be wrong"
now states plainly that the committed rows were measured *unpinned* and that
the pin is a guarantee for the next run, not a fact re-derived for this one. I
did **not** backfill a `model_revision` field into the already-measured rows;
they predate the field and inventing provenance into a measured artifact is the
failure mode this whole PR exists to avoid.

### P2 — exclude stale metric revisions before rendering (accepted, real)

This one was already written down as a guarantee and not implemented — the
`METRIC_REVISION` docstring said *"every row records its own value so a stale
row can never be silently mixed into a fresh table"* while `render()` consumed
every generation at once.

Kept the revision in `_write_row`'s replacement key on purpose (bumping it must
not destroy the measurements that justified the bump) and added the other half:
`_current_revision_only()` filters at render time, logs exactly which
`(benchmark, model)` cells it dropped and at which revisions, and `main` exits
with a clear error if nothing at the current revision survives. The file keeps
everything; the report shows one revision.

### P2 — validate the crosscheck artifact before rendering (accepted, real)

Crosscheck entries now carry `metric_revision`, and `render()` admits an entry
only if it is **both** at the current revision **and** about a
`(benchmark, model)` cell that is actually in the rows being rendered.
Otherwise Section F is omitted entirely rather than silently attributing a gap
measured somewhere else. This also covers your `--render-only --rows <alt>`
case without adding a flag: an alternate rows file simply fails the
cell-membership test and the section drops out.

The tracked crosscheck JSON is stamped `metric_revision: 1` by hand. That
records the revision the run was measured under — it is not a new measurement.

### New tests

`tests/examples/test_external_reproduction.py` (6 tests). The harness
previously had none, so "suite green" was not evidence it ran at all. Covers
the pins, the stale-revision filter, the crosscheck admission rule, and that
the tracked report is byte-identical to what the tracked rows render to.

Per this repo's habit of shipping gates that cannot observe their own failure,
I checked that last one can: appending a single line to the report makes
`test_render_reproduces_the_tracked_report` fail, and it passes again after a
re-render.

### Verification

`uv run ruff check .` ✅ · `uv run ruff format --check .` ✅ (574 files) ·
`tests/examples/` 19 passed. Full suite before this round: **4805 passed, 2
skipped**, repo coverage 96.61%, contract gate 98.30% vs the 98 floor.

@codex review

fxd24:
## Disposition of the second Codex round — all four accepted (`bf87ce8`)

Worth saying plainly: **round 2 found a bigger problem than round 1.** Three of
these four are fix-induced or fix-adjacent, which is the argument for
re-requesting after every round rather than once.

### P1 — "Prove gold-set inclusion before reporting bounds" (accepted; this is the important one)

This is correct and it undercuts a load-bearing claim, so it gets the most
space. `lower = hits / G_theirs` and
`upper = (hits + (G_theirs − G_ours)) / G_theirs` are bounds **only if
`G_ours ⊆ G_theirs`**. What I actually have is two aggregate match counts and a
provenance argument (our CSVs are the DeepMatcher labelled splits, drawn from
the same original benchmarks after a blocking pass). As you say: equal or
smaller counts do not establish subset membership, and neither paper publishes
a pair list, so **nothing in this study can settle it**.

I did not try to rescue the claim. Section B2 is retitled *"a **conditional**
interval"* and now carries an explicit note: what the condition is, that it is
unverified, that equal counts do not prove it, what breaks if it fails (the
`inside?` column and the 4-of-6 headline both move), and what would convert it
into a real bound (either paper's candidate or gold pair list — the same
missing artifact already named under *What could still be wrong*).

The verdict section now also states which conclusions survive if the
assumption fails: `dblp_scholar`, where the gold *count* is DeepBlocker's own
5,347 at their own 392,400-pair budget, and `dblp_acm`, where the interval is
4 pairs wide and we clear their value at either end. Those two were already the
ones the verdict rested on, so the headline conclusion is unchanged — but it no
longer leans on the 4-of-6 count for its strength.

### P1 — "Omit the fixed verdict from partial renders" (accepted, real)

Your `febrl_person` example is exact, and it was reachable from a test I had
just written. `VERDICT` was fixed prose appended unconditionally, so a
one-benchmark render still asserted the DBLP 98.82%, "4 of 6", and a k=20
crosscheck it did not contain.

It is now emitted only when every canonical `(benchmark, model)` cell is
present. A partial render gets a Section G that says so and names how many
cells are missing, plus a log warning listing them.

### P2 — "Match crosschecks on checkpoint revision" (accepted, real)

Right, and it is a hole I opened in round 1: I added `model_revision` to both
`measure()` and `crosscheck()` and then keyed admission on
`(benchmark, model, metric_revision)` only. Section F's "one set of embeddings"
would be false after a re-measure on a moved pin.

`model_revision` is now part of the admission identity. Note the artifacts that
predate the field read `None` on both sides and therefore still match — which
is correct rather than a loophole: they *were* a single run.

### P2 — "Keep alternate rows from overwriting the canonical report" (accepted, real)

Fixed without adding a flag: the output path is derived from the input. Only
`--rows <the canonical file>` writes the canonical report; anything else writes
`<rows>.md` beside itself and logs that it did.

### Tests

6 → 9. The new ones are the moved-pin crosscheck drop, the partial-render
verdict omission, and — deliberately — its **complement**, asserting that a
full render still *keeps* the verdict. Without that third one the omission test
could pass forever by never firing, which is the failure mode this repo has now
hit enough times to have a rule about.

### Verification

`uv run ruff check .` ✅ · `uv run ruff format --check .` ✅ (574 files) ·
`uv run pytest tests/examples/` **279 passed, 1 deselected**. Report
re-rendered from the committed rows after formatting, so the byte-identity test
is meaningful rather than incidental.

@codex review

fxd24:
## Disposition of the third Codex round — all five accepted (`6c95413`)

### P1 — "Stop treating equal gold counts as identical pair sets" (accepted; my round-2 fix was itself an overclaim)

This is the finding that mattered, and it catches me over-correcting. In round 2
I labelled the B2 interval conditional — correctly — and then immediately
carved out an exemption: *"the two benchmarks the verdict actually rests on are
the ones that do not need it: `dblp_scholar`, where the gold count is
DeepBlocker's own 5,347, and `dblp_acm`, where the interval is 4 pairs wide."*

That does not follow, for exactly the two reasons you give. Recall over **our**
5,347 pairs bounds nothing about recall over a **different** 5,347 pairs; and
`dblp_acm`'s 4-pair-wide interval is computed *under* the nesting assumption, so
its narrowness is not independent evidence for it. I fixed the label and then
smuggled the original claim back in one paragraph later.

Rather than patch that sentence, the verdict is restructured into **two answers
that must not be run together**:

**Unconditional — does the harness compute the literature's quantity?** Yes, and
none of this touches anyone's gold pairs:
- the candidate set is theirs to the pair — `150 x 2,616 = 392,400` against
  DeepBlocker's printed `392.4k`;
- `PQ = hits/(k·|A|)` is their formula, reproducing their own printed PQ to 2 dp
  on five of six UniBlocker rows — and sharply enough that the sixth resolves
  only if `|A| = 533`, i.e. it finds a typo in their Table 2;
- the Section F `score_blocking` crosscheck is langres-internal on both sides
  and reproduces the embedder ladder digit-for-digit.

**Conditional — do we reproduce their recall?** 0.7–1.5 pp everywhere a
comparison exists, and **every single one of those is conditional, on every
benchmark, with no exceptions.** The `fodors_zagat` PQ result moved *up* into
the unconditional list, since it is a statement about the formula rather than
about a gold set.

"What this does and does not license" now separates the same two, and explicitly
refuses "a settled reproduction of anyone's recall".

The practical effect on the deliverable: the thing this study was for —
confidence that the harness measures what the literature measures — is intact
and now rests only on evidence that needs no assumption. The recall agreement is
still reported, just no longer as settled.

### P1 — "Require canonical measurements before emitting the verdict" (accepted, real)

Correct, and a hole in the round-2 fix: I gated on key presence, and your
`--k-max 20` example keeps all 16 keys while destroying the k=150 number the
prose quotes. A cell now counts only if it was searched to the full canonical
depth (`k_max >= min(K_MAX, n_b)`), and the gate additionally requires the
Section F crosscheck the verdict cites to have survived admission.

### P2 — "Preserve canonical crosschecks during trial runs" (accepted, real)

`--crosscheck 5 --benchmarks febrl_person` truncating the tracked artifact is a
data-loss shape this repo has a rule about — an overwritten measurement is not
recoverable from git once committed over. It now **merges**, keyed on
`(benchmark, model, model_revision, k, metric_revision)`: a rerun replaces its
own cell and leaves every other one alone. Chose merge over a derived output
path here because the failure mode is losing evidence, and merge fixes that
without making the operator remember a second flag.

### P2 — "Use the revision-pinned model name in the example" (accepted, real)

Also fix-induced: round 1 added `MODEL_REVISIONS` with an exact-match lookup,
and the docstring's one-model example still passed the bare
`all-mpnet-base-v2` alias — so the documented command took the unpinned branch
the pin was added to remove. Now uses the canonical
`sentence-transformers/all-mpnet-base-v2` and says why the exact name matters.
I did not add alias normalisation; one correct example beats a lookup that
guesses.

### P2 — "Report the actual search cap when PC90 is not reached" (accepted, real)

`>100` was hardcoded while `_k_at` searches to `min(UNIBLOCKER_K_CAP, k_eff)`,
so a `--k-max 20` row claimed a range that was never searched. Now renders the
depth actually searched.

### Tests

9 → 12: the reduced-k cell rejection (the P1 above, asserting the 98.82% claim
disappears), the searched-depth rendering, and crosscheck cell-key uniqueness —
the merge keys on that tuple, so a duplicate would mean silent cell loss.

### Verification

`uv run ruff check .` ✅ · `uv run ruff format --check .` ✅ (574 files) ·
`tests/examples/test_external_reproduction.py` 12 passed. Report re-rendered
from the committed rows *after* formatting, so the byte-identity test is
meaningful. `docs/research/README.md`'s index row re-synced to the new verdict —
it had already gone stale once this PR, which is why it is checked each round.

@codex review

fxd24:
## Disposition of the fourth Codex round — all six accepted (`e0c9e6a`, `b6a85ad`)

Four of these are prose overclaims. Three are mine from round 3; one survived
the round-3 rewrite untouched. Worth stating plainly: **each round has found
real accuracy defects in the summary prose, and the pattern is me tightening one
sentence while an equivalent claim two paragraphs away keeps the old strength.**

### P1 — "Restrict the recall-gap claim to supporting rows" (accepted)

Correct, and it contradicts my own tables. "Our recall sits 0.7–1.5 pp from the
published values wherever a comparison is possible" is true of two rows and
false of four: `fodors_zagat` is 5.4 pp away — the report says so three
paragraphs later — and the product rows are 4–7 pp apart as points, reaching
"consistent" only *through* the conditional interval, which is not the same as
agreeing.

Both the headline and the licensing paragraph now state the spread:

> Two rows land within 0.7–1.5 pp (`dblp_scholar` +0.72, `dblp_acm` +1.48). One
> is 5.4 pp off in our favour and unexplained. The three product benchmarks are
> 4–7 pp apart as raw points and only *overlap* once the gold-set interval is
> applied — which is not the same as agreeing.

### P1 — "Keep Fodors pair identity conditional" (accepted)

This is the equal-count inference I had *just* rejected as a universal caveat,
still sitting two paragraphs below it: "its gold set is *identical* to theirs,
so unlike the product benchmarks the residual cannot be a denominator effect."

Now: the **count** is identical (112), so a *size* difference cannot explain the
residual — and that is as far as it goes, because a different 112 pairs gives a
different recall.

### P1 — "Do not equate the local ablation with the literature blocker" (accepted)

Also right. Section F is entirely langres-internal — one set of *our*
embeddings under two of *our* protocols — and I used it to assert that
langres's 0.81 and the literature's ~0.90 are "the same blocker measured two
ways". It cannot reach their number: that would additionally need the inferred
checkpoint identity, the unverified gold correspondence, and a serialization
that on `amazon_google` demonstrably differs (3 fields to UniBlocker's 4 — my
own Section E says so).

Restated: the closure accounts for the whole gap **between our two protocols**,
and is a *sufficient-looking* explanation for the distance to the literature,
not a demonstrated one.

### P1 — "Validate exact evidence before appending the fixed verdict" (accepted)

The round-3 gate checked presence and "at least one crosscheck survived", which
leaves the two holes you name. Both closed:

- Every figure `VERDICT` states as a measurement of ours is now registered in
  `VERDICT_CLAIMS` as `(benchmark, model, k, field, percent)` and re-checked
  against the rendered rows to 2 dp before the prose is emitted. A re-pin that
  moves a number downgrades the report instead of letting fixed text assert a
  value nobody re-read.
- The **specific** crosschecks the prose cites (`abt_buy`, `amazon_google` at
  k=20) are required, not any survivor.

I stopped short of deriving the prose from the artifacts. The verdict is
argument, not a table — the value is in the reasoning about what is conditional
on what, which does not generate. Validating every number it quotes gets the
safety without pretending the analysis is mechanical.

### P2 — "Isolate caches for unpinned checkpoint revisions" (accepted)

Right, and it was a hole I opened in round 1: every unpinned model shared one
`__unpinned` namespace, so after the Hub's `main` moved a run would be served
the previous checkpoint's vectors while recording no revision. Caching is now
**disabled entirely** for an unpinned model. Re-encoding costs time; a silently
wrong number costs the study.

### P2 — "Preserve measurements across checkpoint revisions" (accepted)

Consistent with the crosscheck merge from round 3 and with the cache namespace —
`model_revision` was identity everywhere except the one place that *deletes*
data. It is now in `_write_row`'s key, so rows are additive per generation, and
`_select_generation` chooses at render time: the current pin, else the single
row present, else **drop the cell with a warning** rather than guess. A dropped
cell then downgrades the verdict, which is the honest outcome.

(Rows predating the field read `model_revision: None` and are accepted as the
fallback — the tracked artifact is exactly that shape, and I am still not
backfilling invented provenance into a measured file.)

### Tests

12 → 18: the claim registry matches the committed rows (or it gates nothing), a
moved measurement drops the verdict, the *named* crosschecks are required, an
ambiguous multi-generation cell is dropped rather than guessed, the current pin
beats a legacy row, and an unpinned model gets no disk cache.

### Verification

`uv run ruff check .` ✅ · `uv run ruff format --check .` ✅ (574 files) ·
`tests/examples/test_external_reproduction.py` 18 passed. Report re-rendered
after formatting. `docs/research/README.md`'s index row carried the same
0.7–1.5 pp overclaim and is corrected in `b6a85ad`.

@codex review

fxd24:
## Disposition of the fifth Codex round — all six accepted (`52e9b4f`)

### P1 — "Do not claim the checks exclude a wrong split" (accepted; the strongest sentence in the doc was the wrongest)

Correct, and this was the single most load-bearing overclaim left. I had written
that a wrong split "could not survive those checks". Walking each one:
`|C| = K·|A|` verifies candidate-set **cardinality**; the PQ reconciliation is
arithmetic on **their** printed numbers; the ladder crosscheck compares two of
**our** implementations on **our** data. A different pair set with the same table
sizes and the same match count sails through all three.

Narrowed to what they actually exclude — a wrong candidate-set convention and a
wrong metric — with an explicit paragraph saying they cannot exclude a wrong pair
set and that nothing in this study can, because no paper publishes one. The
headline claim is correspondingly weaker: *"langres builds the literature's
candidate set and computes the literature's formulas"*, not *"...so our ground
truth must be right"*.

### P1 — "Remove the surviving Fodors pair-identity claim" (accepted)

Right, and embarrassing: the round-3 rewrite left a **duplicate** paragraph and
the survivor was the *pre-correction* one, sitting directly below the corrected
count-only version. Deleted.

Method note, since this is the second time a fix has missed a sibling site: I
grepped `identical` / `strict subset` across the whole file rather than trusting
the flagged line, which is how I found the next item too. Two other sites were
already correct.

### P1 — "Keep the gold-subset statement conditional" (accepted)

The "Reading order" preamble still said *strict subset* — stating as fact the
nesting the rest of the document treats as unverified, and doing it several
sections *before* the assumption is introduced, so a reader meets the conclusion
first. Now: fewer pairs, nesting explicitly conditional, pointer to B2.

### P1 — "Validate every measurement quoted by the fixed verdict" (accepted)

Fair — the registry covered six row values while the prose also hard-codes the
`wdc_computers` figures and the crosscheck deltas. Extended to **ten** row
claims plus a new `VERDICT_CROSSCHECK_CLAIMS` table for the four crosscheck
figures (95.80 / 80.43 / 81.08 / 93.68), all re-checked to 2 dp before the prose
is emitted.

**Where I stopped, deliberately:** I did not derive the verdict from the
artifacts. It is an argument about *what is conditional on what*, and that does
not generate from a table — a generated verdict would be a worse document. Every
number it quotes is now gated; the reasoning stays hand-written. Flagging that
as a judgement call rather than a completed instruction.

### P1 — "Do not equate two unknown checkpoint revisions" (accepted, with the fix in prose rather than code)

The structural point is right: sweep and crosscheck are genuinely separate
invocations, both predate the pin, and `None == None` is not evidence they
loaded the same weights.

I did **not** reject the legacy pair, because that would delete a real result
from the report to express a doubt I can actually resolve. What resolves it is
checkable and now stated in Section F: this machine holds exactly **one** cached
snapshot of `all-mpnet-base-v2` — `e8c3b32edf5434bc2275fc9bab85f82640a19130`,
which is also its `refs/main` — so no other weight set was available to either
run. The note says plainly that this is evidence about the machine rather than a
guarantee from the artifact, and that it applies only to the artifacts committed
with this study; every run from here records its revision and the admission rule
compares it.

If you would rather the code refuse legacy pairs outright, that is a one-line
change and Section F disappears from the committed report — I judged the caveat
more useful than the deletion, but it is a judgement call.

### P2 — "Drop legacy-plus-stale checkpoint groups as ambiguous" (accepted, real logic bug)

Correct, and it violated the policy I wrote in the same docstring: `pinned or
legacy` picked the legacy row when a cell held a legacy row *and* a stale SHA
with neither current. Only the current pin is a positive selection now; the
"single row present" fallback cannot apply to a multi-row group by definition.

### Tests

18 → 20: the crosscheck claim registry matches the committed artifact, and a
legacy-plus-stale cell is dropped rather than falling back.

### Verification

`uv run ruff check .` ✅ · `uv run ruff format --check .` ✅ (574 files) ·
`tests/examples/test_external_reproduction.py` 20 passed. Report re-rendered
after formatting; the extended gate passing on the committed artifacts is itself
confirmation that all fourteen registered figures are real. Checked
`docs/research/README.md` for the same two phrasings this round — clean.

@codex review

fxd24:
## Disposition of the sixth Codex round — all five accepted (`3241025`)

### P2 — "Correct the `Benchmark.load` return position" (accepted; verified against the source, and I had it exactly backwards)

Checked rather than assumed — `src/langres/data/benchmark.py:141`:

```python
def load(self) -> tuple[list[RecordT], list[set[str]], set[frozenset[str]]]:
    """Return ``(corpus, gold_clusters, gold_pairs)`` for the full dataset."""
```

So the closure is derived from the **second** return and the **third** is the raw
positive list — the inverse of what the docstring said, in the one sentence
explaining the distinction the entire study turns on. My own `_load()` unpacks it
correctly (`corpus, clusters, raw_gold`), so the code was right and only the prose
was wrong, which is the version of this bug that survives review longest. Fixed in
both places it appeared (grepped for the phrase rather than fixing the flagged
line).

Filed as P2; I'd have called it P1. It is a false statement about a public data
contract in a document about not making false statements.

### P1 — "Gate all measurements used by the fixed verdict" (accepted — the test was self-referential)

Correct and slightly embarrassing, since I added that test *as* the gate. Iterating
`VERDICT_CLAIMS` can only confirm entries that are already there; it can never
notice one the prose asserts and the registry omits. A check that cannot observe
the failure it exists to catch is the exact pattern this repo has a rule about, and
I wrote a fresh one.

Replaced with an audit from the other direction: pull **every** decimal out of
`VERDICT` by regex and require each to be either a registered measurement or
explicitly listed in `_ACCOUNTED_FOR` with a note saying what it is instead. It
found eight unregistered numbers on the first run, including your `abt_buy` k=13
example.

The registry now also carries the six PC-at-their-k values that drive every printed
interval and therefore the "4 of 6" conclusion. The remainder are annotated:
published values (cross-checked against the reference JSON by a second new test),
derived differences, bound endpoints, and section references.

`_ACCOUNTED_FOR` is an allow-list and allow-lists rot closed rather than open — a
new unexplained number fails loudly rather than slipping through, which is the
direction I want it to fail in.

### P1 — "Narrow the crosscheck validation claim" (accepted)

Right. "Two independent implementations agreeing to four decimals means the delta
is a real protocol difference and not a bug in either" — they are not independent
end to end. They share the corpora, the gold data, the serialization, the embedder
and the FAISS index; only candidate assembly and the recall computation differ.

Now stated precisely: the agreement is evidence about **those steps** — this
script's symmetric-kNN branch assembles the same pairs and scores them the same way
as `score_blocking` — it says **nothing** about the directional branch producing
the other side of the delta, which has no second implementation, and a defect in a
shared input would move both sides together and be invisible here.

### P1 — "Reject unpinned crosschecks outside the legacy artifact" (accepted)

Correct, and it was the loophole in my round-5 answer: I justified the `None`
admission with an argument specific to three committed cells, then left the code
admitting *any* unpinned pair — and Section F would have printed the hard-coded
`all-mpnet-base-v2` cache explanation for an unrelated model. Restricted to
`LEGACY_CROSSCHECK_CELLS`, the three cells actually committed here. Everything else
must carry a revision.

### P2 — "Match crosscheck claims on model identity" (accepted)

Same shape as the round-3 rows fix and I missed it here: the predicate keyed on
`(benchmark, k, field)` while the prose names `all-mpnet-base-v2` explicitly. Model
is now part of both the presence and the value predicates.

### Tests

20 → 22, and one existing test replaced with a non-self-referential version.

### Verification

`uv run ruff check .` ✅ · `uv run ruff format --check .` ✅ (574 files) ·
`tests/examples/test_external_reproduction.py` 22 passed. Report re-rendered after
formatting.

### Where this stands

Six rounds, 32 findings, all accepted and fixed. The prose ones have been the
valuable half throughout — the numbers were never wrong, but the sentences
describing what they support kept being stronger than the evidence, and each round
has found the next one. I have stopped short of deriving the verdict from the
artifacts (flagged in round 4); every number in it is now gated instead.

@codex review

PR #250, opened 2026-07-28
measure: should fit(derive_threshold=True) be the default? (54 cells, held-out, $0)

fxd24:
## Measured: should `fit(derive_threshold=True)` be the default?

#241 landed `fit(..., derive_threshold=True)` — derive the match cut from labels
(Youden's J), then **race it against the incumbent** and keep the winner. It is
off by default, while the front door hard-codes `threshold: float = 0.5` in six
places. So: should the flag default to `True`?

**This PR measures it and changes nothing.** No default moves, no `src/` behavior
changes — following the precedent set two days ago in the same area: *the
measurement is the deliverable; changing a default is a separate PR.*

- Harness: `examples/research/threshold_default_study.py` ($0 — `rapidfuzz` /
  `embedding_cosine`, no paid call)
- Raw data: `examples/research/results/threshold_default_study.json` (54 cells)
- Write-up: `docs/research/20260728_threshold_default.md`

**54 cells** = 9 loadable benchmarks × 2 $0 scorers × seeds 0/1/2. **Every number
is graded on a corpus-disjoint held-out split** — see "the split trap" below for
why it is not `fit`'s own `split=0.3`.

### The answer

| outcome of `fit(derive_threshold=True)` | cells |
|---|---|
| derived cut kept **and** better held-out | **45** |
| race declined → threshold unmoved, Δ exactly 0 | **7** |
| derived cut kept **and** worse held-out | **2** (both `tiny_fixture`, whose held-out corpus has **1** gold pair) |

On the **8 real benchmarks the shipped outcome never loses, at any seed, for
either scorer.** Wins run +0.0011 → **+0.8887** pair-F1. Full per-benchmark,
per-seed table in the write-up (§2); it is never averaged.

**Those are outcomes of *derive then race*, not of the derived cut alone** — the
distinction is the whole case for the race. Scored on its own, the raw Youden
candidate is **45 better / 1 tied / 8 worse**, six of the losses on real
benchmarks (`dblp_scholar`/`rapidfuzz` ×3, `wdc_computers`/`rapidfuzz` ×3). They
show up as ties above only because the race refused them.

**The race earns its seat**: it declined 7 times and was right **7/7** — in 6 the
derived cut really was worse held-out, the 7th was an exact tie. A plain
derive-and-apply would have shipped six real held-out regressions on
`dblp_scholar` and `wdc_computers` as "improvements". It is not infallible the
other way (it kept the 2 losing cuts), which is #241's own documented residual —
"selecting on `train` bounds the risk, it does not remove it" — now observed.

### Did I flip the default? No — and the reason is mechanical, not cautious

`derive_threshold: bool = False → True` makes six documented call shapes raise.
The first three were **run**, not inferred — each is `OK` today:

```
with derive_threshold=True:            today (False):
  ValueError  fit(records)               OK  fit(records)
  ValueError  fit(records, labels=[…])   OK  fit(records, labels=[…])
  ValueError  fit(records, pairs=…,      OK  fit(records, pairs=…, method=Platt())
              method=Platt())
```

Plus three by code read: `fit(pairs=…)` would start requiring `[trained]`
(sklearn), start raising for decider matchers, and raise on an `_ops` chain with
no `ThresholdSelect`. The repo's own suite would go red on the first test.

So the question splits in two with different answers: *"when asked, should `fit`
derive?"* — **yes, decisively**; *"should `fit` derive without being asked?"* —
an API change, not a measurement consequence. §3.3 lays out three options
(leave opt-in / tri-state `bool | None` / fix the constant) with each one's real
cost and reversal, and recommends **leave opt-in now, fix the constant next,
tri-state only as a deliberate 0.4 API change**.

### Two findings that outrank the flag

**1. The split trap** — `align_pairs(split=0.3)` holds out **zero pairs in 26 of
27 (benchmark, seed) rows** on a dense label set. Its entity-disjoint split
assigns whole union-find components, and a k-NN candidate graph is essentially
one component. Not scale-dependent: identical at 26 labeled pairs and at 1.7 M.
`split=` never affects the derived cut (selection runs on `train`), only what
`FitReport` reports — so `FitReport.metrics` / `held_out_f1` are untrustworthy
there and **nothing warns**. `docs/EXPERIMENTS.md` now says so; a warning in
`align_pairs` when the achieved fraction misses the requested one is worth its
own PR.

**2. `0.5` is wrong per *score family*, and `derive_threshold` cannot fix the
out-of-the-box case at all** (it only ever helps a user who already has labels):

| method | benchmarks | min derived t | median | max derived t | spread |
|---|---|---|---|---|---|
| embedding_cosine | 9 | 0.8090 | 0.8634 | 0.9451 | **0.1361** |
| rapidfuzz | 9 | 0.1744 | 0.5590 | 0.6952 | **0.5209** |

For cosine a per-family default near 0.86 would land close to the derived cut on
every dataset — **with no labels**. For rapidfuzz the derived cut varies 4×, so a
shared constant is a far less promising fix (this study measures F1 only at `0.5`
and at each dataset's own derived cut, so it cannot rule a shared constant *out*).
One number cannot serve both, which is exactly what `MethodSpec.default_threshold`
already exists to express ("score scales differ per family" — E12) and exactly
what the six architectures bypass. *(Stated as an inference from the spread: a
fixed-constant arm was not measured. §5.2 flags it.)*

### The six `0.5`s — proposal only (§5)

It is **seventeen** sites in `src/`, not six, all meaning the same decision cut —
spelled three different ways, so no single grep finds them (§5.1 lists the two
hits that are a *different* cut and the two sites the obvious pattern misses).
`architectures/reranker.py` is the instructive exception — `threshold: float`
with no default at all. Proposal: one named `DEFAULT_MATCH_THRESHOLD` beside
`predicted_match` so the caveat is stated once where the value lives; do **not**
sweep in the report/curation cuts (same concept, different owner); and treat the
constant as the floor, not the fix — routing the architectures through
`MethodSpec.default_threshold` is the better change, in its own PR with its own
measurement. Nothing is changed here.

### Scope / caveats

- **Published as an exploratory run**, in the sense `docs/REPRODUCIBILITY.md`
  defines: the artifact records cell metrics, not a provenance envelope, and the
  blocker uses the mutable `all-MiniLM-L6-v2` alias with no pinned revision. The
  conclusion rests on a direction that holds across all 8 real benchmarks and on
  a two-cell exception with an identified cause, not on a decimal being
  reproducible. Not relabelled clean after the fact (§1).
- **`weighted_average` was not raced**, and it is what `FuzzyString` actually
  uses. `rapidfuzz` is a close stand-in (same fields, same weights, same
  `score_type="heuristic"`) but it is a stand-in.
- No paid method, so nothing here covers `VectorLLMCascade` / `RetrieveLLM`.
- Every registered benchmark is *linkage*; nothing here speaks to dedup.

### Checks

`ruff format --check .`, `ruff check .`, `mypy src`, and the fast suite all green
locally. No `src/` file is touched. `--render <json>` reprints every table from
the tracked artifact, so the write-up's tables are regenerated rather than
transcribed.

The harness enforces its own two claims rather than only stating them, and both
guards were verified by watching them fire: `--methods` is restricted to
`ZERO_SPEND_METHODS` (a paid scorer would otherwise reach `make_resolver_factory`
with `llm_client=None` and build a *billed* client from the environment), and the
canonical artifact is published only by a run that measured the complete matrix —
decided by checking the three inputs that define it (benchmarks × methods × seeds)
rather than by blacklisting the flags that narrow it, since `--methods rapidfuzz`
and `--seeds 0` both sailed past the earlier `--fast`/`--only` guard.


fxd24:
@codex review

fxd24:
@codex review

Round 2 fixes are in `2a7e1f5`. Notes on the round-1 findings:

- **Partial sweeps replacing the canonical artifact** — fixed in `0112606`, and the finding was re-posted verbatim against the fixed code (its AGENTS.md link still points at `7f45ac5`). `run_benchmark` now yields, the canonical write happens once after the failure check, and `main()` exits non-zero on failure. Verified by interrupting a real sweep at 18 cells: the 54-cell artifact was left byte-intact and the cells survived in the `.partial` scratch file.
- **Paid scorers in a $0 CLI** — fixed; `--methods` is restricted to `ZERO_SPEND_METHODS`.
- **Narrowed sweeps bypassing the guard** — fixed, and generalised: the guard now checks the three inputs that define the matrix rather than enumerating narrowing flags, so it cannot rot open again. Both `--methods rapidfuzz` and `--seeds 0` are now refused.
- **Site inventory** — you were right on all three counts (grep returns 17, table summed to 16, prose said 14). Rebuilt from the code, with the two non-match-cut hits excluded by name and the two sites the pattern cannot see added.
- **Even-set median** — fixed via `statistics.median`; the published 9-benchmark figures are unchanged.
- **Provenance** — taken as the "publish as exploratory" option, since `docs/REPRODUCIBILITY.md` also says not to relabel a dirty run clean after the fact, which retro-fitting an envelope onto an already-measured artifact would amount to.

fxd24:
@codex review

Round 3 in `47a8822`. Both of your new findings were correct and both were fix-induced — they existed only because of the round-2 changes.

- **Explicit canonical `--out`** — right, and a real hole: the guard tested `args.out is None`, so it protected only the default path. Now keyed on the resolved destination; `--methods rapidfuzz --out <canonical>` and the absolute-path spelling both exit 2 (verified).
- **Ignored checkpoint** — right, and wrong in this repo's most expensive way. Gitignoring `*.json.partial` made an interrupted sweep's only copy invisible to `git status --untracked-files=all`, which is exactly the check the worktree workflow relies on to catch unsaved work before teardown. Entry reverted; `.gitignore` is now untouched by this PR.

Three findings were re-posted verbatim and are **stale, not unfixed** — flagging rather than silently ignoring:

- *Partial sweeps replacing the canonical artifact* — fixed in `0112606`; the comment's own AGENTS.md link still points at `7f45ac5`. `run_benchmark` yields, the canonical write happens once after the failure check, and `main()` exits non-zero. Verified by interrupting a real sweep at 18/54 cells: the 54-cell artifact was left byte-intact.
- *Reject paid scorers* — fixed in `2a7e1f5`. The comment anchors on the argparse declaration and doesn't see the validation below it; `--methods llm_judge` exits 2.
- *Provenance* — taken as the "publish as exploratory" branch you offered, since `docs/REPRODUCIBILITY.md` also forbids relabelling a dirty run clean after the fact, which retro-fitting an envelope onto already-measured cells would be.

fxd24:
@codex review

Round 4 in `97f7b42`. Your new finding was right again, and again fix-induced by the previous round:

- **Existing checkpoint clobbered on re-run** — correct. Round 3's comment argued the checkpoint must stay visible so an interrupted sweep forces a commit/copy/discard decision, but the code reused the path with an empty `results`, so the first write truncated it. A comment asserting a guarantee the code never enforced. Now an existing `<out>.partial` refuses the run with instructions; verified the run exits 2 and the checkpoint survives intact. No `--resume` flag — refusing is what makes the decision actually happen, and resume isn't a demonstrated need.

The other three in that round were byte-identical re-posts of findings already fixed and verified (partial-sweep publication, fixed `0112606`; paid scorers, fixed `2a7e1f5`; provenance, answered via the exploratory branch). Their anchors point at the argparse declarations and the `except` handler rather than at the guards added below them, and one still links AGENTS.md at `7f45ac5`. Flagging rather than silently dropping them — if any is genuinely unfixed at `97f7b42`, the specific line that still permits it would help.

fxd24:
@codex review

Round 5 in `1cd15d9`: the checkpoint-write finding was right and is fixed. The per-cell write sat inside the benchmark try/except, so a persistence failure was misclassified as a benchmark failure and the closing message still claimed "N cell(s) kept at <scratch>" over cells that existed only in memory. Checkpoint failures now raise `CheckpointError`, which the handler re-raises instead of swallowing. Verified by forcing ENOSPC on every checkpoint write: the run aborts on the first cell rather than continuing.

Four findings in that round were byte-identical re-posts of already-fixed issues (partial-sweep publication → `0112606`; paid scorers → `2a7e1f5`; interrupted-checkpoint clobbering → `97f7b42`, the very commit under review; provenance → answered via the exploratory branch). Each anchors on a declaration above the guard that fixes it. If any is genuinely still open at `1cd15d9`, please point at the line that permits it.

fxd24:
@codex review

Round 6 in `e453b2f`, plus two self-caught follow-ups in `9f2ebea` / `9c2a7e1`. Your three analysis findings were all correct and all mattered more than the harness ones:

- **45/7/2 are race outcomes, not the derived cut's** — right, and I've made both framings explicit. Verified against the artifact: the raw Youden candidate alone is **45 better / 1 tied / 8 worse**, six of the losses on real benchmarks (`dblp_scholar`/`rapidfuzz` ×3, `wdc_computers`/`rapidfuzz` ×3) that appear as ties only because the race refused them. This strengthens the conclusion — it is the measured case for keeping the race — but the old phrasing let the race's save read as the candidate's win.
- **`split=` does change the derivation set** — right. `resolver.py:1276` aligns before selecting, so pairs moved into `valid` leave `train`. Selection never *reads* `valid` (that part stands), but "only changes what `fit` reports" is true only when `valid` is empty — 26 of 27 rows here, which is why it held for every number in this study and why I wrongly generalised it. Corrected in the write-up, the harness docstring and `docs/EXPERIMENTS.md`.
- **Seed stability overstated** — right, and worse than flagged. Measured worst case is `walmart_amazon`/`rapidfuzz` at 0.0384 (~8× the claimed ±0.005). Replaced with a measured per-family table.

Two further errors I caught myself while verifying those fixes rather than assuming them: the replacement text named the wrong tightest cell (`febrl_person`/`rapidfuzz` is 0.0000, not `dblp_acm`), and then named `wdc_computers` (0.0097) among cells exceeding 0.01. Also put the four TL;DR win figures on one seed basis. I then checked the 54-row table **mechanically** against the artifact — every data row matches; the errors were confined to hand-written prose, which is the argument for `--render` generating the tables.

Also fixed: `write_results` is now atomic (write beside + `os.replace`), since a truncating write interrupted part-way destroys the previous checkpoint and leaves a file that looks fine until parsed.

Not changed: the scratch-under-`tmp/` visibility point — `tmp/` is this repo's declared-ephemeral convention, and the checkpoint beside the canonical artifact (the 40-minute sweep, the case that matters) is visible to `git status -uall`.

fxd24:
@codex review

Round 7 in `8fc8689`. All five analysis findings were correct, and three were overclaims in the direction that flattered the conclusion — the most valuable round so far:

- **"Six call shapes raise" was three.** You were right that some already fail today. Verified by running it: `fit(pairs=...)` on a non-supervised matcher **raises today** and is **OK under the flip** — `_fit_from_pairs` refuses `pairs=` when the matcher isn't a `SupervisedFitMixin` and the flag is false (`resolver.py:1267`). The flip *legalizes* that path. The three executed shapes stand (the first is the plain `fit(records)` no-op, which the repo's own suite calls), so the argument survives, but the count is three.
- **tiny_fixture F1 isn't binary.** Disproved by my own artifact — `tiny_fixture`/`embedding_cosine` is 0.5000 on that same single gold pair. Corrected to: recall is binary, sample size is what makes the cells uninformative.
- **Two inventory sites aren't effective cuts.** `RapidfuzzMatcher.threshold` is never read by `forward()` (its own docstring says so) and `EmbeddingScoreMatcher.threshold` only picks a `decision_step` label. The table now marks effectiveness per row.
- **The LLM-recipe recommendation was inert.** `Parse` emits `decision=True/False` and `predicted_match` gives decision precedence, so 0.5→0.7 on `RetrieveLLM`/`RetrieveRerankLLM` changes nothing. I had the topology right and the conclusion wrong; the real question is why a decider pipeline exposes a threshold at all.
- **"No rapidfuzz constant works" isn't measured.** Right — F1 was measured only at 0.5 and at each dataset's own derived cut, and Youden's J is flat across a separating gap, so far-apart argmaxes don't preclude a shared value. Softened to what the spread supports, with the missing arm named.

Also fixed: the interrupted-checkpoint guard now covers the `.writing` staging file.

Unchanged and why: three findings re-posted from earlier rounds are already fixed (partial-sweep publication `0112606`, paid scorers `2a7e1f5`, checkpoint clobbering `97f7b42`); and the scratch-under-`tmp/` visibility point — `tmp/` is this repo's declared-ephemeral convention, while the checkpoint beside the canonical artifact (the 40-minute sweep) is visible to `git status -uall`.

fxd24:
@codex review

Round 8 in `2c002e0`. Your `--render` finding was the root cause of this entire review thread, and I've treated it that way:

**`--render` promised "every table" and emitted three of six.** The outcome counts, the seed-stability summary and the decline audit were typed into the write-up by hand — and **every numeric error found across rounds 6–8 was in one of those three; not one was in a generated table.** All three are now generated (`to_outcome_markdown` / `to_seed_stability_markdown` / `to_decline_markdown`), and generating them independently reproduced the corrected numbers exactly: race 45/7/2, raw candidate 45/1/8, 6 of the 8 raw losses caught by the race, win span +0.0011..+0.8887, cosine spread 0.0000–0.0067 (0 of 9 cells >0.01), rapidfuzz 0.0000–0.0384 (3 of 9), 6 correct declines + 1 exact tie. That is the check the hand-written versions never had. The outcome table prints **both framings side by side**, since conflating them was the most consequential error in the earlier draft.

Also fixed, all duplicates of corrections already made elsewhere:
- `docs/research/README.md` still advertised "six raising paths" → three.
- The per-cell table's `tiny_fixture` note still called its F1 "a coin flip between 0.0 and 1.0" → recall is binary, F1 is not (the cosine rows read 0.5000).
- `to_spread_markdown`'s docstring still said a wide spread makes the fixed-constant objection "die" → it weakens it; ruling it out needs a fixed-grid sweep this harness does not run.
- The interrupted-checkpoint guard now also covers the `.writing` staging file.

Still unchanged, with reasons: the partial-sweep, paid-scorer and checkpoint-clobbering findings are re-posts of issues fixed in `0112606` / `2a7e1f5` / `97f7b42` and verified by running them; and the scratch-under-`tmp/` point — `tmp/` is this repo's declared-ephemeral convention, while the checkpoint beside the canonical artifact (the 40-minute sweep) is visible to `git status -uall`.

PR #252, opened 2026-07-28
research(embeddings): do instruction prompts help ER blocking? per model, per benchmark

fxd24:
## What this is

A measurement, with its evidence. **Zero files under `src/`** — a harness, its tests, and the results. No wiring, no default change, no API change. Whoever ships the prompt wiring later gets a committed results table to justify it.

Answers: *do better instructions make instruction-following embedders measurably better at blocking, and can we pick good ones rather than guessing?*

## The answer

**Yes — but only the checkpoint's own *retrieval* prompt, and "documented" is not sufficient.**

| what you do | outcome | evidence (Δ candidate recall @k=20, `wdc_computers`) |
|---|---|---|
| The checkpoint's **documented retrieval prompt** | **Reliable win** — significantly positive on ≥1 benchmark for all 4 instruction-trained models, never significantly negative | Gemma **+0.0394**, bge **+0.1224**, Qwen3 **+0.0646** (all exclude 0). e5 is the weakest case: its wdc **+0.0152** closes on zero, so its evidence is `abt_buy` **+0.0069** [+0.0010, +0.0128] |
| **Your own task text** in its template shape | **Coin flip**, model-specific; never beat the model's own default on candidate recall | Qwen3 **+0.0388** (helped) vs Gemma **−0.1589** (hurt) |
| A **raw English sentence** outside any template | **Reliably harmful** | e5 **−0.1038**, Qwen3 **−0.0907**, bge **−0.0775**, MiniLM **−0.0632** |

**The trap:** EmbeddingGemma's `clustering` template is every bit as official as its retrieval one and costs **−0.3183** recall. Anyone skimming for "use the official prompt" takes away the wrong instruction.

Two model-card predictions were tested and **one failed**:
- **e5's card is wrong for this task.** It says use `query: ` on both sides for *"symmetric tasks such as semantic similarity"*. ER blocking looks exactly like that and does not behave like it: symmetric is **−0.0145** on wdc while the asymmetric `query:`/`passage:` recipe is **+0.0152**, and those two intervals do not overlap. (The asymmetric arm's own margin over *no prompt* on wdc is a boundary case — `[+0.0000, +0.0310]` — so what wdc establishes is asymmetric **>** symmetric, not asymmetric **>** nothing; the latter rests on `abt_buy`.)
- **Qwen3's card is right.** It claims 1–5%; we measured +0.4% to +6.5%. But its advice to *"create tailored instructions specific to your tasks"* lost to its own generic **web-search** default (+0.0388 vs +0.0646).

## Official formats, from primary sources

Every prompt string was read from the checkpoint as published — its own `config_sentence_transformers.json` or model card in the local HF snapshot — never inferred from another model's convention. They have nothing in common; guessing one from another measures the wrong thing. Full citations with snapshot hashes and line numbers in §1 of the findings doc.

`e5` → `query: `/`passage: ` · `bge` → a query sentence and **explicitly no document prefix** · `Gemma` → `task: … | query: ` / `title: … | text: ` · `Qwen3` → `Instruct: …\nQuery:` with a **literally empty** document prompt.

## Proof the prompts reached the encoder

A prompt that never reaches the encoder produces identical numbers, which reads as *"instructions don't help"* — this repo shipped exactly that bug once (#239/#242). So the harness **proves** it and **aborts** rather than recording a flat row. Three independent signals per arm: corpus-vector shift (index-build path), query-vector shift (search path — the seam that used to discard it), and doc·query cosine (must be 1.0 for symmetric recipes, <1 for asymmetric — catches a prompt that reached one path but not the other). Plus candidate-pair Jaccard, proving changed vectors changed the *retrieved neighbours*.

**The guard has been seen to fail**, which is the only reason to trust it: it aborted the first Qwen3 run at `cosine=1.000126`. A cosine *above* 1 is impossible for distinct unit vectors, so that was float16 round-off. The tolerance is now bounded on both sides by measured values — ~5× above worst observed round-off, ~11× below the smallest genuine divergence — and `tests/examples/test_prompt_axis.py` exercises every failure branch **and pins that margin** so it cannot be loosened into uselessness.

## Method

- **Candidate recall at fixed `k`**, never F1 at a threshold — a threshold would confound the retrieval effect with the cut.
- CIs are paired bootstraps resampled **by gold cluster**, never by dependent pair rows.
- 5 models × 4 benchmarks × up to 7 arms × 4 `k` values = **400 rows**, all committed.
- **Nothing is averaged across models** — the whole question was whether the effect is model-specific, and it is.
- **Control:** `all-MiniLM-L6-v2` (not instruction-trained) gained nothing anywhere; `er_symmetric` significantly hurt it on 2 of 3. The null result the design predicted.

## Two corrections to rules we had been repeating

1. **"Always drive both halves" is wrong as stated.** For bge and Qwen3 the documented recipe is **query-side only, document side deliberately bare** — and for bge that one-sided recipe is the largest gain in the sweep. The rule is *drive the sides the checkpoint was trained on*.
2. Consequently I checked whether **#242's half-driven warning false-fires on those correct configurations**. Settled by construction, not from the changelog: bge, Qwen3 and Gemma documented shapes are all **silent**, and the actual trap still **WARNS**. No false positive. Already covered by `test_silent_when_only_the_query_side_is_driven`, so no new test was added.

## Also worth keeping

`fodors_zagat` is saturated (every model hits 1.0000 with no prompt) and would normally be dropped as signal-free. It earned its place: of 25 arms, exactly one moved it — Gemma's `clustering` template, 1.0000 → 0.9375 — the same arm that was catastrophic elsewhere. **Saturated benchmarks are useful as harm detectors**, and their silence is cheap.

## Not measured

`walmart_amazon` (24,628 records — the scale test) was cut for runtime; the harness supports it with one flag. `gte`, `nomic`, `mxbai` and the larger Qwen3 tiers are not in the local cache. `nomic-embed-text-v1.5` is the notable gap — it is prefix-based (`search_query: `/`search_document: `) and OSI-licensed.

## Licence note

`google/embeddinggemma-300m` is **gated and `gemma`-licensed — not OSI-approved, use-restricted**. It is labelled as such in every table and is never a default recommendation, despite having the cleanest result. No token is committed or printed.

## Reproducibility, and what re-running it caught

The seeded bootstrap was **not reproducible**: gold clusters arrive as `set`s, so Python's per-process string-hash randomisation changed the order of the `cluster_ids` list being resampled — two runs with `seed=0` gave different intervals. Fixed by sorting clusters before resampling; two runs are now byte-identical apart from timings.

The whole sweep was then re-run deterministically. **All 400 recall values and every point estimate were unchanged**, so the fix cost nothing — but every *interval* moved slightly, and re-deriving the doc from the new rows turned up four claims that did not survive:

1. **e5's wdc win became a boundary case** — `[+0.0009, +0.0308]` → `[+0.0000, +0.0310]`, which no longer excludes zero. This is the only significance verdict in the sweep that moved.
2. **"the only arm positive and clear of zero on all three unsaturated benchmarks" was false** — and contradicted the PR's own Qwen3 table, where `er_in_official_template` also clears zero on 3/3. Gemma's pair is the only **documented** arm that does.
3. **"Gemma's clustering template has the single largest shift in the sweep" was false** — `0.3625` is the largest *documented* shift; `0.3825` (our ER sentence on the control) is larger.
4. **"never beat the model's own default" needed scoping to candidate recall** — on `amazon_google` our text's macro Δ clears zero where the default's does not, while its actual recall is *lower*.

Every number in the findings doc is now re-derived from the committed rows programmatically, resolved by `(model, benchmark, arm)` table context and asserted against each row's point estimate — not transcribed. Worth noting how the first check missed this: it compared **verdicts and point deltas**, both of which were unchanged, so it reported "1 flip" while 43 intervals were stale. Checking the summary rather than the claim is the same failure mode this repo keeps hitting.

## Round 2 of review: what a second pass caught

A second automated review found six more issues, including **a defect in round 1's own fix** and one that invalidated a headline claim:

1. **The OOM fix from round 1 did not work.** `del embedder, index, doc_vectors` released nothing — `blocker` is bound in the innermost loop, outlives it, and holds `index` → embedder → model. I had written a comment asserting "exactly one checkpoint resident" and never verified it; the comment became the evidence.
2. **The weights were never pinned, and two citations were wrong.** Every model loaded by mutable repo name. This machine has **two** Qwen3 snapshots and **three** MiniLM snapshots cached, and for both models `refs/main` resolves to a *different commit than this study cited* (`c54f2e6e…`→`97b0c614…`, `c9745ed1…`→`1110a243…`). Verified rather than assumed that the substance survives: the two Qwen3 snapshots ship byte-identical prompt configs and the resolved MiniLM revision still registers no prompts. Hashes corrected; the harness now pins each revision and **refuses to run against another**.
3. **"The *identical* `er_in_official_template` arm"** was **false** — Gemma drives both sides, Qwen3 query-only — so that gap confounded model, wording and placement, and could not support the model-specificity claim it was cited for. Rebuilt on `er_symmetric`, the one genuinely cross-model-identical arm: spread **0.1041** across five checkpoints on `wdc_computers`.
4. **The documented command did not reproduce the study** — `walmart_amazon` sat in the default sweep unmeasured, so a default `--resume` run would add a fifth benchmark and rewrite the report.
5. **The cache canary could pass vacuously** (on a miss it writes then compares to itself — a check that cannot fail). Fixed structurally: the revision is now part of the cache namespace, so no partition can outlive its weights.
6. **A crashed rerun could look complete** to `--resume`, because fresh arms merged into stale ones left every `(arm, k)` key present.

Two independent checks that the guards observe what they claim:

- **The revision guard was made to fail** against a wrong pin, then pass against the right one.
- **Cold-cache reproduction**: under the new revision-keyed namespace, an end-to-end run of the control model on `fodors_zagat` reproduced the committed rows **exactly** — 3 rows × 8 measured fields, **0 mismatches**. The six fixes changed no measurement.

## Round 3 of review

Seven more, including **data destruction created by round 2's own fix** and a second conclusion-level correction:

1. **A typo in `--arms` silently deleted committed rows and exited 0.** Round 2 added `_drop_cell` *before* evaluation so a crashed rerun would look partial — which made an unknown arm name destructive: it matched nothing, cleared every requested cell, evaluated nothing, and reported success. Unknown arms are now rejected before anything touches `--rows`. The test seeds the real committed rows and compares byte-for-byte, and I disabled both guards to watch it fail first. *Lesson: a fix that adds an early write extends the blast radius of every validation gap upstream of it.*
2. **The revision pin could detect drift but not survive it** — comparing against `main` meant that once upstream moved, the pinned weights could sit in the cache while the run aborted. Now the pinned snapshot directory is loaded directly.
3. **`--resume` ignored experiment identity.** Rows now carry `revision` + a fingerprint of the prompt strings; a row without provenance counts as *unknown*, never as a match. The committed rows are therefore recomputed rather than back-filled — I could have inferred their provenance, but asserting one that was never recorded is exactly the error this PR keeps finding.
4. **"Reliably harmful" was selective.** The evidence table listed five negatives and omitted that EmbeddingGemma is *significantly positive* under the same arm on `amazon_google` (+0.0118 [+0.0046, +0.0196]). The rows say **4 of 5 checkpoints**, and the doc now says that.
5. The generated report still carried both superseded snapshot hashes.
6. Gemma's own prompts contain `|`, and backticks don't protect a pipe from GFM — those table rows rendered misaligned.
7. A `--benchmarks` run wrote rows that no headline table displayed.

## Verification

`ruff` clean · `ruff format` clean · `mypy src/` clean (215 files) · fast suite **4720 passed, 2 skipped** · repo coverage **96.30%** (floor 90) · harness guard tests **14 passed**.

Every guard in this PR has been **made to fail before being trusted** — the encoder-reach guard, the cache canary, the revision pin, and the arms guard. A check never seen to fail is a hypothesis, not a safety net.

**No `src/` changes in this PR** — it adds a research harness, its tests, and the findings. The revision pin is resolved through `huggingface_hub` inside the harness, deliberately not by adding a `revision` field to `SentenceTransformerEmbedderConfig`, which would be a library contract change riding in on a research PR.


fxd24:
Thanks — all six were real. Dispositions below; the review was against `eba0c33`, fixes are in `12b9189` + `81e1de6`.

### P1 — Reject unchanged embeddings within the roundoff tolerance ✅ fixed
Correct, and this was the worst of the six: it was a **hole in the guard whose entire purpose is to catch an ignored prompt**. `shift == 0.0` never fires on float16 — the committed Qwen3 rows carry residuals like `−0.0002207` — so an ignored prompt would have sailed through and published a flat row as evidence the prompt was applied. Replaced every identity check with `np.array_equal` on the vectors, which has no such hole regardless of what the cosine rounds to. Pinned by `test_float16_residual_does_not_hide_an_ignored_prompt`, which asserts the guard still raises at `cosine=1.000126`.

The cosine tolerance now survives in exactly one place — the symmetric cross-path check, where the two sides are encoded through *different* code paths (`prompt_name=` at index build vs `prompt=` at search) and cannot be compared bitwise. It is bounded on both sides by measured values (~5× above worst observed round-off, ~11× below the smallest genuine divergence) and `test_tolerance_still_rejects_the_smallest_real_divergence_measured` fails if anyone loosens it past the real signal.

### P1 — Bind persistent caches to an immutable checkpoint ✅ fixed
Agreed, and worth stating why it mattered more than a normal staleness bug: **every prompt-reach guard compares arms against each other**, so uniformly stale vectors look perfectly self-consistent and no guard can see them. Added `_assert_cache_matches_checkpoint`, which re-encodes a canary through the live checkpoint and refuses to proceed on disagreement — a comparison against ground truth rather than against a sibling arm. Covered by `test_a_cache_disagreeing_with_the_checkpoint_is_refused`.

### P1 — Report deltas using the displayed recall estimator ✅ fixed (as documentation, not as a code change)
Correct: the Δ is macro-over-records while `candidate_recall` is micro-over-pairs, and they genuinely diverge — bge/wdc is `+0.1098` micro vs `+0.1224` macro. I kept the macro estimator rather than switching, because a paired bootstrap needs a per-entity score to resample by cluster and a corpus-wide micro ratio has no per-entity decomposition. So the fix is honesty about which quantity carries the interval: the field is renamed `delta_per_record_recall`, and the divergence is called out in the rows, the generated report header, and a standing note at the top of the findings doc.

### P2 — Release prompt-specific models before loading the next one ✅ fixed
Real: Gemma has six distinct document prompts, so the old shape retained ~7.2 GB of checkpoints for one benchmark. Restructured the loop to group arms by document prompt, keeping exactly one model resident and `del`-ing it between groups. `test_document_prompt_order_puts_the_bare_group_first` pins the ordering the restructure depends on (the `none` group must run first, since it is the baseline every other arm is compared against).

### P2 — Replace the unsupported cross-model effect example ✅ fixed
You were right that neither number is in the rows. Worth recording the cause: they came from an **earlier document of mine** (`embedder_ladder.md`) that I re-used as fact instead of re-deriving. Replaced with the real maximum spread, taken from the rows: Gemma `−0.1589` vs Qwen3 `+0.0388` on `wdc_computers`/`er_in_official_template` — spread `0.1977`, one benchmark, one prompt, two models.

### P2 — Correct the claimed vector-shift ranges ✅ fixed, and the conclusion was withdrawn
Recomputed: `er_symmetric` spans `0.0511`–`0.3825` and the documented prompted sides span `0.0111`–`0.3625`. They **overlap almost entirely**, which does not just correct the numbers — it destroys the mechanism those numbers were offered as evidence for ("an untrained prefix displaces vectors so far it crowds out record-specific signal"). That paragraph now states the displacement magnitude carries no information about the recall ordering, and labels the mechanism an open question rather than a finding.

---

### Also found while fixing these, not from the review

Re-running the sweep to validate the fixes exposed that **the seeded bootstrap was not reproducible**: gold clusters arrive as `set`s, so per-process string-hash randomisation changed the order of the `cluster_ids` list being resampled, and two runs at `seed=0` disagreed. Fixed by sorting clusters before resampling; runs are now byte-identical apart from timings.

The deterministic re-run left **all 400 recall values and every point estimate unchanged**, but moved every interval slightly — and re-deriving the document from the new rows caught four further claims:

1. e5's `official_asymmetric` on `wdc_computers` became a **boundary case**, `[+0.0000, +0.0310]` — the only significance verdict in the sweep that moved.
2. *"the only arm positive and clear of zero on all three unsaturated benchmarks"* was **false**, and contradicted the document's own Qwen3 table (`er_in_official_template` also clears zero on 3/3). Gemma's pair is the only **documented** arm that does.
3. *"the single largest shift in the sweep"* was **false** — `0.3625` is the largest *documented* shift; `0.3825` is larger.
4. *"never beat the model's own default"* needed scoping to candidate recall.

Every number in the findings doc is now re-derived from the committed rows programmatically — resolved by `(model, benchmark, arm)` table context and asserted against each row's point estimate — rather than transcribed.

One note on how these were nearly missed, since it is the same failure mode as the P2 above: my first post-fix check compared **verdicts and point deltas**, both unchanged, and reported "1 flip" — while 43 intervals were stale. It checked a summary of the claim instead of the claim.

fxd24:
@codex review

fxd24:
Round 2: all six were real, and one of them was a defect in round 1's own fix. Fixed in `b350e4e` + `60c070f`.

### P2 — Release the blocker before loading the next prompt group ✅ fixed
**This is the important one: round 1's OOM fix did not work.** You are right that `del embedder, index, doc_vectors` released nothing — `blocker` is bound in the innermost `for k` loop, outlives it (Python loop variables leak), and holds `index` → cached embedder → loaded model. So the next group loaded a second checkpoint while the previous one was still reachable, which is precisely the OOM the grouping was introduced to prevent. Added `blocker` to the `del`.

Worth naming the pattern: I wrote a fix, described it in a comment as "exactly one checkpoint resident", and never verified the claim. The comment became the evidence. The code now says why `blocker` must be in that list, so deleting it again reads as a regression.

### P2 — Pin the checkpoint revision used by each model ✅ fixed — and it was not hypothetical
Checked before fixing, and the check found a live problem. This machine's HF cache holds **two** `Qwen3-Embedding-0.6B` snapshots and **three** `all-MiniLM-L6-v2` snapshots, and for *both* models `refs/main` — what an unpinned `SentenceTransformer("org/name")` actually loads — resolves to a **different commit than this study cited**:

| model | cited | actually resolves |
|---|---|---|
| `Qwen/Qwen3-Embedding-0.6B` | `c54f2e6e…` | **`97b0c614…`** |
| `sentence-transformers/all-MiniLM-L6-v2` | `c9745ed1…` | **`1110a243…`** |

I verified the substance rather than assuming it survived: the two Qwen3 snapshots ship **byte-identical** `config_sentence_transformers.json` prompts, and the resolved MiniLM revision still registers no prompts — so every documented format and every measured number stands. But nothing in the harness had established that, and the citations were wrong. Both hashes corrected in the harness and the findings doc; `ModelSpec.revision` now pins all five, and `_resolve_checkpoint_revision` **refuses to run** against anything else rather than silently measuring other weights.

### P1 — Validate every cached prompt partition ✅ fixed structurally
Agreed, and the sharper version of your point is that the canary could pass **vacuously**: on a cache miss `cached.encode` delegates to `base` and stores, so `served == fresh` by construction and the check carries zero information — a check that cannot fail. Rather than patch the canary per-partition, I removed the hazard: the revision is now part of the cache namespace, so a partition cannot outlive the weights that wrote it, for *any* prompt partition. The canary stays as a backstop for same-revision runtime changes.

### P1 — Invalidate old rows before rerunning a cell ✅ fixed
Correct. Added `_drop_cell`, called before the first arm is flushed, so a rerun that dies part-way leaves a cell that *looks* partial — which is what `--resume` needs to see. Previously the fresh arm merged into the stale ones, every `(arm, k)` key was present, and `--resume` reported a two-run cell as complete.

### P2 — Remove the unmeasured benchmark from the default sweep ✅ fixed
Right, and it broke the property that matters most for a research artifact: **the documented command did not reproduce the committed study.** `walmart_amazon` was in the default while the rows covered four benchmarks, so a default `--resume` run would have added a fifth and rewritten the report. Removed, with a test asserting `set(BENCHMARKS)` equals the benchmark set in the committed rows — that test fails on the old default.

### P2 — Avoid calling the cross-model treatments identical ✅ fixed, and the headline claim was rebuilt
This was the most valuable finding in either round, because it invalidated a **conclusion**, not an implementation detail. You are right that `er_in_official_template` is not one prompt: Gemma drives `task: entity resolution | query: ` on **both** sides, Qwen3 drives `Instruct: …\nQuery:` on the **query side only** with bare documents. Their gap confounds model, wording and side placement, so it cannot support "the effect is model-specific" — which is exactly what I used it for.

The arm itself is correctly designed (each checkpoint gets its own template shape); the *description* was wrong. The claim now rests on `er_symmetric`, the one arm that really is a single byte-identical string on both sides for all five checkpoints: on `wdc_computers` it spans **−0.1038** (e5) to **+0.0004** (Gemma), a spread of **0.1041** from one prompt held fixed across five models. The doc now carries an explicit note on which arms are cross-model comparable and which are not.

---

### Verification

Beyond the unit tests (13 pass), two checks that the guards can actually observe what they claim:

- **The revision guard was made to fail.** Against a wrong pin it raises; against the right one it resolves. A guard never seen to fail is a hypothesis.
- **Cold-cache reproduction.** Under the new revision-keyed namespace (so nothing was served from the old cache), an end-to-end run of the control model on `fodors_zagat` reproduced the committed rows **exactly** — 3 rows × 8 measured fields, **0 mismatches**. That is evidence the six fixes changed no measurement, and that the study reproduces from cold.

`ruff` clean · `ruff format` clean · `mypy src/` clean (215 files) · fast suite **4719 passed, 2 skipped**, coverage **96.30%** (floor 90).

No `src/` changes in this round — the revision pin is resolved through `huggingface_hub` in the harness, deliberately not by adding a `revision` field to `SentenceTransformerEmbedderConfig`, which would be a contract change carried by a research PR.

fxd24:
@codex review

fxd24:
Round 3: all seven were real. Fixed in `840d680`. One of them was **data destruction introduced by round 2's own fix** — the second time in this PR that a fix created the next finding, which is itself worth recording.

### P1 — Validate requested arms before clearing measurements ✅ fixed
The worst bug in the whole PR, and mine. Round 2 added `_drop_cell` *before* evaluation so a crashed rerun would look partial. That made an unknown arm name **destructive**: `--arms offical_asymmetric` matched no recipe, cleared every requested cell, evaluated nothing, and **exited 0**. A typo silently deleted committed measurements and reported success.

Unknown arms are now rejected before anything touches `--rows`, and a model with no applicable arm is skipped rather than cleared. The test asserts the **data survives** — it seeds the rows file with the real committed content and compares byte-for-byte afterwards, because an empty file would pass trivially. I disabled both guards and watched it fail before trusting it.

The general lesson I'm taking: **a fix that adds an early write extends the blast radius of every input-validation gap upstream of it.** Round 2 moved a destructive operation earlier without moving validation earlier with it.

### P2 — Load the explicitly pinned checkpoint revision ✅ fixed
Correct, and it defeated the point of round 2's fix. Comparing `spec.revision` against `main` only *detects* drift — it cannot survive it. The moment upstream advanced, the pinned commit could be sitting in the cache while the run aborted, and the documented remedy ("fetch the pinned revision") was unachievable. Now `snapshot_download(..., revision=spec.revision)` resolves the pinned commit directly and the embedder loads **that snapshot directory**, so the weights are pinned whether or not `main` has moved.

### P1 — Include experiment identity in resume completeness ✅ fixed
Agreed. `Row` now carries `revision` and a `recipe_fingerprint` (a hash of the two prompt strings — the thing that actually defines the treatment), and `_cell_complete` requires both to match.

One deliberate choice: a row **without** provenance counts as *unknown*, never as a match. That means the committed rows — which predate the fields — are recomputed by `--resume` rather than back-filled. I could have written `spec.revision` into them, and the inference would probably be right (an unpinned load resolves `main`, which is what these hashes are), but it would be asserting a provenance that was never actually recorded. Recomputing costs cached-vector time; a wrong provenance claim costs the artifact's credibility.

### P1 — Qualify the raw-prompt conclusion with the Gemma result ✅ fixed
This is the second conclusion-level finding, and you are right that the evidence table was selective: it listed five negatives and omitted that EmbeddingGemma is **significantly positive** under the same `er_symmetric` arm on `amazon_google` (`+0.0118 [+0.0046, +0.0196]`) and neutral on `wdc_computers`.

"Reliably harmful" reads as a universal law; the rows say **4 of 5 checkpoints**. Both the tier table and the summary now say that, the Gemma row is in the evidence table, and there is an explicit note that it is kept deliberately. It does not rescue the recommendation — Gemma's *documented* retrieval prompt still beats its raw-sentence arm on all three unsaturated benchmarks — but a 4-of-5 result stated as a universal rule is exactly the overreach this document exists to avoid.

### P2 — Regenerate the report with the corrected snapshot hashes ✅ fixed
Right — I corrected `ModelSpec`, `PROMPT_SOURCES` and the findings doc but left the *generated* report carrying `c9745ed1…` and `c54f2e6e…`, so the two committed artifacts disagreed on provenance. Regenerated; the superseded hashes now appear in exactly one place, the findings doc's correction note, where they are labelled as superseded.

### P2 — Escape prompt pipes in the Markdown table ✅ fixed
Confirmed: backticks do not protect a pipe from GFM's table parser, and Gemma's own prompts contain `title: none | text: `, so those rows rendered with extra misaligned cells. Added a `_cell()` helper that escapes delimiters, applied to the prompt columns and the sources column.

### P2 — Render benchmarks present in custom result rows ✅ fixed
Agreed — a `--benchmarks` run wrote rows and intervals that no headline table displayed, so a measurement could be paid for and then hidden. Order now follows `BENCHMARKS` for the known ones and appends any extras present in the rows.

---

### Verification

`ruff` clean · `mypy src/` clean (215 files) · fast suite **4720 passed, 2 skipped**, coverage **96.30%** · harness guard tests **14 passed**, including new coverage that the arms guard, the provenance guard and the cell-drop behave as claimed — each checked by making it fail first.

fxd24:
@codex review

fxd24:
Round 4: all four were real. Fixes are in `3a4dbf1` (code) and `c5fda79` (docs); head is now `8c9d33d8`, which also merges `origin/main`.

The first finding is the **third consecutive round in which the same destructive write produced a defect**, so I stopped patching it and deleted it.

### P1 — Stage replacement rows before dropping the existing cell ✅ fixed by removing the deletion entirely — `3a4dbf1`
You found three more ways the up-front `_drop_cell` destroys data: a failed snapshot or benchmark load wipes the cell before anything can replace it; a **valid** narrowing selector like `--arms none --k 20` silently deletes every unselected measurement; and the shipped legacy rows fail the provenance check and get deleted on sight.

Rather than stage-then-swap, I removed the deletion. Its purpose is already served by the provenance fields from round 3: a row is only reusable when its `revision` **and** `recipe_fingerprint` match, so a cell rebuilt under changed weights or edited prompts is rejected by `_cell_complete` regardless of what is lying beside it, and `merge_rows` replaces each `(model, benchmark, arm, k)` as its replacement actually arrives.

The through-line, since it cost four rounds: **I was deleting data to maintain a bookkeeping invariant.** Recording what the data *is* achieves the same invariant, costs nothing, and cannot lose a row. Every guard layered on top of the deletion — validate `--arms`, skip empty selections — was treating symptoms of a design that should not have existed.

### P1 — Validate every prompt partition in the cache ✅ fixed — `3a4dbf1`
Correct on both counts. `DiskCachedEmbedder` keys each explicit query prompt separately, so checking only the document/default partition left every bge / Qwen3 / ER prompted partition unvalidated — and, as you note, the prompt-reach guards cannot substitute, because stale prompted vectors still differ from the baseline and therefore look like a working prompt.

The canary now probes **every** partition the recipes will read, and only **populated** ones: it looks the entry up via `_get_from_db` first, because on a miss the wrapper delegates to the base embedder and stores the result, so comparing a miss to a fresh encode compares a value to itself — a check that cannot fail. A test asserts the empty-cache case performs **no** probe at all.

### P2 — Create the report parent directory before writing ✅ fixed — `3a4dbf1`
Right, and the failure mode was nasty: it raised only *after* an expensive cell had finished, aborting the rest of the sweep. `mkdir(parents=True, exist_ok=True)` now precedes both the sweep's write and `--render-only`.

### P2 — Regenerate the shipped rows with provenance ⏳ in flight, 208 of 400 verified identical
Agreed, and worth insisting on: without `revision`/`recipe_fingerprint` the raw artifact could not substantiate the findings document's pinned-checkpoint claim, which is the entire point of having pinned it.

I re-ran the **whole sweep** rather than back-filling. Back-filling would have been an inference — probably a correct one, since an unpinned load resolves `main` and that is what the pinned hashes are — but writing an unverified provenance into a research artifact is exactly the error class this PR keeps surfacing.

Status: the first three models (MiniLM, bge, e5 — 208 rows) are complete, carry provenance, and match the committed study at **0 mismatches across 208 rows × 13 measured fields**, from a **cold cache** under the pinned revisions. Gemma and Qwen3 are running now. **No number has moved so far**, and I will state explicitly if one does before this is merged.

---

### The re-run hit a silent deadlock, and the diagnosis is worth recording

It hung for **3h16m at 0% CPU** — no error, no traceback — right after logging the `embeddinggemma-300m` load. Three models had already completed, so it presented as "a problem with Gemma" rather than an environment problem.

Ruled out by measurement rather than reasoning: not a download or auth stall (`lsof -a -i` showed **zero** network connections, with the 1.2 GB weight blob mmap'd locally); not the local-snapshot-path load introduced in round 3 (that path loads in **1.3 s** standalone); not detached/stdin, since it hung identically in the foreground.

The isolating experiment:

| condition | Gemma load |
|---|---|
| no faiss imported | **1.3 s** |
| faiss imported + index built, `KMP_DUPLICATE_LIB_OK=TRUE` only | **never returns** (killed at 180 s) |
| same, plus `OMP_NUM_THREADS=1` | **1.6 s** |

It is the faiss/torch duplicate-OpenMP conflict on macOS — but **`KMP_DUPLICATE_LIB_OK` suppresses the abort, not the deadlock**; `OMP_NUM_THREADS=1` is the variable that matters. A fresh git worktree has no `.env` (gitignored), so `uv run --env-file .env` fails there and the tempting move is to drop the flag and set `KMP_DUPLICATE_LIB_OK` by hand — which lands you in exactly this state. Recorded in `docs/FRICTION_LOG.md` and the harness docstring (`c5fda79`), because a run can have the variable set, emit nothing at all, and hang forever.

### Two older threads that are resolved but may still read as open

- **`Report deltas using the displayed recall estimator`** (13:24 review) — resolved in `12b9189`, deliberately as *documentation* rather than a change of estimator. The delta stays macro-over-records because a paired bootstrap needs a per-entity score to resample by cluster and a corpus-wide micro ratio has no per-entity decomposition. The field was renamed `delta_per_record_recall`, and the divergence (bge/wdc: micro `+0.1098` vs macro `+0.1224`) is stated in the rows, the generated report header and the findings doc.
- **`Invalidate old rows before rerunning a cell`** (14:22 review) — resolved, but **the mechanism moved**, so searching for `_drop_cell` will find it gone. It was fixed by `_drop_cell` in `b350e4e`, then that function was *deleted* in `3a4dbf1` because it was itself destroying data (the P1 above). The guarantee now lives in `_cell_complete`, which requires a matching `revision` and `recipe_fingerprint` — a cell whose rows came from different weights or different prompt text is not complete, so it is recomputed rather than silently reused.

### A note on my own monitoring, since it is this PR's subject

My progress monitor reported the re-run "finished" at 20 of 400 rows. The run was healthy; the monitor asked `pgrep` whether the process was alive, `pgrep` ran sandboxed and could not see the process at all, and **"cannot observe" came back as "not running."** Exactly the failure this PR is about — a check reporting a clean result about something it is structurally unable to see. Rewritten to watch row-count progress and treat a stalled log as the failure signal, both of which it can actually observe.

fxd24:
@codex review

fxd24:
## Round 5 disposition — head `2e4ba3f`

Three findings from the last round, plus a re-run. All addressed. **One of them changed the study's answer**, so that is first.

---

### 1. Multiplicity — Holm, and it cost us e5

The sweep publishes **80** separate 95% intervals. Read one at a time each controls its own error rate; read as the set a claim like *"this recipe helps this checkpoint"* is actually read, they do not.

Each `(model, arm)` is now one **family** across benchmarks, corrected with **Holm's step-down** at family-wise α=0.05. Holm is uniformly at least as powerful as Bonferroni and assumes nothing about dependence, which matters because the arms of one model share a baseline and a corpus.

**The verdicts are emitted by the harness, not typed.** Every headline row in `20260728_prompt_axis.md` carries a `Holm` column (`**holds**` / `withdrawn` / `-` / `n/a`) beside its interval, so the `**` marker cannot be read without the correction next to it, plus a full section listing every comparison and p-value.

**The p-values are approximate, and that is stated in the artifact.** Holm needs p-values; the rows carry percentile intervals and the bootstrap replicate distributions were not retained (the exact route costs the whole sweep). Each p is recovered by reading the interval as normal and scaling from the bound **facing zero** only — the side that decides exclusion — so no symmetry is assumed. Two properties then follow by arithmetic rather than judgement, which is what makes it safe for a *retraction*:

- a spanning interval yields `p > 0.05`, so **correction can only remove claims, never add one** — nothing below was promoted;
- a bound resting *exactly* on zero yields `p = 0.05` exactly, so it fails every Holm threshold in a family of more than one. The four such cells drop out without anyone deciding they should.

**Counts, recomputed from the 400 committed rows, not from memory.** 80 comparisons → **61 testable**, **19 not** (the arm changed no record's recall, so estimate and both bounds are exactly zero and the bootstrap has no support; all 19 on the saturated `fodors_zagat`). Of the 61, **40 intervals exclude zero uncorrected, 37 survive**.

**All three withdrawals are `intfloat/e5-base-v2`, and they are its entire documented story:**

| withdrawn | Δ per-record | 95% CI | approx p |
|---|---:|---|---:|
| `official_asymmetric` / `abt_buy` | +0.0069 | [+0.0010, +0.0128] | 0.0221 |
| `official_symmetric` / `amazon_google` | −0.0029 | [−0.0059, −0.0004] | 0.0229 |
| `official_symmetric` / `wdc_computers` | −0.0145 | [−0.0274, −0.0023] | 0.0197 |

Both e5 families fail at Holm's **first** step (smallest p `0.0221` and `0.0197` against a `0.05/3 = 0.0167` threshold), so **no e5 documented arm is distinguishable from no prompt on this data**. All three were *just barely* significant (p 0.0197–0.0229) — exactly the population a correction exists to thin out.

Consequences carried through the prose:

- **Tier 1 is now "three of the four instruction-trained models", not four.** It is in the four-line summary, in the §4.2 tier table, and in the e5 section.
- **The arm-versus-arm argument is withdrawn with them.** An earlier paragraph argued asymmetric beats symmetric on `wdc_computers` even though neither beats no-prompt, on non-overlapping intervals. That was never a test, the arms are paired so their errors are correlated, and both cells have now failed against the baseline. A comparison between two effects neither of which is established is not evidence.
- What survives for e5 is the **negative** result about its card, which is still the useful one: following the card's symmetric advice bought no measurable improvement, and the recipe it advises against is not measurably better either.

**Nothing else moved.** Gemma, bge, Qwen3 and the MiniLM control survive unchanged — including the `official_clustering` harm on the saturated benchmark (`p = 0.0061` at the last step of a family of four) and the EmbeddingGemma counter-example in §4.2 (`p = 0.0013`).

**Sensitivity is reported, because the family is a choice.** Correcting over the whole sweep as one family of 61 leaves **24** standing rather than 37. It costs margin, not conclusions: Gemma still clears it on all three benchmarks, bge on two, Qwen3 on one — so "positive on ≥1 benchmark for three of four" holds under either family.

**A bug I introduced and had to be caught by its own test.** The first version classified boundary cells with `p == ALPHA` and printed *"0 testable comparisons sit exactly on the boundary"* — over four real ones, because the recovered p lands on `0.05000000000000004`. Same shape as the failures in this repo's rules file: a check reporting all-clear because it cannot observe what it exists to catch. It reads the stored bound now (exact), and `test_a_boundary_cell_is_found_structurally_not_by_comparing_p_to_alpha` asserts the discarded classifier would have to be blind.

### 2. `merge_rows` raises on two checkpoints of one model — and deletes nothing

`_cell_complete` already refuses to *resume* a cell whose revision moved, but it decides one cell at a time: a run narrowed with `--benchmarks`/`--models` recomputes only what it was pointed at and the rest keep the old checkpoint's numbers, leaving the merged file reporting two sets of weights under one model heading with nothing on the surface to show it. Only a check over the whole merged set can see that.

It **raises and names the conflict** rather than dropping the stale side. Which revision was meant is a question about intent that this function cannot answer — and an earlier draft of this harness resolved a similar ambiguity by deleting, which destroyed measured rows three review rounds running.

### 3. Render-time provenance assertion

`render_report` printed "Where each prompt came from" and "The arms" from today's `MODELS` while the numbers came from the rows file, with nothing coupling the two. An edited prompt string plus a re-render — no re-measurement — produced a document that *describes* one experiment and *reports* another, with no visible seam.

It now refuses to render unless every row's `revision` and `recipe_fingerprint` match the spec and recipe the prose describes. Absent provenance fails: a row that cannot say what produced it cannot be shown to agree. Tested by mutation on each of the four branches, with the committed 400 rows as the positive control so the guard is not vacuously passing.

### 4. Cold-cache re-run (from the previous round) landed

Every measured field reproduced **exactly** across all 400 rows — including Qwen3 and Gemma — with `revision` and `recipe_fingerprint` now populated and each checked against its model's pinned `ModelSpec.revision` (populated-but-wrong fails, not just null). The regenerated report was byte-identical.

### 5. MPS ceiling documented as a real limitation

The single-process sweep OOMed at row 324 of 400 with torch requesting **42.44 GiB** on MPS while loading a **0.6B fp16** model — the MPS allocator caches freed blocks across model loads, so a long enough sweep exhausts it. Fixing the one real leak (`blocker` pinned the embedder from the inner loop) moved the ceiling without removing it.

Remedy is one `(model, benchmark)` per process, which cell-granularity `--resume` plus the on-disk embedding cache makes free. In `docs/FRICTION_LOG.md` with the driver loop, and in the findings' new **§5 Limitations** as a constraint on reproduction — a bare `python prompt_axis.py` with no narrowing will not finish on this machine. The trap worth naming: it fails **late and partially**, so it reads as *"Qwen3 is too big"* rather than *"this process has been alive too long"*; running Qwen3 first would have hidden it entirely.

§5 also states the remaining bounds honestly: the approximate p-values (and that the three withdrawn cells are the ones nearest a threshold, missing by 18%, 33% and 38%), one `k`, one seed, one size per model family, four English product/restaurant benchmarks with no person names or dedup shape, and that the strongest result in the sweep is the one model that is **not** OSI-licensed — with bge (MIT) named as the shippable recommendation.

### 6. The retraction had to be carried into the index, and nearly wasn't

`docs/research/README.md` is the row a reader sees first, and it still advertised *"significantly positive on ≥1 benchmark for all four instruction-trained models"* and *"e5's card … loses"* — both withdrawn. Fixed, along with the summary's *"see the next section"* cross-reference, which pointed at §1 rather than at §4's correction subsection.

Worth naming because of **how** it was nearly missed: the sweep meant to catch stale claims was `grep -rn "instruction-trained" | grep -v 20260728_prompt_axis_findings.md`, and the README line **contains that filename as a link target** — so the exclusion filter hid the one file that needed the fix. A second, differently-shaped pass (grepping the *values* — `0.1224`, `three tiers`, `prompt_axis`) over `docs/` and `README.md` found it and confirmed no other document carries this study's numbers.

---

### Verification

Every number in the new prose was recomputed from the 400 committed rows by script, not recalled: **25 of 25** `Δ [lo, hi]` triples present in the rows, all **7** counts, all **6** quoted p-values, both recall figures. 0 mismatches.

### Gate

- `pytest -m "not integration and not slow and not finetune"` → **4775 passed, 2 skipped**, TOTAL 96%
- `ruff format --check .` + `ruff check .` → clean
- `mypy src` → **no issues in 216 source files**
- `tests/examples/test_prompt_axis.py` → **24 passed** (9 new this round)

CI at `2e4ba3f`, read per check-run against that exact SHA rather than off the PR (a check that settled against an earlier push is not a verdict on this one): **CodeQL, Analyze (python), lint, build, test (3.13), test-core-only, test-finetune, docs-clean-install — all `success`**. `deploy`, `test-full` and `Sourcery review` report `skipped`, which is their normal state on a PR here.

@codex review

Please review against `2e4ba3f` specifically. The highest-value target is **§1**: whether the family definition, the interval→p-value recovery, and the retraction it produces are defensible — and whether any surviving claim in the findings doc still rests on a comparison that was withdrawn.

fxd24:
@codex review

Round 6 fixes are in at `64e2ac1`. What changed since your last pass:

- **The publish guard can now observe its own failure.** A headline row with an interval but no p-value was dropped by the correction, which shrank its Holm family, while the report printed a family size counted from the survivors. One predicate (`_missing_p_value`) now backs both `--resume` and the publish guard, and the family size is asserted against the *declared* benchmark count rather than against itself. Both have tests that fail without them.
- **Holm via adjusted p-values**, tie-safe. Verified rather than assumed: it gives verdicts identical to stop-at-first-failure on every family in this artifact, so it moved no published number.
- **B=20000 justification arithmetic redone** under `sqrt(p(2-p)/B)`: 0.00629 at B=1000 (half the 0.0125 threshold), 0.00141 at B=20000. The prior text carried the right formula and the wrong numbers under it.
- **The "if and only if" is now the precise statement**: `p <= t` *implies* the interval excludes zero, exactly; the converse can fail inside `4/(B+1)`, and the p-value is the conservative side.
- **`withdrawn` no longer conflates two failures** — a claim multiplicity removed vs one that never cleared the uncorrected α (`n.s.`).
- **The dependence argument** now points at the family being corrected (disjoint benchmarks, near-independent) rather than at across-family dependence.
- **Rows swapped to the completed B=20000 re-run.** This changed the answer: the retraction list goes from **three** cells to **five**. The two additions are bge's, and they are what the normal-approximation p-value was hiding.
- **46 hand-typed intervals in the findings doc were stale** after the re-run. Rewritten from the rows, and a new test now refuses any interval in the prose that no measured row holds — proven to fail by reintroducing one of the 46.

`experiments/statistics.py` coverage measured explicitly since no per-PR gate covers `experiments/*`: **97%**.

fxd24:
Head moved to `adcaf6c` — two documentation corrections found by re-deriving claims from the rows rather than re-reading the prose. CI was **all green** at `64e2ac1` before these two docs-only commits.

1. `fodors_zagat` has 25 headline rows at k=20, but five of them **are** the no-prompt baselines. "Of the 25 arms measured against it" counted the baselines among the things measured against themselves; it is 20 prompted arms against 5 baselines. Conclusion unchanged.

2. More substantive: the `er_symmetric` table's caption called EmbeddingGemma "the exception" to that arm's harm, while the table itself **omitted Gemma's own significant harm** on `abt_buy` (−0.0078, holds under Holm). Qwen3's −0.0907 and bge's −0.0368 were missing too. The arm significantly harms **all five** checkpoints somewhere; Gemma is the only one it also helps somewhere. The original framing ("reliably harmful") over-claimed, the correction to it over-swung, and both were selections from the same rows — the table now lists every `er_symmetric` cell that survives the correction.

@codex review — please review at `adcaf6c`.

fxd24:
Head `944a300`. A review pass over the round-6 fixes found **a false claim I had just published**, plus two weaker items. All three fixed.

**1. The band-vs-noise inequality was asserted at exactly the point where it fails.** The docstring said the `4/(B+1)` converse band is "narrower than **one** Monte-Carlo standard error at every attainable p". It is not: `sqrt(p(2-p)/B)` grows with `p` and the curves cross near `p = 8B/(B+1)²`, so at the smallest attainable p-values one standard error is *half* the band — at B=20000 the floor `p = 2/(B+1)` gives `1.0e-4` against a band of `2.0e-4`. **Measured: 25 of the 80 headline comparisons sit at one of the three p-atoms where the sentence is false.** It was rendered verbatim into the published report. **Three** standard errors clear it at every attainable p and at both B — and 3 is already the margin `_resolution_limited` uses, so the sentence now supports the conclusion it draws instead of contradicting it.

**2. My tie-safety rationale for the adjusted-p Holm was false.** The threshold form was *already* tie-safe: thresholds `α/(m−i+1)` increase with `i`, so if the earlier member of a tie passes so does the later one, and if it fails the walk stops before reaching it. And since `α < 1`, the two forms are identical for **every** input, not just for distinct p-values. The code was correct; the reason I gave for it was not — a change justified by a property that was never at risk, which is the same "prose drifts from the artifact" shape this PR keeps hitting. The docstring now states the equivalence and keeps the adjusted-p shape for legibility.

**3. The family-size guard had a false negative one level up.** Sizes are counted over families *present in the verdicts*, so a family that lost **every** benchmark vanished rather than failing — `sizes` still read `[4]` and the guard passed. Reachable through the documented CLI: `--models intfloat/e5-base-v2` against a fresh rows file renders a report whose prompt-source and arms tables describe all five models while the results describe one. The observed family set is now checked against `MODELS`, and the new test first asserts the size check is *blind* to this case before asserting the new guard catches it.

Also: the Holm oracle in the test was built on the same `_holm_steps` as the code under test, so a wrong multiplier would have corrupted both sides identically — it is now written independently, and the step multipliers are pinned separately. Three stale/ambiguous docstring passages fixed (a `_cell_complete` parenthetical about the committed rows that is no longer true, the "understates by ~40%" direction, and the converse-failure mechanism).

Note on the previous run: `test-finetune` failed at `adcaf6c` in step 3, *"Install uv and set the python version"* — an infrastructure step, on a docs-only commit that had passed the same job minutes earlier at `64e2ac1`. Not a regression; this push re-runs it.

@codex review — please review at `944a300`.

fxd24:
Head `0a1d8d1`. All six findings from the last review addressed; four were real, and the P1 was **a regression I introduced one round earlier**.

**P1 — the family guard killed the documented sweep.** Correct, and worse than reported: `prompt_axis_sweep.sh` runs one `(model, benchmark)` per process under `set -e`, and `main()` enforced publication at the end of *every* invocation. Against a fresh rows file the first cell leaves the design incomplete, the guard refuses, `main` returned 1, and the sweep died after its most expensive step. The D2 guard turned "renders a partial report" into "kills the sweep".

Fix: a run narrowed by `--models`/`--benchmarks`/`--arms` logs that it is not publishing and exits 0; the sweep now publishes once at the end via `--render-only`, which is where a refusal is a real failure. **Verified by running both against a deliberately partial rows file** (e5 only): scoped invocation exit **0**, `--render-only` exit **1**. A test pins the two halves together, because either alone is satisfiable by breaking the other. (My first attempt to verify this read `tail`'s exit code through a pipeline rather than python's — same shape as everything else this PR has been fixing, so: re-measured without the pipe.)

**P2 — family membership vs count.** Correct, and it was wrong in *both* directions at once: three declared benchmarks plus a substituted fourth passed as complete, while a coherent `--benchmarks` sweep adding a fifth to every family was rejected — your two P2s are the two halves of the same bug. The guard now compares each family's benchmark **set** against the other families', which is the property Holm actually needs: what invalidates the correction is a denominator that varies with the outcome, i.e. two families disagreeing about what they contain. A deliberately smaller or larger design is coherent and every table names the benchmarks it used. Tests updated: the "short family" case now drops a benchmark from **one** family (dropping it from all is a different, legal design), and a new test asserts a uniformly-narrower sweep still publishes and reports `m = 3`.

**P2 — atomic rows write.** Correct and accepted. That file is the sole artifact of an hours-long sweep and is rewritten in full after every arm; this repo has already lost a paid run to this exact class of accident. Now a sibling temp file plus `os.replace`.

**P2 — stale `0.0010` in the limitations section.** Correct. It survived because I fixed the number where the argument is *made* and not where it is *restated* 200 lines later. Fixed.

Local gate: 4791 passed, ruff + mypy clean. CI was all green at `944a300` including `test-finetune`, confirming its failure at `adcaf6c` was the transient `Install uv` step.

@codex review — please review at `0a1d8d1`.

PR #253, opened 2026-07-28
measure + ship: a per-score-family default threshold (sim_cos 0.5 -> 0.90; heuristic stays)

fxd24:
## What this does

`MethodSpec.default_threshold` has existed since the v0.3 registry unification —
declared per method, documented, and read by **nothing**. Six front doors
hard-coded their own `0.5` instead, so the field could say one thing while the
shipped behaviour said another and no test would notice.

This PR (a) **measures** whether a better fixed constant exists, (b) **wires**
`threshold=None` to the registry, (c) **sets** constants only where the
measurement supports it, and (d) fixes three `threshold` parameters that
silently did nothing.

Every table below is copied from generated harness output; no number in this
description is retyped from memory.

**Headline:** `sim_cos`'s default moves `0.5` → **`0.90`** (measured, big).
`heuristic` **keeps `0.5`** — also measured, and deliberately rejected. The LLM
and fitted-matcher families are untouched and this PR is not evidence about them.

---

## 1. Measured (Phase 1)

PR #250 measured that *deriving* the cut from labels beats `0.5` in 45 of 54
cells and recommended fixing the constant next — but its §5.2 is explicit that it
never scored a *shared replacement constant* anywhere. So the derived cuts'
medians were a **documented prior, not a measurement of the constant**, and
shipping them would have been exactly the error this repo keeps re-learning.

New harness: `examples/research/threshold_constant_sweep.py` ($0, offline,
committed before it was run). Artifacts:
`examples/research/results/threshold_constant_sweep{,.e5}.json`. Write-up:
`docs/research/20260728_threshold_constant.md` — **every table in it is generated
from the artifacts** by `tools/render_threshold_constant_writeup.py`, because
three PRs on 2026-07-28 shipped factual errors and every one was in hand-typed
prose beside a correct generated table.

Protocol, each part load-bearing:

- **Held-out means a disjoint corpus.** `Benchmark.split` assigns whole gold
  clusters to one side, so no entity and no match pair straddles the boundary.
  The label-derived cut is fitted on train; every reported number is from test.
- **Leave-one-benchmark-out selection.** The constant for benchmark *b* is chosen
  from the *other* benchmarks only, so the reported number is out-of-sample with
  respect to the dataset — which is what shipping a default actually asks.
- **Paired cluster bootstrap.** The resampling unit is the gold cluster, never
  the pair row (pairs touching one entity are dependent; resampling rows gives
  intervals far too narrow). Each pair is assigned to `min` of its endpoints'
  units, so it is counted exactly once. Precisely: a *gold* pair has both
  endpoints in one cluster, so the recall side is exact; a cross-cluster
  candidate pair is attributed to one endpoint, so the precision side
  under-counts co-movement and intervals are slightly **too narrow**. Disclosed
  rather than changed — see "Review findings" below for why no verdict turns on
  it.
- **Exact oracle**, sweeping every distinct score — so "how much of the
  achievable gain does a constant capture?" is measured against the true ceiling,
  not the grid's best point.
- **Never an average across benchmarks in a reported number.** Aggregation
  appears in exactly one place, the LOBO *selection* criterion, as a median over
  benchmarks of a mean over seeds, and is labelled a selection statistic.
- **The ship rule was pre-registered** in the harness (`SHIP_RULE`) before the
  numbers existed; the verdict table is computed by applying it to the artifact.
  The checkpoint-transfer rule (`TRANSFER_RULE`) was likewise pre-registered
  before the variant artifact existed.
- **The self-check gate is real.** `_assert_matches_classify_pairs` recomputes
  three grid points per cell through the library's own `classify_pairs` and
  aborts on any disagreement > 1e-9. A test feeds it a deliberately drifted curve
  and requires the error — a gate nobody has watched fail is a hypothesis, not a
  safety net.

### The verdict, computed from the artifact

> Pre-registered rule: **ship iff (1) LOBO constants span <= 0.05, (2) no eligible benchmark is significantly worse (95% CI entirely < 0) in a majority of its seeds, and (3) the median per-benchmark mean delta-F1 is > 0**.

| family | LOBO spread | benchmarks significantly worse | median per-benchmark Δ | verdict | constant |
|---|---|---|---|---|---|
| `heuristic` | 0.03 | abt_buy | +0.0339 | **DO NOT SHIP** | 0.74 |
| `sim_cos` | 0.03 | none | +0.5842 | **SHIP** | 0.90 |

### `sim_cos` — why this is not marginal

Leave-one-out constant, graded on the held-out benchmark (extract; full table in
the write-up). `F1@0.50` is the incumbent, `capture` is the fraction of the exact
oracle's headroom recovered:

| family | benchmark | seed | t | F1@0.50 | F1@t | Δ | 95% CI | oracle | capture |
|---|---|---|---|---|---|---|---|---|---|
| `sim_cos` | abt_buy | 0 | 0.89 | 0.0666 | 0.2446 | +0.1781 | [+0.1423, +0.2208] | 0.2479 | 98% |
| `sim_cos` | amazon_google | 0 | 0.89 | 0.0171 | 0.4252 | +0.4082 | [+0.3596, +0.4529] | 0.4463 | 95% |
| `sim_cos` | dblp_acm | 0 | 0.90 | 0.2200 | 0.8808 | +0.6609 | [+0.6338, +0.6861] | 0.8974 | 98% |
| `sim_cos` | dblp_scholar | 0 | 0.88 | 0.0129 | 0.6377 | +0.6249 | [+0.5784, +0.6664] | 0.6983 | 91% |
| `sim_cos` | febrl_dedup | 0 | 0.91 | 0.0689 | 0.6559 | +0.5870 | [+0.5511, +0.6217] | 0.8366 | 76% |
| `sim_cos` | febrl_person | 0 | 0.90 | 0.0688 | 0.7935 | +0.7247 | [+0.6694, +0.7771] | 0.9078 | 86% |
| `sim_cos` | fodors_zagat | 0 | 0.90 | 0.0630 | 0.9254 | +0.8623 | [+0.7811, +0.9310] | 0.9394 | 98% |
| `sim_cos` | walmart_amazon | 0 | 0.89 | 0.0025 | 0.0517 | +0.0492 | [+0.0409, +0.0585] | 0.0701 | 73% |
| `sim_cos` | wdc_computers | 0 | 0.89 | 0.0129 | 0.0601 | +0.0471 | [+0.0366, +0.0579] | 0.0757 | 75% |

Every eligible cell improves and every interval sits entirely above zero — no
benchmark, no seed, loses. Look at the `F1@0.50` column for the reason: `0.5` on
a normalized-embedding cosine is not a mildly mistuned cut, it accepts nearly
everything the blocker proposed and precision collapses. **This is less "we found
a better constant" than "the shipped constant was never on the right scale."**

### `heuristic` — measured, and rejected. It is *not* instability

The obvious guess (rapidfuzz has no single constant because every dataset wants a
different one) is **wrong here**: LOBO spread is `0.03`, so dropping any dataset
barely moves the choice. Clause (1) passes. What kills it is clause (2):

| family | benchmark | seed | t | F1@0.50 | F1@t | Δ | 95% CI |
|---|---|---|---|---|---|---|---|
| `heuristic` | abt_buy | 0 | 0.74 | 0.0625 | 0.0083 | **-0.0542** | [-0.0682, -0.0420] |
| `heuristic` | abt_buy | 1 | 0.74 | 0.0630 | 0.0145 | **-0.0486** | [-0.0633, -0.0318] |
| `heuristic` | abt_buy | 2 | 0.74 | 0.0642 | 0.0227 | **-0.0416** | [-0.0592, -0.0219] |

Every seed, interval entirely below zero, while it helps 7 other benchmarks.
**Both F1 conventions agree** (the end-to-end all-gold table in §5 shows the same
margins, and `febrl_dedup`'s mild losses *harden* there) so this does not rest on
the denominator. A default that lifts the median while predictably harming a
known data class is a recommendation with an undisclosed victim — clause (2) was
pre-registered precisely so this could not be argued away afterwards by pointing
at a healthy median.

### A `sim_cos` cut belongs to the *encoder*, not the family tag

A validity threat that was easy to miss and is not hypothetical: **every
benchmark loader pins `all-MiniLM-L6-v2`** (7 files set it in code under
`src/langres/data/`; corroborated by the artifact's own generated
"Blocker checkpoint(s): all-MiniLM-L6-v2 (benchmark pin)" line)
while **`DEFAULT_EMBEDDING_MODEL` is `intfloat/e5-base-v2`**
(`core/model_ref.py:144`). Two models both emitting "cosine similarity" disagree
about what `0.9` means, so assuming transfer would be the same
prior-as-finding error this study exists to avoid. It was measured instead —
the identical protocol re-run on the shipped default, **in both directions**,
because one direction alone is a cherry-pick:

| direction | constant | pre-registered verdict on the other checkpoint |
|---|---|---|
| MiniLM-selected → e5 | **0.90** | **TRANSFERS** — 0 of 27 eligible cells significantly worse; median Δ +0.0004 |
| e5-selected → MiniLM | 0.96 | **DOES NOT TRANSFER** — significantly harms `abt_buy`, all 3 seeds |

So the two checkpoints **agree that `0.5` is the wrong order of magnitude** and
**disagree about how high** — which is exactly what "the scale belongs to the
encoder" predicts. `0.90` is the value that is safe on both; `0.96` is a bigger
win on e5 but has a demonstrated victim, i.e. the same failure that disqualifies
`heuristic`, applied to our own preferred answer rather than only to the one we
were ready to reject.

**Stated plainly, because it is the honest caveat:** on e5 `0.90` is *safe but
close to inert* — several benchmarks score identically to `0.5` there, because on
that encoder almost every candidate pair already sits above the cut. A shipped
default's job is to stop the out-of-the-box path being catastrophically wrong,
not to be optimal on your encoder. With labels, `derive_threshold` (PR #250's
seam) is the better tool for that — though **not because labels dominate**: see
the correction below.

---

## 2. Wired (Phase 2)

- `langres.core.score_type.DEFAULT_THRESHOLDS` — a read-only mapping over all
  seven `ScoreType` values, on the stdlib leaf, so it adds no import-graph edge.
- `resolve_threshold(threshold, score_type)` — the one seam every
  `threshold: float | None = None` parameter goes through. An explicit value is
  returned untouched, *including one equal to the default* (not the same
  statement as omitting it — hence the `mode="before"` validator on `MethodSpec`,
  since after validation "omitted" and "explicitly 0.5" are indistinguishable).
- Seven constructors converted: `FuzzyString`, `VectorLLMCascade`, `Retrieve`,
  `RetrieveRerank`, `RetrieveLLM`, `RetrieveRerankLLM`, `Reranker.for_schema`.
- **The wiring commit changed no value**: all 13 registered methods resolved to
  exactly what they had before. Eight specs stopped repeating the same two
  literals (2× `0.5`, 6× `0.7` — counted from the diff, not from memory). The
  constant then moves in Phase 3 as a separate, measured change.

**Precisely what was fixed, without overclaiming:** `MethodSpec.default_threshold`
*still* has no runtime reader — a front door has a score **family**, not a method
name, so it resolves through `DEFAULT_THRESHOLDS` directly. (Verified on
`origin/main`: `default_threshold` occurs in no `src/` file except
`method_registry.py` itself.) What changed is that the field and the shipped
behaviour now derive from the **same** map, so they can no longer disagree, and
`test_every_registered_method_matches_its_family` pins it — the drift that was
previously unobservable is now observable.

**`Reranker.for_schema` joined the scheme** (it used to *require* an explicit
threshold). It cuts a score produced by
`MatcherScore(WeightedAverageMatcher(feature_specs=<all>))` — the same class,
over the same feature set, as `FuzzyString`'s matcher. Same construction, same
scale, same measured family, so demanding a number there while `FuzzyString`
defaulted was an inconsistency rather than a safety feature.

**Behaviour change, and it is INFERRED not measured:** `VectorLLMCascade`'s
out-of-the-box cut moves `0.5` → `0.7`. That is the value the registry has
declared for `prob_llm` all along; this makes it read rather than ignored.
Sweeping the LLM families costs a paid completion per score and is a separate
study. `threshold=0.5` restores the previous behaviour.

---

## 3. Constants set (Phase 3)

Every entry in `DEFAULT_THRESHOLDS` is now in exactly **one of three states**,
documented per family in its docstring and **pinned against literals** in
`tests/core/test_score_type_defaults.py`:

| family | value | state | why |
|---|---|---|---|
| `sim_cos` | `0.5` → **`0.90`** | **measured, changed** | Wins on every eligible benchmark+seed, all CIs > 0; safe on both checkpoints tested |
| `heuristic` | `0.5` (unchanged) | **measured, deliberately kept** | A stable better-on-median constant (`0.74`) exists and the pre-registered rule **rejected** it — it reliably harms `abt_buy` |
| `prob_llm`, `prob_group_llm` | `0.7` (unchanged) | **not measured** | A paid completion per score; a portfolio grid sweep is a real invoice |
| `calibrated_prob`, `prob_fs`, `prob_rf` | `0.5` (unchanged) | **not measured** | *Fitted* matchers a label-free user cannot run at all — "best out-of-the-box constant" is not the same question |

**User-visible effect:** `Retrieve` constructed **without** `threshold=` now
cuts at `0.90` instead of `0.5`. Any explicit `threshold=` is untouched. It does
**not** change `RetrieveRerank` (review caught an earlier draft claiming it did):
that chain is `RetrieveOp -> Rerank -> ThresholdSelect`, whose single cut is
tagged `heuristic` and stays `0.5` — there is no cosine threshold in it to move. Two registered methods (`embedding`,
`embedding_cosine`) also now *declare* `0.90`, though as noted that field still
has no runtime reader.

**A note on the value-pinning tests.** Every pre-existing assertion in this area
compared `DEFAULT_THRESHOLDS` to *itself* (`assert x == DEFAULT_THRESHOLDS[f]`) —
correct for checking wiring, and **useless for checking the number**, since both
sides move together and editing a constant leaves the suite green. That is this
repo's recurring "gate decoupled from the thing it checks" failure, so the values
are now pinned once against literals, with the docstring saying that changing one
is a claim about measured data, not a code change.

---

## 4. Three inert `threshold` parameters (Phase 4)

Each claim verified by running the code, and each now pinned by a test rather
than only a docstring.

- **`RetrieveLLM` / `RetrieveRerankLLM` now warn.** Their chain ends
  `... -> Generate -> Parse -> ThresholdSelect`; `Parse` emits a `decision` with
  `score=None`, and `predicted_match` gives a decision precedence over any score.
  So the `ThresholdSelect` still drops `decision=False` rows and abstentions, but
  its *number* cannot change any outcome. Verified: byte-identical clusters at
  `threshold` 0.0 / 0.5 / 0.99, with `Retrieve` as a **not**-invariant control —
  without that control the invariance test would pass on a broken pipeline.
  Warned rather than raised: the parameter is on four public constructors and is
  swept by the experiment matrix, so a hard error would break callers to fix a
  no-op. Only an *explicit* value warns; `None` expresses no preference.
- **`RapidfuzzMatcher.threshold` documented as inert.** Range-checked, stored,
  never read — it is a ranker, so the caller's cut decides. Verified three ways:
  identical score at 0.0 / 0.5 / 1.0, `_build_rapidfuzz` passes no threshold at
  all, and there is no `config`/`from_config` to round-trip one. Its docstring
  examples no longer imply otherwise, and a stale "compatibility with Optimizer"
  claim is gone (no `Optimizer` exists — only
  `autoresearch.blocker_optimizer.BlockerOptimizer`).

Out of scope but recorded: PR #250's inventory also names a fourth decorative
threshold, `EmbeddingScoreMatcher.threshold`
(`core/matchers/embedding_score.py:86`), which only picks the `decision_step`
label and provenance. It is already honestly documented in its own docstring and
is untouched here.

---

## Measured vs. inferred, in one place

**Measured** — held-out, multi-seed, with cluster-bootstrap intervals, from the
committed artifacts: per-cell F1 at all 99 grid constants / `0.5` / the derived
cut / the exact oracle, for `heuristic` and `sim_cos`; the LOBO constant's Δ-F1
vs `0.5`; its stability under dropping a dataset; and checkpoint transfer in both
directions.

**Inferred, not measured** — `prob_llm`'s `0.7` (the registry's declared value,
now honoured instead of ignored — that is what changes `VectorLLMCascade`);
`prob_fs` / `prob_rf` / `calibrated_prob` (untouched); and `RetrieveRerank`'s
family tag, which is `heuristic` because `Rerank`'s `out_space` defaults to it
even though it cuts a **cross-encoder's** score — a coarse tag over two different
scales, called out in the class docstring. Nothing from rapidfuzz data justifies
a constant for it, and none was set.

---

## Review findings addressed

Codex reviewed commit `910e3bc`; an independent claim-audit re-rendered all
~139 numeric cells from the artifacts and checked ~45 structural claims. Each
finding was verified against the code before acting, not relayed.

**Fixed — factual errors in my own prose, all of the exact class this PR names as
its failure mode (hand-typed text disagreeing with a correct generated table):**

1. **`RetrieveRerank` is not affected by the `sim_cos` change.** Its cut resolves
   against `heuristic` (`retrieval.py:386`) and stays `0.5`. `Retrieve` is the
   only front door that moves. Corrected in the PR body and CHANGELOG.
2. **"Deriving from labels is worth substantially more than any constant" is
   false, and my own ladder table said so.** The LOBO constant scores *higher*
   than the label-derived cut on **5 of 9** benchmarks, for both families.
   Neither dominates. Corrected in three places to the claim the data supports:
   deriving wins *where it matters* — on `abt_buy`, the benchmark whose harm
   vetoes the constant, it beats both the constant and `0.5`.
3. **The `Retrieve` op stamps rows `score_type="heuristic"`, not `"sim_cos"`**
   (`resources/retrieve.py:174`, hard-coded). `0.90` is still correct — the
   scores really are cosines, which is what was swept — but it is justified by
   *what the score is*, not by the tag. One coarse tag now covers three scales
   (rapidfuzz, cross-encoder, cosine). Disclosed in the docstring and write-up;
   **fixing the tags is a resource-contract change, and deliberately not done
   here.** Tracked as **#254** so the disclosure does not go invisible — a
   disclosed-but-untracked limitation is one nobody finds in a month.
4. **The bootstrap's stated justification was overclaimed** ("an entity's pairs
   move as a block"). Now stated precisely (exact on recall, approximate on
   precision). Not re-run, because the alternative — requiring both endpoint
   clusters to be drawn — makes inclusion quadratic and measures a different
   estimand, and no verdict is marginal: `sim_cos` ships on point estimates
   positive in every eligible cell, which no interval width reverses, and
   `heuristic`'s veto fails safe (its effect is to keep the incumbent).
5. **Resume could publish a lying header.** Cells are *reused* but the report
   header is rewritten from the new flags, so resuming under a different
   `--resamples` published old intervals under a new count, and a different
   `--embedder` pooled two cosine scales. `main()` now refuses both, with a
   negative-control test proving the guard is selective rather than blanket.
6. **`docs/reference/architectures.md` pinned `ThresholdSelect(0.5)`** for the two
   LLM recipes — doubly wrong: never read, *and* no longer the resolved default.
7. **The `MethodSpec` validator docstring claimed an unknown `score_type` is
   "left to the field's own validation".** There is none — `score_type` is plain
   `str`, so a typo'd family silently gets `0.5`. Documented honestly; tightening
   the annotation is a separate change.
8. `heuristic`'s constant harms `febrl_dedup` mildly too, not just `abt_buy`;
   "six front doors that **cut** at 0.5" — only four cut, two never read it.

**Second review round** (Codex re-reviewed `3c8e688`) found two more real code
defects, both fixed:

9. **A resume deadlock I had just created.** The parent clears only `worker_out`,
   never the worker's own `.partial`, and relaunched the worker *without*
   `--resume` — so the fresh worker hit the very "an earlier sweep was
   interrupted" guard added in fix #5 and refused, permanently. The guard meant
   to protect unsaved cells was blocking the retry that would have recovered
   them. Workers now always pass `--resume`.
10. **`--in-process` resume duplicated cells.** The subprocess path drops a
    benchmark's cells before re-measuring; the in-process path only appended, so
    an interrupted run re-added every finished cell on resume and compounded on
    each further resume. Both paths now share the benchmark-granular rule.

Plus three more prose claims contradicted by the generated tables: "on every
benchmark `0.5` leaves F1 in the low hundredths" (false — `dblp_acm` ≈ 0.22; the
*direction* is unanimous, the magnitude is not); "aggregation appears in exactly
one place" (it is two — the ship/transfer rules also take a median per-benchmark
mean; reframed as decision-statistic vs. reported-performance); and "wiring
`threshold=None` to the registry makes `VectorLLMCascade` use 0.7" (the
architecture never reads the registry — it calls `resolve_threshold(...,
"prob_llm")` directly, exactly as the CHANGELOG already said).

**Third review round** (Codex re-reviewed `d0d2b2a`) found three more, all fixed.
The first is the worst error in the PR — it was wrong about *langres*, not about
my data:

11. **`prob_fs` was described as unrunnable without labels. It is not.**
    `FellegiSunterMatcher.fit_unlabeled` (`core/matchers/fellegi_sunter.py:190`)
    does an unsupervised random-pair u-estimate plus EM. Only `prob_rf` truly
    requires labels. The exclusion stands but the *reason* changes: `prob_fs` /
    `prob_rf` scores come from a **per-dataset fit**, so their scale is
    re-estimated for your data and "one shared out-of-the-box constant" is not
    the question. Corrected in the generated rationale, `score_type.py` and the
    write-up.
12. **The worker's artifact was deleted before the parent committed its cells.**
    A parent crash (or a `checkpoint()` `OSError`) in that window lost an hour of
    compute outright, because the successful worker had already removed its own
    `.partial`. The unlink now happens only after `checkpoint()` returns.
13. **Resume validated the arguments but not the code.** Resuming after the
    harness, a matcher, a loader or the data changed kept the old cells, measured
    the rest with the new implementation, and pooled both into one artifact
    attributed to the current source — which the ship rule then selects from.
    `SweepReport` now carries a `source_fingerprint` (harness file hash + repo
    commit, `+dirty` marked) and resume refuses on mismatch. Two components
    because each alone has a blind spot: the file hash catches uncommitted edits
    a git SHA misses; the commit catches `langres` changes the file hash misses.
    `None` means *unknown* and is refused, never assumed to match — tested.

**Fourth review round** (Codex re-reviewed `1d466da`). Every new finding was a
hole in a *previous round's fix* — which is the reason each round is
re-requested rather than treated as self-evidently correct:

14. **The fingerprint hashed a dirty *bit*, not dirty content.** Two different
    uncommitted matcher edits at one commit both rendered `<HEAD>+dirty` with an
    unchanged harness hash, so the guard called them one revision and pooled the
    very cells it was added to separate. Now hashes `git diff HEAD` content. What
    it still cannot see (untracked files; `nogit` == `nogit`) is stated in the
    docstring rather than implied away.
15. **The worker's fingerprint was discarded.** `run_benchmark_isolated` returned
    only `.cells`, but each benchmark relaunches the current on-disk script — so
    an edit part-way through a multi-hour sweep made later workers run different
    code while the parent stamped everything with its startup fingerprint. It now
    reads the report whole and refuses a mismatch.
16. **The deferred-unlink fix contradicted itself.** Fix #12 stopped deleting the
    worker artifact until the parent had committed — then the resume path opened
    by unconditionally unlinking that same artifact before relaunching, discarding
    exactly the completed benchmark the deferral protected. A finished artifact
    with a matching fingerprint is now **adopted**; a stale one is announced and
    replaced.
17. Two prose sites still carried claims corrected elsewhere: `_select_constant`'s
    docstring still said it was the only cross-benchmark aggregation (the
    *executable* protocol contradicting the write-up), and
    `docs/TECHNICAL_OVERVIEW.md` still carried the false Fellegi-Sunter claim.
    The latter is the "grep the claim, not your phrasing" failure — I fixed it in
    three places in round three and missed a fourth, so this time I swept every
    phrasing across `src`, `docs`, `examples` and the CHANGELOG.

**Fifth and sixth review rounds.** Round five found two data-integrity bugs and
four overclaims; round six found that two of those fixes had *bounded* a
data-loss window without closing it — a fix that makes a bug rarer reads as done
and is not:

18. **Adopting a worker artifact checked only the code fingerprint.** Matching
    code is not matching *invocation*. Now validates `bootstrap_resamples`,
    embedder and the exact `(benchmark, method, seed)` set.
19. **Data destruction: old cells were dropped before the retry ran.** A failed
    worker left the filtered list in memory and the *next* benchmark's
    `checkpoint()` persisted the loss. Round six then showed the in-process path
    still truncated — yielding seed 0 and dying on seed 1 dropped a durable seed
    1. Now reconciled **per identity**: an old cell survives until its own
    replacement exists.
20. **A rejected worker artifact was deleted immediately.** Rejected for *this*
    invocation ≠ worthless; in the crash window it can be the only durable copy.
    Moved aside to `.rejected`, visible to `git status -uall`.
21. **"No free constant is safe" was a conclusion-level overclaim.** The sweep
    rejected *the median-F1 selector's* candidate; it never searched subject to
    the safety clause. Review found `t≈0.52` (also `0.51`, `0.54`) satisfies the
    stated aggregate criteria post hoc — recorded, and explicitly *not* treated
    as grounds to ship one.
22. **"Same portfolio" hid a scope change** — 54 cells / 9 benchmarks vs 60 / 10,
    the extra being `febrl_dedup`, the registry's only dedup benchmark.
23. **"`0.5` is beaten on every eligible benchmark" is false on e5** — 14 exact
    *tie* rows across `dblp_acm`, `fodors_zagat`, `walmart_amazon`,
    `wdc_computers`. Restated as **never worse**.
24. **The no-hand-typed-numbers guarantee was itself overclaimed** — twice. The
    ladder count is now generated, but prose still quotes `0.90`, `0.22`, the tie
    benchmarks and the intervals claim by hand. So the promise is **narrowed** to
    what is true: every table and the counts in them are generated and none is
    transcribed; quoted figures are hand-typed and checkable against a table in
    the same document.

Items 19 and 20 are pinned by tests **confirmed to fail against the previous
code** before being kept.

**Not actioned, with reasons:** changing the bootstrap scheme (#4) or the emitted
`score_type` tags (#3) — both are separate changes with their own blast radius,
and neither moves a verdict here. The tag defect is filed as **#254**; the
bootstrap limitation is argued in full in a PR comment rather than closed out,
because it is a disagreement I could be wrong about, not a task.

**Note on `Sourcery`:** it did **not** review this PR — it replied "your pull
request is larger than the review limit of 150000 diff characters". Its
`skipping` check status is not a pass.

## Gates

All run sequentially in one worktree (concurrent `uv run` races `.venv`/`.coverage`
and fabricates both failures and undercounted coverage):

| gate | result |
|---|---|
| `pytest -m "not integration and not slow and not finetune"` | **4836 passed, 2 skipped**, 121 deselected |
| contract coverage (CI's `CONTRACT_COVERAGE_INCLUDE`, floor 97.5) | **97.96%** |
| `ruff format --check .` | clean |
| `ruff check .` | All checks passed |
| `mypy src` | Success: no issues found in 216 source files |

Reproduce the study (all $0, offline):

```bash
OMP_NUM_THREADS=1 uv run --env-file .env python \
    examples/research/threshold_constant_sweep.py
uv run --env-file .env python tools/render_threshold_constant_writeup.py
```


fxd24:
@codex review

Fixes pushed at `3c8e688`. All three of your findings were verified against the code and acted on:

- **P2 RetrieveRerank** — confirmed and corrected. `retrieval.py:386` resolves `heuristic`, so nothing in `RetrieveRerank` moves. Fixed in CHANGELOG and the PR body.
- **P2 resume validation** — implemented. `main()` now refuses a resume whose checkpoint disagrees with the current `--resamples` or `--embedder`, with a negative-control test proving the guard is selective rather than a blanket refusal.
- **P1 bootstrap** — your description of the `min` attribution is correct and the docstring overclaimed. Now stated precisely: a *gold* pair has both endpoints in one cluster so the recall side is exact; a cross-cluster candidate pair is attributed to one endpoint, so the precision side under-counts co-movement and intervals are slightly too narrow. Disclosed rather than re-run — requiring both clusters to be drawn makes inclusion quadratic and measures a different estimand, and neither verdict is marginal (`sim_cos` ships on point estimates positive in every eligible cell; `heuristic`'s veto fails safe by keeping the incumbent). Happy to be argued out of that call.

A separate claim-audit also caught three factual errors in my prose, now fixed — notably that the ladder table **contradicts** the claim that deriving from labels beats any constant (the constant scores higher on 5 of 9 benchmarks), and that the `Retrieve` op stamps `heuristic`, not `sim_cos`.

fxd24:
@codex review

Second round pushed at `d0d2b2a`. Both new findings were real and are fixed:

- **P1 worker resume** — correct, and it was self-inflicted: the guard I added for your earlier P2 was the thing deadlocking the retry. Workers now always pass `--resume` (no-op with no partial; a mismatched partial now fails the metadata check with a message instead of pooling). Two tests pin the argv.
- **P2 in-process resume duplicates** — correct. The subprocess path already dropped the benchmark's cells before re-measuring; the in-process path only appended. Both now share the benchmark-granular rule.

All three doc findings were also correct and are fixed — notably `aggregation appears in exactly one place` (it is two: the ship and transfer rules also take a median per-benchmark mean) and the `VectorLLMCascade` wording, which contradicted the CHANGELOG's own correct statement that `MethodSpec.default_threshold` has no runtime reader.

**Re-reported but already addressed:** the bootstrap `min()` attribution (P1). I agree with your description and the docstring no longer overclaims — it now states that gold pairs are attributed exactly (both endpoints share a cluster) while cross-cluster candidate pairs are attributed to one endpoint, biasing intervals slightly too narrow. I did not change the scheme, deliberately: requiring both clusters to be drawn makes inclusion quadratic and measures a different estimand, and no verdict here is marginal — `sim_cos` ships on point estimates positive in every eligible cell, and `heuristic`'s veto fails safe by keeping the incumbent. If you think that reasoning is wrong I'd rather hear it than have it re-flagged.

fxd24:
@codex review

Third round pushed at `1d466da` — note the previous rounds' fixes were reviewed, but **these three are not**, which is why I'm re-requesting rather than assuming.

All three findings were correct:

- **P2 Fellegi-Sunter** — the most important one, because it was wrong about *langres* rather than about my data. `FellegiSunterMatcher.fit_unlabeled` (`fellegi_sunter.py:190`) really does an unsupervised u-estimate + EM, so a label-free user can run it; only `prob_rf` requires labels. The exclusion stands but the reason is now the per-dataset fit (the scale is re-estimated, so a shared constant isn't the question). Fixed in the generated rationale, `score_type.py` and the write-up.
- **P2 worker artifact durability** — correct, and it is the repo's own "commit before it can disappear" lesson. The unlink now happens only after the parent's `checkpoint()` returns.
- **P1 source revision** — implemented. `SweepReport.source_fingerprint` = harness file hash + repo commit (`+dirty` marked); resume refuses on mismatch. Two components because each alone has a blind spot, and `None` (an artifact predating the field) is refused rather than assumed to match — both tested.

**Re-reported but already fixed in earlier rounds** (verified still fixed at this SHA): the worker `--resume` deadlock, the in-process resume duplicates, and the resume metadata validation. If any of those still looks broken to you at `1d466da`, please quote the current line — I'd rather chase a real one than dismiss it as stale.

**Still deliberately unchanged:** the bootstrap `min()` attribution, disclosed with its exact scope (exact on recall, approximate on precision) and why no verdict turns on it.

fxd24:
@codex review

Fourth round pushed at `0bd691f`. Every finding this round was a hole in a **previous round's fix**, which is exactly why I keep re-requesting instead of assuming a fix is correct. All four were right:

- **P1 dirty bit vs dirty content** — correct and important: two different uncommitted edits at one commit collided on `<HEAD>+dirty`, so the guard would have pooled the cells it exists to separate. Now hashes `git diff HEAD` content. I've also written down what it still *cannot* see (untracked files; `nogit` == `nogit`) rather than implying completeness.
- **P1 worker fingerprint discarded** — correct. `run_benchmark_isolated` now reads the worker report whole and refuses a fingerprint mismatch, so a mid-sweep edit can't be published under the parent's startup identity.
- **P2 unlink-before-relaunch** — correct, and my two halves genuinely contradicted each other: I deferred the unlink to protect a completed benchmark, then deleted it on the resume path. A matching artifact is now **adopted** without re-measuring; a stale one is announced and replaced.
- **P2 `_select_constant` / P2 `TECHNICAL_OVERVIEW`** — both correct. The second was a "grep the claim, not your phrasing" miss: I fixed the Fellegi-Sunter claim in three places last round and missed a fourth, so this time I swept every phrasing of it across `src`, `docs`, `examples` and the CHANGELOG.

**Re-reported but fixed in earlier rounds** (please quote a current line if any still looks broken at this SHA): the worker `--resume` deadlock, in-process resume duplicates, and resume metadata validation.

**Still deliberately unchanged:** the bootstrap `min()` attribution — disclosed with its exact scope (exact on the recall side, approximate on precision) and the argument that no verdict turns on it. If that argument is wrong I'd genuinely rather hear the counter-argument than have the finding repeated.

fxd24:
### Review status at `0bd691f` — converged

The fifth Codex pass returned **five findings and no new ones** — the first round where nothing was introduced by the previous round's fixes. Since they render as open threads, here is each one checked against the current code rather than dismissed as stale:

| finding | status at `0bd691f` |
|---|---|
| Validate checkpoint metadata before resuming | **Fixed** — `bootstrap_resamples` guard at :1534, `--embedder` guard at :1541 |
| Refuse checkpoints from a different source revision | **Fixed** — `source_fingerprint` guard at :1549 |
| Resume worker partials before relaunching | **Fixed** — worker argv carries `--resume` (:1302); a completed artifact is adopted rather than deleted |
| Skip completed cells when resuming in-process | **Fixed** — benchmark-granular filter at :1606, matching the subprocess path at :1619 |
| Preserve both endpoint clusters in the bootstrap | **Deliberately unchanged, disclosed** — see below |

**On the bootstrap (the one standing disagreement).** The critique is correct as stated and the docstring no longer overclaims: a *gold* pair has both endpoints in one cluster, so the recall side is attributed exactly; a cross-cluster candidate pair goes to one endpoint, so the precision side under-counts co-movement and intervals are biased slightly **too narrow**.

I did not change the scheme, for two reasons stated so they can be argued with:

1. **The alternative changes the estimand.** Requiring both endpoint clusters to be drawn makes a pair's inclusion probability quadratic, shrinking the effective sample and measuring a different quantity — not a strictly better one.
2. **No verdict here is marginal, and the one that could be sensitive fails safe.** `sim_cos` ships on point estimates that are positive on *every* eligible cell (+0.05 to +0.87); no interval width turns a positive point estimate negative. `heuristic`'s significant-harm veto is the direction a too-narrow interval could distort — but its `abt_buy` deltas sit several interval half-widths below zero on three independent seeds, and the veto's effect is to **keep the incumbent**, so an error there is conservative.

If either argument is wrong, that is worth a re-run and I'd rather be told. Otherwise this is a disclosed limitation, not an unaddressed defect.

**CI is green on every non-skipped check.** Note `Sourcery review` reports `skipping` but actually replied *"your pull request is larger than the review limit of 150000 diff characters"* — it never reviewed this PR, so its status is not a pass.

fxd24:
@codex review

Fifth round pushed at `33c640b`. I had wrongly reported the previous round as clean — I queried the comments before this batch landed. All six findings were valid.

**Code (both were holes in earlier fixes in this same PR):**
- **P1 adopt validation** — correct. Matching *code* is not matching *invocation*. Now validates `bootstrap_resamples`, embedder and the exact `(benchmark, method, seed)` set, and reports every reason it refused. Six tests, including a negative control that fails if adoption stops happening at all.
- **P1 preserve old cells** — correct, and this was real data destruction: the filter ran before the worker, so a failed worker left the filtered list in memory and the *next* benchmark's `checkpoint()` persisted the loss. Subprocess path now stages and swaps atomically; in-process keeps per-cell checkpointing (it is what writes the worker's own `.partial`) but checkpoints `kept + fresh`, so it can neither duplicate nor pre-emptively delete. **I verified the regression test fails against the old ordering before keeping it.**

**Docs — three of four contradicted the study's own generated tables:**
- **"no free constant that is safe"** — you were right that this doesn't follow. The sweep rejected *the median-F1 selector's* candidate; it never searched subject to the safety clause. I've recorded your `t≈0.52`/`0.51`/`0.54` finding explicitly, and equally explicitly that post-hoc inspection of the artifact is not grounds to ship one.
- **"same portfolio"** — corrected: 54 cells / 9 benchmarks vs 60 / 10, the extra being `febrl_dedup`, the registry's only dedup benchmark.
- **ties** — confirmed 14 exact-tie rows on e5 across `dblp_acm`, `fodors_zagat`, `walmart_amazon`, `wdc_computers`. Restated as **never worse**, which is what the decision actually rests on.
- **the no-hand-typed-numbers guarantee** — the irony is fair and I've fixed it the way this repo learned to: the **ladder generator now emits the count**, prose points at it. Generated line reads 5 of 9 for both families, matching the hand count it replaced.

**Still deliberately unchanged:** the bootstrap `min()` attribution, disclosed with its exact scope and the argument that no verdict turns on it.

Coverage, measured on the authoritative full suite (not the PR gate): **9225 stmts / 100 missed / 98.29%**, identical to main — this PR is coverage-neutral at +14 statements. Tag defect tracked as #254.

fxd24:
@codex review

Sixth round pushed at `5ee278f`. All three new findings were valid, and two of them are refinements of **my own round-five fixes** — both had bounded a data-loss window without closing it, which is the pattern worth naming: a fix that makes a bug rarer reads as done and is not.

- **P1 in-process retry truncation** — correct. Yielding seed 0 and dying on seed 1 dropped a seed-1 result that was durable and never re-measured. Now reconciled **per `(benchmark, method, seed)`**: an old cell survives until its own replacement exists.
- **P1 rejected artifact deleted** — correct, and the framing is right: rejected for *this* invocation does not mean worthless. Moved aside to `<worker_out>.rejected` rather than unlinked, so it stays visible to `git status -uall` and an operator decides instead of a crash deciding.
- **P2 the guarantee is still false** — also correct, and a better catch than the first version. Generating the ladder count did not rescue it, because the prose still quotes `0.90`, `dblp_acm`'s `0.22`, the four e5 tie benchmarks, and the all-intervals-above-zero claim by hand. Rather than chase generation further I **narrowed the promise** to what is actually true: every table and the counts inside them are generated and none is transcribed; quoted figures are hand-typed, can drift, and are checkable against a table in the same document.

Both P1 fixes are pinned by tests I **confirmed fail against the previous code** before keeping them.

**Still deliberately unchanged:** the bootstrap `min()` attribution — disclosed with its exact scope (exact on recall, approximate on precision) and the argument that no verdict turns on it, since `sim_cos` ships on point estimates positive in every eligible cell and `heuristic`'s veto fails safe by keeping the incumbent.

Gates: fast suite 4836 passed / 2 skipped; contract coverage 97.96% (per-PR floor 97.5) and **98.29% on the authoritative full suite** (floor 98) — identical to main, so this PR is coverage-neutral at +14 statements.

fxd24:
@codex review

Seventh round pushed at `879706f`. All three valid.

- **P1 `.rejected` collision** — correct, and this is the third consecutive round where a data-safety fix of mine left a narrower version of the same hole. Worth naming rather than glossing: each fix made the loss rarer and therefore read as done. Now allocates the first free `.rejected` / `.rejected.N` slot; nothing under that name is ever overwritten.
- **P2 `.writing` staging file** — correct, and the direction matters: `.writing` exists only when a checkpoint was fully staged and crashed before `os.replace()`, so it holds **more** cells than `.partial`, not fewer. Resume now refuses and asks the operator to compare rather than guessing which file is authoritative — the wrong guess is the data loss.
- **P2 duplicate `--only`** — correct; deduplicated. It measured an expensive benchmark twice, kept one copy of its cells, counted both toward `expected_cells`, and refused to publish at the very end after paying for the work.

Each is pinned by a test.

Gates at this SHA: fast suite **4839 passed / 2 skipped**; contract coverage **97.96%** (per-PR floor 97.5); ruff format/check clean; mypy clean on 216 files. The authoritative full-suite contract figure measured earlier this session was **9225 stmts / 100 missed / 98.29%** against floor 98 — identical to main.

**Still deliberately unchanged:** the bootstrap `min()` attribution, disclosed with its exact scope and the argument that no verdict turns on it.

PR #255, opened 2026-07-29
results(lfm25): LiquidAI LFM2.5 encoders on ER candidate blocking

fxd24:
## What this is

Measures LiquidAI's LFM2.5 encoders on ER candidate blocking, using the existing
`examples/research/embedder_ladder.py` harness. Two **separate** artifacts, because
they answer different questions and must not share a ranking table:

- **Study A (like-for-like):** `LFM2.5-Embedding-350M` vs `intfloat/e5-base-v2`
  (the shipped default and this study's baseline), `BAAI/bge-base-en-v1.5`,
  `all-MiniLM-L6-v2`.
- **Study B (NOT like-for-like):** `LFM2.5-Encoder-350M` / `-230M` are base
  masked-LM checkpoints that ship no pooling config, so sentence-transformers
  attaches an **untrained mean-pooling head**. Plus a **random-init control**.

The write-up is generated by `examples/research/lfm25_report.py` from the committed
rows. No measured number is typed into prose, and the generator is verified to
reproduce its own output byte-for-byte.

## Headline

`LiquidAI/LFM2.5-Embedding-350M` vs. `intfloat/e5-base-v2` — langres's current `DEFAULT_EMBEDDING_MODEL` and therefore the baseline every interval below is measured against.

- Ahead with an interval excluding zero on: **no benchmark/arm**.
- Behind with an interval excluding zero on: `abt_buy` (instruct), `abt_buy` (none), `amazon_google` (instruct), `amazon_google` (none), `wdc_computers` (instruct), `wdc_computers` (none).

**These are per-ARM verdicts, not a checkpoint-level one**, and the distinction changes the reading. A paired interval needs both models' per-record vectors on the same records, so an arm the baseline never runs cannot be tested at all — it is silently absent from the two lines above rather than counted as a loss:

- `abt_buy` arm `documented`: **0.9741** vs the baseline's best **0.9703** — **above the baseline**, and untested.
- `amazon_google` arm `documented`: **0.8187** vs the baseline's best **0.8338** — below the baseline, and untested.
- `fodors_zagat` arm `documented`: **1.0000** vs the baseline's best **1.0000** — level with the baseline, and untested.
- `walmart_amazon` arm `documented`: **0.8764** vs the baseline's best **0.8810** — below the baseline, and untested.
- `wdc_computers` arm `documented`: **0.7084** vs the baseline's best **0.7363** — below the baseline, and untested.

So on any benchmark listed above, the honest statement is *"behind on the arms that could be tested"*, not *"behind"*. The untested arms are the checkpoint's own documented prompts, which is where a vendor would expect it to look best; closing that gap needs the baseline re-run under the same prompts, which this study did not do.

**It cannot become langres's default regardless of how it scores** — see the licence section. That is a legal constraint, not a measurement.

Mean **per-record** recall difference, bootstrap-resampled **by gold cluster** (`langres.experiments.statistics.paired_entity_bootstrap`) — never by pair rows, which are dependent inside one entity and give intervals that are far too tight. The point estimate is the statistic the interval bounds, not a difference of aggregate recalls.

| model | benchmark | arm | Δ recall vs baseline | 95% CI | clusters | excludes 0 |
|---|---|---|---|---|---|---|
| `all-MiniLM-L6-v2` | `abt_buy` | `instruct` | -0.0397 | [-0.0531, -0.0279] | 1012 | yes |
| `all-MiniLM-L6-v2` | `abt_buy` | `none` | -0.0324 | [-0.0442, -0.0206] | 1012 | yes |
| `all-MiniLM-L6-v2` | `amazon_google` | `instruct` | -0.0254 | [-0.0360, -0.0157] | 995 | yes |
| `all-MiniLM-L6-v2` | `amazon_google` | `none` | -0.0056 | [-0.0120, +0.0000] | 995 | no |
| `all-MiniLM-L6-v2` | `fodors_zagat` | `instruct` | 0.0000 | [+0.0000, +0.0000] | 112 | no (bound **exactly** 0) |
| `all-MiniLM-L6-v2` | `fodors_zagat` | `none` | 0.0000 | [+0.0000, +0.0000] | 112 | no (bound **exactly** 0) |
| `all-MiniLM-L6-v2` | `walmart_amazon` | `instruct` | -0.0242 | [-0.0360, -0.0150] | 846 | yes |
| `all-MiniLM-L6-v2` | `walmart_amazon` | `none` | -0.0159 | [-0.0253, -0.0088] | 846 | yes |
| `all-MiniLM-L6-v2` | `wdc_computers` | `instruct` | -0.1163 | [-0.1417, -0.0902] | 877 | yes |
| `all-MiniLM-L6-v2` | `wdc_computers` | `none` | -0.1528 | [-0.1796, -0.1242] | 877 | yes |
| `BAAI/bge-base-en-v1.5` | `abt_buy` | `instruct` | 0.0039 | [-0.0020, +0.0098] | 1012 | no |
| `BAAI/bge-base-en-v1.5` | `abt_buy` | `none` | 0.0044 | [-0.0029, +0.0123] | 1012 | no |
| `BAAI/bge-base-en-v1.5` | `amazon_google` | `instruct` | -0.0161 | [-0.0234, -0.0096] | 995 | yes |
| `BAAI/bge-base-en-v1.5` | `amazon_google` | `none` | -0.0023 | [-0.0077, +0.0021] | 995 | no |
| `BAAI/bge-base-en-v1.5` | `fodors_zagat` | `instruct` | 0.0000 | [+0.0000, +0.0000] | 112 | no (bound **exactly** 0) |
| `BAAI/bge-base-en-v1.5` | `fodors_zagat` | `none` | 0.0000 | [+0.0000, +0.0000] | 112 | no (bound **exactly** 0) |
| `BAAI/bge-base-en-v1.5` | `walmart_amazon` | `instruct` | -0.0055 | [-0.0110, -0.0011] | 846 | yes |
| `BAAI/bge-base-en-v1.5` | `walmart_amazon` | `none` | -0.0061 | [-0.0120, -0.0011] | 846 | yes |
| `BAAI/bge-base-en-v1.5` | `wdc_computers` | `instruct` | -0.0778 | [-0.1008, -0.0526] | 877 | yes |
| `BAAI/bge-base-en-v1.5` | `wdc_computers` | `none` | -0.1215 | [-0.1461, -0.0972] | 877 | yes |
| `LiquidAI/LFM2.5-Embedding-350M` | `abt_buy` | `instruct` | -0.0235 | [-0.0343, -0.0138] | 1012 | yes |
| `LiquidAI/LFM2.5-Embedding-350M` | `abt_buy` | `none` | -0.0265 | [-0.0383, -0.0157] | 1012 | yes |
| `LiquidAI/LFM2.5-Embedding-350M` | `amazon_google` | `instruct` | -0.0128 | [-0.0208, -0.0059] | 995 | yes |
| `LiquidAI/LFM2.5-Embedding-350M` | `amazon_google` | `none` | -0.0442 | [-0.0568, -0.0323] | 995 | yes |
| `LiquidAI/LFM2.5-Embedding-350M` | `fodors_zagat` | `instruct` | 0.0000 | [+0.0000, +0.0000] | 112 | no (bound **exactly** 0) |
| `LiquidAI/LFM2.5-Embedding-350M` | `fodors_zagat` | `none` | 0.0000 | [+0.0000, +0.0000] | 112 | no (bound **exactly** 0) |
| `LiquidAI/LFM2.5-Embedding-350M` | `walmart_amazon` | `instruct` | -0.0037 | [-0.0089, +0.0011] | 846 | no |
| `LiquidAI/LFM2.5-Embedding-350M` | `walmart_amazon` | `none` | -0.0015 | [-0.0055, +0.0022] | 846 | no |
| `LiquidAI/LFM2.5-Embedding-350M` | `wdc_computers` | `instruct` | -0.0363 | [-0.0582, -0.0139] | 877 | yes |
| `LiquidAI/LFM2.5-Embedding-350M` | `wdc_computers` | `none` | -0.0345 | [-0.0552, -0.0139] | 877 | yes |

## Study B — base encoders (untrained pooling; not comparable to study A)

| model | benchmark | arm | Δ recall vs baseline | 95% CI | clusters | excludes 0 |
|---|---|---|---|---|---|---|
| `LiquidAI/LFM2.5-Encoder-230M` | `abt_buy` | `instruct` | -0.4426 | [-0.4728, -0.4105] | 1012 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `abt_buy` | `none` | -0.5039 | [-0.5337, -0.4760] | 1012 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `amazon_google` | `instruct` | -0.2992 | [-0.3259, -0.2737] | 995 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `amazon_google` | `none` | -0.2105 | [-0.2341, -0.1856] | 995 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `fodors_zagat` | `instruct` | -0.1161 | [-0.1786, -0.0625] | 112 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `fodors_zagat` | `none` | -0.0714 | [-0.1163, -0.0268] | 112 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `walmart_amazon` | `instruct` | -0.1412 | [-0.1613, -0.1165] | 846 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `walmart_amazon` | `none` | -0.1143 | [-0.1342, -0.0952] | 846 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `wdc_computers` | `instruct` | -0.2746 | [-0.3071, -0.2440] | 877 | yes |
| `LiquidAI/LFM2.5-Encoder-230M` | `wdc_computers` | `none` | -0.3573 | [-0.3907, -0.3268] | 877 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `abt_buy` | `instruct` | -0.4309 | [-0.4636, -0.3988] | 1012 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `abt_buy` | `none` | -0.6235 | [-0.6534, -0.5965] | 1012 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `amazon_google` | `instruct` | -0.5059 | [-0.5355, -0.4763] | 995 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `amazon_google` | `none` | -0.2518 | [-0.2778, -0.2243] | 995 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `fodors_zagat` | `instruct` | -0.1518 | [-0.2232, -0.0893] | 112 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `fodors_zagat` | `none` | -0.0536 | [-0.0982, -0.0179] | 112 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `walmart_amazon` | `instruct` | -0.2439 | [-0.2722, -0.2154] | 846 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `walmart_amazon` | `none` | -0.1593 | [-0.1824, -0.1352] | 846 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `wdc_computers` | `instruct` | -0.3215 | [-0.3553, -0.2886] | 877 | yes |
| `LiquidAI/LFM2.5-Encoder-350M` | `wdc_computers` | `none` | -0.3192 | [-0.3522, -0.2892] | 877 | yes |
| `random-init-control-350M` | `abt_buy` | `instruct` | -0.0426 | [-0.0620, -0.0245] | 1012 | yes |
| `random-init-control-350M` | `abt_buy` | `none` | -0.0299 | [-0.0469, -0.0128] | 1012 | yes |
| `random-init-control-350M` | `amazon_google` | `instruct` | -0.1243 | [-0.1445, -0.1047] | 995 | yes |
| `random-init-control-350M` | `amazon_google` | `none` | -0.1030 | [-0.1227, -0.0836] | 995 | yes |
| `random-init-control-350M` | `fodors_zagat` | `instruct` | 0.0000 | [+0.0000, +0.0000] | 112 | no (bound **exactly** 0) |
| `random-init-control-350M` | `fodors_zagat` | `none` | 0.0000 | [+0.0000, +0.0000] | 112 | no (bound **exactly** 0) |
| `random-init-control-350M` | `walmart_amazon` | `instruct` | -0.0171 | [-0.0275, -0.0072] | 846 | yes |
| `random-init-control-350M` | `walmart_amazon` | `none` | -0.0194 | [-0.0295, -0.0099] | 846 | yes |
| `random-init-control-350M` | `wdc_computers` | `instruct` | -0.1320 | [-0.1584, -0.1032] | 877 | yes |
| `random-init-control-350M` | `wdc_computers` | `none` | -0.1218 | [-0.1488, -0.0978] | 877 | yes |

## Noise floor

**A named finding, not a footnote.** `random-init-control-350M` is the LFM2.5 350M architecture with **seeded random weights and no training whatsoever**, run through the identical pipeline. A random transformer is a random feature map: near-duplicate strings still land near each other, so it retrieves real candidates. Its recall is therefore what a benchmark hands a model that knows *nothing*, and the gap above it is the benchmark's entire usable dynamic range.

| benchmark | random-init control | best tuned model | range | control's paired CI | verdict |
|---|---|---|---|---|---|
| `abt_buy` | 0.9109 | 0.9741 (`LiquidAI/LFM2.5-Embedding-350M`) | **+0.0632** | [-0.0469, -0.0128] | separates, wide |
| `amazon_google` | 0.7058 | 0.8338 (`intfloat/e5-base-v2`) | **+0.1281** | [-0.1445, -0.1047] | separates, wide |
| `fodors_zagat` | 1.0000 | 1.0000 (`intfloat/e5-base-v2`) | **+0.0000** | [+0.0000, +0.0000] | **cannot separate trained from random** |
| `walmart_amazon` | 0.8626 | 0.8810 (`intfloat/e5-base-v2`) | **+0.0183** | [-0.0275, -0.0072] | separates, but narrow (<0.05) |
| `wdc_computers` | 0.5788 | 0.7363 (`intfloat/e5-base-v2`) | **+0.1575** | [-0.1488, -0.0978] | separates, wide |

**Benchmarks that cannot tell a trained retriever from random weights: `fodors_zagat`.** The control's paired interval there does not exclude zero, so a model ranking read off it is not evidence about any embedder.

**Benchmarks that do separate the two, but narrowly (&lt;0.05 of range): `walmart_amazon`.** This is a statement about **resolution, not significance**. The control *is* significantly below the tuned models here — the separation is real. But the whole trained-vs-random range is small, so a reported margin can be a large fraction of it, and a ranking built from differences of that size is fragile rather than meaningless.

Benchmarks that separate the two with room to spare: `abt_buy`, `amazon_google`, `wdc_computers`.

**A correction, stated plainly because it was published the other way round for one revision of this document.** An earlier version called a benchmark *uninformative* whenever the tuned-vs-random gap was under five recall points, and on that basis said `walmart_amazon` could not distinguish a trained retriever from noise. That was wrong, and this study's own rows say so: the control's paired intervals on `walmart_amazon` are `[-0.0275, -0.0072]` and `[-0.0295, -0.0099]`, both excluding zero. The gap between one seeded control and the best tuned score is a point estimate, not a variance estimate, and it cannot be used as a significance test — least of all against measured intervals sitting in the same file. Only `fodors_zagat` genuinely fails to separate.

**The control also outscores the real base encoders.** All three wear the same untrained mean-pooling head, so pooling is held fixed throughout; whether *pretraining* is the only remaining difference depends on the backbone, and is stated per row:

- Random weights match or beat `LiquidAI/LFM2.5-Encoder-230M` on **5 of 5** benchmarks (`abt_buy`, `amazon_google`, `fodors_zagat`, `walmart_amazon`, `wdc_computers`) — *different backbone and size*, so this pair confounds pretraining with model size and is an observation only.
- Random weights match or beat `LiquidAI/LFM2.5-Encoder-350M` on **5 of 5** benchmarks (`abt_buy`, `amazon_google`, `fodors_zagat`, `walmart_amazon`, `wdc_computers`) — **matched backbone**, so this pair isolates pretraining.

**What this supports, and what it does not.** On `LiquidAI/LFM2.5-Encoder-350M` vs the control, architecture and pooling are held fixed and only pretrained-vs-random weights differ. So the *finding* is exactly this: **under this untrained mean-pooling configuration, the pretrained checkpoint scores worse than random weights.** The 230M row points the same way but cannot even support that, because it varies backbone and size as well as pretraining.

A tempting explanation is that MLM training shapes token representations for a head this pipeline never attaches, so averaging them destroys more than averaging random projections does. That is a **hypothesis, not a result** — this experiment holds pooling *fixed*, so it cannot attribute the gap to pooling rather than to the checkpoint. Testing it means varying the pooling over the same features (CLS, last-token, a trained head) and seeing whether the ordering reverses. Not done here. What is safe to conclude either way is narrower and sufficient: an untuned encoder must not be ranked against a retrieval-tuned one.

This reframes `fodors_zagat` specifically. It was already labelled *saturated* — every usable embedder scoring near the ceiling. The control shows it is stronger than that: it is **uninformative in the strict sense**, because an untrained network also scores near the ceiling. Saturation says the models agree; this says the benchmark cannot tell a trained retriever from noise. It must never be cited as evidence that one embedder beats another.

**Recalibrating the earlier portfolio study.** `docs/research/20260727_embedder_ladder.md` reports per-benchmark margins that predate any noise-floor control, so it has no way to say which of them clear noise. Read against this control:

- On `fodors_zagat`, the best tuned model does not beat random weights at all (range +0.0000) — any ranking read off it is uninterpretable.
- On `walmart_amazon`, the **entire** distance between the best tuned model and a network with random weights is **0.0183**. A margin is best read as a fraction of that, not in isolation.

Concretely, every `walmart_amazon` margin that study reports as a win, with its interval, set against the 0.0183 total trained-vs-random range measured here:

- `BAAI/bge-base-en-v1.5`: **+0.0098** [+0.0020, +0.0187]  ← 54% of the benchmark's entire range
- `BAAI/bge-small-en-v1.5`: **+0.0118** [+0.0033, +0.0213]  ← 64% of the benchmark's entire range
- `google/embeddinggemma-300m`: **+0.0092** [+0.0004, +0.0189]  ← 50% of the benchmark's entire range
- `intfloat/e5-base-v2`: **+0.0159** [+0.0078, +0.0246]  ← 87% of the benchmark's entire range

**4 of 4** are smaller than the whole range. Those intervals exclude zero, so they are **real effects and this study does not call them noise** — `walmart_amazon` does separate trained models from random weights. The narrower point is about resolution: on a benchmark whose entire range is 0.0183, a margin of that order is a large share of everything there is to measure, so it is a fragile basis for ranking and should be quoted with the range beside it rather than on its own.

This study does not refute those margins and does not call them noise — different models, measured intervals that exclude zero, and nothing here contradicts them. The recalibration is about **resolution**: quote a `walmart_amazon` margin with the trained-vs-random range beside it, because a margin that is a large fraction of everything the benchmark can measure is a fragile basis for ranking even when it is real. The earlier document is left as it stands; this is a recalibration note, not a rewrite.

The control also exists because this harness *shipped* the bug it now guards against: a substitution that silently left random weights running scored 0.9911 recall and 0.9971 AUC on `fodors_zagat`, and nothing in the row looked wrong. A permanent noise-floor arm is how that stays visible instead of being rediscovered.

## Licence

All three repos ship a byte-identical LFM Open License v1.0 (sha256
`4d28ca14…`), committed here as `docs/research/20260729_lfm25_license.txt` and
quoted from that file rather than from the model-card summary. §5 conditions
Commercial Use on not exceeding a **$10M annual-revenue threshold**. That is not
an OSI licence, so the model may only ever be an **opt-in named model with a
licence note** and **must not** become `DEFAULT_EMBEDDING_MODEL` — langres is
Apache-2.0. This is a legal constraint independent of how it scores.

## Statistics

Paired bootstrap resampled **by gold cluster**. Intervals are **nominal,
per-benchmark and uncorrected**; the doc states how many nominal hits are
expected by chance and does **not** claim family-wise significance. Bounds
landing exactly on `0.0000` are flagged as *not* excluding zero.

## Also in this PR

- `fix(ladder)`: the memory guard watched the swap *derivative*, and stayed
  silent while the OS killed the sweep mid-cell at 19892/20480M swap. It now
  guards on **available memory**, checked before each cell as well as after.
- `feat(lfm25)`: `LFM25_STUDY` selector — the harness has no skip-completed
  logic, so re-running a finished study is a re-do that churns correct results.
- Determinism fix in the report generator (set iteration + `max` first-maximum
  meant a tie credited an arbitrary model); regression test verified to fail
  without the fix.

## ⚠️ A published claim in this PR was wrong and is retracted

An earlier revision of this branch called a benchmark unable to tell a trained
retriever from random weights whenever the tuned-vs-random **gap** was under five
recall points, and on that basis said `walmart_amazon` could not support an
embedder claim. **That was wrong** — a point estimate used as a variance estimate,
contradicted by measured intervals sitting in the same rows: the control's paired
intervals on `walmart_amazon` are `[-0.0275, -0.0072]` and `[-0.0295, -0.0099]`,
**both excluding zero**. Only `fodors_zagat` genuinely fails to separate.

The verdict now comes from the control's paired **interval**, with range reported
beside it as a separate, weaker statement about **resolution**. The recalibration
of `docs/research/20260727_embedder_ladder.md` is in the generated Noise floor
section above — including its four `walmart_amazon` margins and EmbeddingGemma's
`+0.0092` at ~50% of the benchmark's entire trained-vs-random range. Those are
**real effects, not noise**; the caveat is that a margin that large a share of
everything a benchmark can measure is fragile as a ranking and should be quoted
with the range beside it. That study is **not** rewritten.

This section is deliberately short: the numbers live in the generated section
above. It used to restate them by hand, and the hand-typed copy is the one that
carried the error.

## Gates

`4802 passed, 2 skipped`, 96% total; `ruff format --check`, `ruff check`, `mypy src` all clean.
All four generated docs verified to re-render byte-identical from the committed rows.

## Review status — read this before treating the PR as reviewed

**Sourcery did not review this PR.** It replied *"your pull request is larger than
the review limit of 150000 diff characters"* to both the automatic run and an
explicit `@sourcery-ai review`. Its check shows `skipping`, which is **not** a
clean bill of health — the diff is dominated by the committed rows JSONLs and the
411 KB reference-recall sidecar. Only **Codex** actually reviewed.

Codex has reviewed **four** rounds, raising 22 findings. Every one was verified
against the code, all were real, and all are fixed. Several were defects in the
published artifact itself — an empty reachable-recall table, a causal claim that
outran its matched-backbone pair, an `lfm1.0` model listed under "the OSI-licensed
field", reproduce commands pointing at another study's artifacts, a headline that
claimed a checkpoint-level verdict the paired test never made, and the
noise-floor misclassification retracted above.

**A process failure worth recording, because it nearly shipped the error.** After
round 3 I reported "all findings fixed, round clean". It was not: a review had
landed against the then-current head with six further findings, three of them P1,
and I never read it — I had filtered the comment query on a timestamp *later* than
the review, so it could not return anything, and I read the empty result as
absence of findings. A round is not clean until its body has been read. Waiters
are now keyed on the head **commit SHA**, not on a time window.

### Verified rather than asserted, this round

- The provenance stability check **fires** — a canary file under `src/langres`
  flips `verified_unchanged_during_run` to false and exits non-zero. It had never
  been observed failing, which made it a hypothesis rather than a safety net.
- The new narrow-benchmark regression test **discriminates**: the old predicate
  reproduces the published error on the same fixture; the new one does not.
- The `*.py` tree digest was certifying *this sweep* as unchanged while three
  commits rewrote the drivers mid-run. Drivers are now digested separately and
  the churn is disclosed in the report.
- The README claim-check used to compare the README against itself, so it could
  not fail. It now cross-checks every claim against the generated artifact.


fxd24:
@sourcery-ai review

fxd24:
@codex review — six findings from the previous round were addressed in 596df67. Please re-check the fix commit, in particular the failure-scoping change in run_ladder.sh and the restricted causal claim in lfm25_report.py.

fxd24:
@codex review — round-2 findings addressed in 6be1b0c. The provenance one was the important catch: the section was deriving blob hashes from HEAD, so after the previous fix commit it would have attributed every measured row to code that never ran. Provenance is now persisted at measurement time in a committed sidecar.

fxd24:
@codex review — round-3 findings addressed in dbcbb29. Notably: the headline is now arm-qualified (the documented-prompt arm has no counterpart in the baseline, so it was untestable rather than a loss, and on abt_buy it scores above e5's best); `set -u` + unassigned FAILED_BENCHMARKS was aborting run_ladder.sh before its commit block; and provenance is now captured by a committed writer on the measurement path rather than existing only as an untracked script.

fxd24:
@codex review

fxd24:
@codex review

fxd24:
@codex review

fxd24:
@codex review

fxd24:
@codex review

fxd24:
@codex review

fxd24:
@codex review

fxd24:
@codex review

Round-11 fixes pushed as `4a88442`. All five findings addressed:

1. **Pretraining conclusion gated on surviving rows.** `_pretraining_section()` now builds the claim from `_control_vs_base_comparisons()`: no comparable cells → it states that and concludes nothing; matched-backbone pair missing → the causal sentence is withheld; control loses → the opposite verdict is printed with the count. Proven to be able to fail: the same dead-control rows rendered through the previous commit still carry "scores worse than random weights", through this one they do not.
2. **The `fodors_zagat` reframing** names one benchmark, so it is emitted only while the measured intervals put it in `uninformative`.
3. **run_lfm25.sh** routes a failed render through `abort_with_provenance`, closing and committing the window.
4. **write_provenance.py `--partial`** records `window_complete`; the report forces partial wording instead of reading `studies_measured` as intent.
5. **run_ladder.sh `FAILED_BENCHMARKS`** stores `benchmark:code` so each cell keeps its own exit status.

Gates on this SHA: 4820 passed / 2 skipped (full non-integration suite), `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on both drivers, render idempotent, and the README-vs-artifact claim check re-verified.

fxd24:
@codex review

Round-12 fixes pushed as `fd761f0`. All seven findings addressed; both P1s were reproduced against the previous commit before being fixed, so neither guard is untested.

**P1 — non-granular refusal destroyed rows.** The granular loop excluded exit 2/3 from `FAILED_BENCHMARKS`; the non-granular branch did not, so the flush called `record_process_failure` and committed the deletion while the refusal branch printed "NOT recording a failure row". `tmp/probe_nongranular_gate.sh` runs the same refusal through both revisions: the old driver mutates the rows file, this one leaves it byte-identical.

**P1 — `--start` ran before the PID wait.** Moved the wait into `run_lfm25.sh` ahead of `--start` (still passed down, so standalone `run_ladder.sh` is unchanged). `tmp/probe_wait_before_start.sh`: the old driver runs `uv` while the prior sweep is alive, this one does not, and still proceeds once the pid exits.

**P2 — re-finishing a complete window.** Split out `commit_provenance()` for the already-closed paths (both the render failure *and* the `--finish`-rejected path had this), and `write_provenance.finish()` now refuses to relabel a complete window as partial, before writing anything.

**P2 — `_cross_study_caveat` ignored `window_complete`.** Now warns on an aborted `LFM25_STUDY=both` run.

**P2 — crossed metric revisions.** Now requires one revision across both files rather than two equal sets.

**P2 — ties counted as wins.** `strict` (`>=` everywhere and `>` somewhere) carries the directional claims; the bullets keep "matches or beats". On the committed rows the finding is unchanged — 5 of 5, strictly ahead.

**P2 — inverted vs blind.** `_noise_floor_table` returns `inverted` separately; the blind list, the retraction and the `fodors_zagat` reframing key on `blind`, and inverted benchmarks get their own paragraph.

Gates on this SHA: 4831 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on both drivers, render idempotent, README-vs-artifact claim check re-verified.

fxd24:
@codex review

Round-13 fixes pushed as `ff00fee`. All five findings addressed; the three code fixes were each reproduced against the previous commit before being fixed.

**P1 — the environment was invisible to every hash.** `pyproject.toml` now joins the fatal blobs; `uv.lock` goes into a new `environment_blobs` and is disclosed non-fatally on the same terms as the drivers (it is gitignored, and `uv run` rewrites it unprompted — aborting a multi-hour paid sweep on that would kill work that would have succeeded). A window opened before the field existed reports `null`: *not observed*, which is neither "unchanged" nor "changed". The committed sidecar says exactly that in the rendered report rather than claiming a stability it could not have measured.

**P2 — any re-finish is now refused.** You are right that the guard was one-directional: a plain `--finish` over a partial sidecar flipped `window_complete` to true while keeping the `partial_note`. Keyed on `measurement_window.finished` now, so both directions are refused before anything is written. `tmp/probe_round13_gates.py` shows the previous commit producing exactly that self-contradicting file.

**P2 — non-granular lost the child status.** Now stores `benchmark:code` like the granular path. `tmp/probe_nongranular_code.sh` (two models so the memory guard actually runs, `MEM_FLOOR_MIB` forced to trip): HEAD blames 9, this tree blames 137.

**P2 — the wrong rows file in both study headers.** Derived from `ARTIFACT_PREFIX`, like the reproduce commands already were. All three ladder reports re-rendered.

**P2 — the index drifting from the artifact.** The cross-check existed only as a gitignored `tmp/` script — a gate that dies with the worktree, which is worse than none because it had been run and believed. It is `tests/examples/test_lfm25_index_entry.py` now: every index claim is asserted present in the *generated* report, plus the three retracted claims asserted absent.

Gates on this SHA: 4850 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on both drivers, render idempotent.

fxd24:
@codex review

Round-14 fix pushed as `22291bf`. The single P2 was correct and is fixed.

**The dirty-tree scope missed root-level tracked files.** `start()` scoped `_dirty()` to `TRACKED_TREES`, which are directories, so `pyproject.toml` — tracked individually at the repo root as of the previous commit — fell outside it: an uncommitted dependency edit present at `--start` and untouched during the sweep passed the endpoint comparison *and* left `dirty_at_start` empty.

`DIRTY_SCOPES` is now **derived** from `TRACKED` rather than kept as a second hand-maintained list, so the next root-level entry is covered the day it is added. A parallel list that drifts is the exact failure mode this study keeps hitting.

`tmp/probe_dirty_scope.py` proves it can fail: it really modifies `pyproject.toml`, runs both revisions' `start()`, and restores the original bytes it read first (never `git checkout`). The previous commit records only the `write_provenance.py` edit; this one records both. Plus a test asserting every `TRACKED` path falls inside some `DIRTY_SCOPES` entry, so the invariant holds for future additions.

Gates on this SHA: 4853 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, render idempotent.

fxd24:
@codex review

Round-15 fix pushed as `d4d052b`. The P1 was correct, and it is this repo's own rule — uncommitted changes are sacred, and `start()`'s entire job is to overwrite a tracked file.

`start()` now refuses when the sidecar has uncommitted changes, naming them, with `--force` as the deliberate escape so the refusal is a stop rather than a dead end. An **untracked** sidecar counts as pending too: a first-ever run is trivially forced, while a killed run's brand-new sidecar is exactly the file worth protecting. `run_lfm25.sh` now prints both recovery routes (commit it / copy it out, or `--force`) instead of a bare `exit 1`.

`tmp/probe_start_preserves.sh` proves it can fail — against the **real tracked sidecar path**, so the real `git status` check is the one under test: the previous commit destroys the marker, this one refuses with the bytes intact and says why. It restores the original bytes it read up front rather than `git checkout`.

Gates on this SHA: 4857 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on both drivers.

fxd24:
@codex review

Round-16 fixes pushed as `7844d90`. All five findings addressed; the two driver guards are proven by `tmp/probe_round16.sh` against the real artifacts (a dirty rows file stops the run before any `uv` call; `LFM25_FORCE=1` continues through and forwards `--force`).

**P1 — a new sweep overwrote dirty study artifacts.** Correct, and the last round only covered the sidecar. A granular sweep killed after a benchmark subprocess wrote rows but before the model-level commit leaves the rows, report and reference sidecar dirty while the provenance file is clean — it was committed at startup. `run_lfm25.sh` now refuses on any uncommitted artifact of the selected studies, before `--start`, with the same escape.

**P1 — `metric_revision` is not cohort identity.** Also correct. No new measurement was needed to catch it: every row already carries `n_records` and `n_gold_pairs`, which are properties of the dataset rather than the model, so `_assert_same_populations` compares them per benchmark across the two studies. The committed rows pass; a 2173→2100 record change is refused.

**P2 — the `--force` escape was a dead end.** `LFM25_FORCE=1` now carries the decision into the run that acts on it, rather than requiring an invented commit step.

**P2 — `_uncommitted` read git's silence as safety.** Fixed: a file whose clean tracked state cannot be *established* is pending. Only non-existent is safe. That closes both the outside-the-repo case (which is the recovery copy the refusal message itself recommends) and the gitignored case.

**P2 — the load probe drifted from the rows.** The driver now refreshes and commits the probe *before* rendering, inside the same measurement window, and the reproduce block says so instead of listing it after the render. Non-fatal on failure — killing a completed sweep over its explanatory section trades a large loss for a small one — but the report now cross-checks the probe's recorded transformers version against the installed one and discloses a mismatch, so a probe that could not be refreshed is never passed off as current. Currently both are 4.57.6, so no warning renders; the mismatch branch is covered by test.

Gates on this SHA: 4866 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on both drivers, render idempotent.

fxd24:
@codex review

Round-17 fixes pushed as `4596068`. All five findings addressed, proven by `tmp/probe_round17.sh` running both revisions inside a throwaway clone.

**P1 — the guard sat one caller upstream of the thing it guards.** You are right: every generated study document advertises `run_ladder.sh` directly, and that is the script that merges, renders, stages and commits. The refusal moved into `run_ladder.sh` (`LADDER_FORCE=1` overrides), covering every entry point including the published ones. `run_lfm25.sh` keeps its own copy because it must refuse *before* `--start` opens the window, and forwards the force decision.

**P1 — a pending deletion was skipped.** Dropped the `-e` test entirely; git already distinguishes never-created (silence) from deleted (` D`/`D `).

**P1 — `setdefault` kept only the first population.** Now a set per benchmark, refused both *within* each study and across the two. The residual limit is stated rather than implied: `n_records`/`n_gold_pairs` are cardinalities, so a same-sized replacement still passes, and closing that needs a dataset fingerprint the rows do not carry (schema change + re-measure).

**P2 — freshness inferred from version equality.** The probe now records `captured_at`, and the report compares it against **both** ends of the measurement window. The re-rendered report says this about itself right now: the probe I refreshed by hand postdates the window, and it is disclosed.

**P2 — the failure-row render's exit status was ignored.** Now checked. The rows are still committed — they are the durable record of a real failure — but the stale report is excluded from that commit and the sweep stops (exit 4).

One note worth recording: the first version of my probe ran the **unguarded** revision against the **real** artifacts and did precisely what your P1 describes — `record_process_failure` deleted 40 measured rows and the driver committed the deletion. Twice. Both recovered by hand from the previous commit. It now runs in a clone with its origin removed.

Gates on this SHA: 4870 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on both drivers, render idempotent, README-vs-artifact claims re-verified.

fxd24:
@codex review

Round-18 fixes pushed as `fa80a51`. All five P2s were warranted; each fix is proven able to fail by running the same input through HEAD's revision and the tree's (`tmp/probe_round18.py`, `tmp/probe_round18b.py`).

**Licence — §5(c).** Correct, and worse than a missing quote: the patterns were `^\(a\) `/`^\(b\) `, the shape of *every* lettered sub-clause in the file, so the block also carried §4's redistribution conditions. Now extracted by section — §5 entire, plus exactly the definitions §5's own text uses, so the definition list cannot drift out of step with the clause. The prose is scoped to Commercial Use and states the exemption; the conclusion is unchanged and now rests on the right reason — a default is the path taken by users who never read the section, so a carve-out that requires reading it does not rescue one.

**Significance limited to the paired baseline.** The interval is against study B's baseline; the gap runs to the best study-A model. That is *every* row here, not just `walmart_amazon` — `intfloat/e5-base-v2` is the best tuned model on all four. So rather than compute four missing intervals across two studies, the report now says what it has: the CI column names its own other endpoint, each unpaired row carries `‡`, and the narrow verdict names `LiquidAI/LFM2.5-Embedding-350M` instead of "the tuned models".

**Deleted input.** Reproduced exactly as described — HEAD's `--finish` raises `CalledProcessError` and leaves the window open. A missing file now hashes to `"absent"`, so the deletion lands in `changed_during_run` and the sidecar closes *before* the refusal. Still fatal, just recorded first. Non-zero for a file that does exist stays fatal, so this is not a fail-open. Your finding also surfaced a second abort on the same line: `after["blobs"][p]` assumed both snapshots carry identical key sets, which holds only while `TRACKED` itself does — defaulted, keyed on what the start snapshot actually observed.

**CLI artifact paths.** Refused rather than rendered, because option (a) could only have fixed one of the three printed commands: `run_ladder.sh` takes a single `LADDER_ARTIFACT`, so divergent paths are unrepresentable in the other two. Verified: HEAD renders a report whose reproduce block cites `.../scratch_study_rows.jsonl`, a file that does not exist and is not the one it read.

**Timestamps as instants.** Right, and it fails in both directions. Against this study's real window (`01:18:12+02:00` … `03:44:51+02:00`): `2026-07-29T02:00:00+00:00` is 04:00 local — after the close — and read as comfortably inside; `00:30:00+00:00` is 02:30 local, inside, and read as predating. Now parsed as aware datetimes; an offset-less or unparseable stamp returns `"unanchored"` and is disclosed as *unanswerable* rather than crashing the render or passing silently.

Gates on this SHA: 4891 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on both drivers, render idempotent (`05bbfe9e0af4` twice), README-vs-artifact claims re-verified.

fxd24:
@codex review

Round-19 fixes pushed as `8d97f16`. All eight warranted. The two shell guards are proven able to fail by `tmp/probe_round19.sh` (both revisions in a throwaway clone, **plus a control** that a tracked-and-clean artifact is still allowed through — a guard that refuses everything is not a fix); the rest are pinned by new tests.

**P1 — the preflight guarded the wrong set.** Exactly right, and the distinction is the useful part: the loop covered what the sweep *rewrites*, while the failure is about what the write-up *reads*. `LFM25_STUDY=a` still ends by rendering from both rows files and committing the result. Now both studies unconditionally; `PROVENANCE_STUDIES` stays narrow because "which studies did this window re-measure" is a genuinely different question. Verified: HEAD opens the window and starts, never naming study B's dirty rows; the tree refuses before any `uv` call.

**P1 — git's silence is three answers.** `tmp/probe_study` reproduces it: HEAD ran the sweep over a gitignored artifact whose contents exist in no commit. Same three-way distinction `_uncommitted()` already makes.

**P1 — a vanished sidecar.** `--required`, exported by the LFM driver only, so the portfolio ladder keeps its no-op. Your finding also caught a second thing on that line: the `after["blobs"][p]` KeyError I fixed in `finish()` last round exists **twice**, and I patched one copy. Now defaulted in both, with a test that fails on the unfixed version. That is the third time this PR that a fix landed on one of two identical sites.

**P2 — pending deletion in `_uncommitted`.** The same existence-test blind spot I removed from the shell artifact guards one round earlier, still sitting in the Python one. Git is asked first now; a never-created path stays free.

**P2 — aliases and inflections.** Confirmed both: `"You" (or "Your") shall mean` never matched the anchored pattern, and `"Derivative Works"` failed the substring test against §5's singular *Derivative Work*. The extractor was claiming completeness it did not have — worse than quoting less. Every quoted name before *shall mean* now counts as a name for that definition, and a plural definition matches its singular use. The block went 9 → 13 lines.

**P2 — the retracted reading, reintroduced.** In the same generated document, two paragraphs after its own retraction. Scoped to the observed gap.

**P2 — direction of the interval.** Correct: `separates` accepts either direction and the prose asserted one. Rather than gate on `high < 0` and silently drop the other case into no bucket, the direction is now measured per benchmark and printed beside it — so an above-baseline control gets described, not filtered.

**P2 — quoting.** Fixed with `shlex.quote`, which is the identity for every path in the committed studies, so no artifact changed.

Gates on this SHA: 4915 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on both drivers, render idempotent (`57f237ed2a2d` twice), README-vs-artifact claims re-verified.

fxd24:
@codex review

Round-20 fixes pushed as `cb89709`. All eight warranted.

**#239 correction leaking into every artifact.** The worst of the eight, because it is a *fabricated attribution*: study B's report told readers that a merged PR had described `LFM2.5-Encoder-350M`'s `-0.0536` as query-only. #239 is about EmbeddingGemma and predates this study. Both history blocks now emit only for the document they are about, gated on a predicate derived from the module defaults rather than a typed date. The 2026-07-27 report is byte-identical after the change; the two study reports lost 3 lines each.

**P1 — the resume script.** Right on all three counts, and the third is the expensive one: it left an expensive re-measured cell uncommitted, in the exact situation where the rows are most likely to already hold uncommitted work. Routed through `run_ladder.sh` with the benchmark narrowing moved to `LADDER_BENCHMARKS`. **The probe's own control caught a second bug while proving this**: with the tree's resume script beside HEAD's driver, the override reached a driver that had never heard of it and the sweep silently ran all five benchmarks — a green result from the revision under test. Two clones now, and the control stayed.

**P1 — the driver's own outputs.** Correct, and the same shape as your last round's finding one level in: I widened the loop to cover what the renderer *reads* and still missed what the driver *writes*. Both paths declared once at the top; the preflight is built from those variables so the two lists cannot disagree.

**P1 — a closed window under `required`.** The escape hatch and the guard were the same branch. Refused now, naming the finish timestamp.

**P2 — missing interval ≠ missing arm.** Confirmed against `merge_rows()`: it clears a retained challenger's `vs_reference_*` when the reference is re-measured and keeps the reference's rows, so an ordinary cell landed there and got a fabricated reason. The baseline's own arms are consulted now and the two causes read differently.

**P2 — inverted direction.** Same defect I fixed for `narrow` last round, one bucket over. Two facts, stated separately.

**P2 — hard-coded probe values.** In the one file whose stated premise is that no measured quantity is typed into prose. Interpolated, with a test that moves the sidecar and checks the paragraph follows.

**P2 — README.** Third time this PR that a fix landed on one of two sites; that pattern is now the thing I check for before replying, not after.

Gates on this SHA: 4935 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` strict, `bash -n` on all three drivers, render idempotent (`b4287b105f50` twice), README-vs-artifact claims re-verified.

fxd24:
@codex review

Round-21 fixes pushed at 2e5279f (1 x P1, 4 x P2).

The P1 shell guard is proven able to fail: `tmp/probe_drivers.sh` runs each finding's input through **the last revision that lacked its guard** and through the working tree, with two controls. That per-finding "before" revision matters here — the round-20 guards are now committed, so an arm pointed at HEAD would have shown no defect and silently stopped testing anything. Result: every arm separates, both controls pass (a clean resume still runs, narrowed to `walmart_amazon`, and closes + commits its window; a clean `run_lfm25.sh` still starts).

The other four are pinned by new test classes: `TestAMissingControlIntervalIsNotAFinding`, `TestTheCorrectionQuotesOnlyWhatItClaims`, `TestTheOneEnvironmentClaimIsProspective`, `TestDataIsCodeInTheDigest`, `TestTheResumeOwnsAProvenanceWindow`.

Gate: 484 passed in `tests/examples/`, `ruff format --check` + `ruff check` + `mypy src` all clean, `bash -n` clean on all three drivers.

fxd24:
@codex review

Round-22 fixes pushed at 4941ebf. All seven warranted (3 x P1, 4 x P2) — and **two of them had a second site the review named only one of**, which I fixed as well:

- The instruction-axis caveat was flagged for the blockquote. The paragraph *immediately above it* had the identical defect: it explained the base-encoder study's `instruct` column using `google/embeddinggemma-300m` and the Qwen3-Embedding family, neither of which has a row there. Both the blockquote's lists and those worked examples are now derived from the ladder being rendered (via `documented_arm`, the flag that already means "ships a query-side instruction") rather than typed.
- The load-probe finding is itself the sixth instance on this branch of a guard landing on one of two writers. I put the check on the writer and reused `write_provenance._uncommitted` instead of writing a second one — and moved it *before* the ~2 minutes of probing, since a guard that refuses after the work is done just trains operators to reach for the force flag.

On the licence one: an unrecognised licence id now gets **no invented specifics**, only a pointer to the text. That direction is deliberate — this table can be silent about a licence it does not know, but it must never be wrong about one.

Gate: 502 passed in `tests/examples/`; `ruff format --check`, `ruff check`, `mypy src` clean; `bash -n` clean on all three drivers; all four generated documents re-render byte-identically twice in a row, and the README claim-checker (including its own self-test that a false claim would be caught) passes.

fxd24:
@codex review

Round-23 fixes pushed at 02f9843. All five warranted (2 x P1, 3 x P2).

**The `documented_arm` finding is the important one, and it was my own regression from round 22.** Round 22 replaced a hand-typed model list with a derived one — and the derived version made a *false* claim where the typed one had merely made an irrelevant one. That field records whether **this harness configured** a checkpoint-owned arm; the sentence read it as "was not trained with a query-side instruction". So the portfolio ladder asserted that `intfloat/e5-base-v2` and both `Qwen/Qwen3-Embedding-*` rows were never trained with an instruction — while the paragraph two above described Qwen's documented instruction in detail. One generated document contradicting itself.

I checked whether `registered_prompts` could carry the claim instead: it cannot, e5 registers none. Nothing in this harness records training history, so **the claim is dropped rather than re-derived**. The lesson I'd draw: deriving from an artifact is only safer than typing when the field carries the meaning the sentence claims — otherwise it upgrades an obvious error into a confident one.

Two more notes:

- The `lfm25_report.py` writer guard is the **third** site of the round-22 guard, and my round-22 fix is what made it reachable (teaching the resume to re-render pointed a new caller at an unprotected writer). It has a **known false positive** I did not paper over: re-rendering twice before committing makes the second run see the first run's own output as uncommitted work. The guard cannot tell that from a hand edit, so it refuses and says exactly that in both branches of the message.
- Unflagged but fixed: the two new controls asserted against the real artifacts, which are legitimately dirty mid-render, so they would have failed for a reason unrelated to the guard being correct.

Gate: 509 passed in `tests/examples/`; `ruff format --check`, `ruff check`, `mypy src` clean; `bash -n` clean on all three drivers; all four generated documents re-render byte-identically; the README claim-checker passes including its self-test; and the new writer guard was verified to actually fire on the real path, not just in tests.

fxd24:
@codex review

Round-24 fixes pushed at c669185. All six warranted (4 x P1, 2 x P2).

**Four of the six landed on the study-A resume, and the cause was ORDERING, not six oversights** — so I restructured the script rather than patching it six times. Irreversible steps now run last: ask the driver's preflight → open + commit the window → measure → close it *keeping its status* → render only on success → commit the window plus the report *only if this run wrote it* → publish only if everything succeeded.

The preflight is a new `LADDER_PREFLIGHT_ONLY` mode on `run_ladder.sh` — the driver's **own** check, not a copy. A second implementation of one safety rule is the drift this PR keeps paying for.

On the arm-claim P1: this is round 23's mistake one level up. Round 23 was *deriving from the wrong field*; this was *deriving from the wrong source* — the configured ladder instead of the measured rows. Same shape, and worth naming as such.

**Two things you did not flag, both found by re-running `tmp/probe_drivers.sh` against the fix:**

1. Its R21 assertion encoded the OLD dirty-tree behaviour, so it failed on the corrected code. Re-pointed — and the window-opening assertion moved to the clean control, where it belongs.
2. Its "both ends committed" check grepped the commit **subject** for "provenance window". Renaming the closing commit this round turned that check off silently while it kept printing OK for the opening one — **a gate decoupled from what it checks, inside the probe written to catch exactly that.** It counts commits touching the sidecar now: match the artifact, not prose about it.

Also: one pre-existing test pinned the exact "aborted before finishing" wording, so it *agreed with* the defect you flagged. It now asserts the caveat fires without asserting how it explains itself.

Gate: 520 passed in `tests/examples/`; `ruff format --check`, `ruff check`, `mypy src` clean; `bash -n` clean on all three drivers; `tmp/probe_drivers.sh` reports DRIVER GUARDS PROVEN with both controls passing; all four generated documents re-render byte-identically; README claim-checker passes with its self-test.

fxd24:
@codex review

Round-25 fixes pushed at 0053777. All three warranted — and all three are refinements of **round 24's own fixes**, each a guard I placed at the caller I happened to be looking at rather than at the thing being guarded.

- The child push (P1): `run_ladder.sh` publishes each row as it commits it, so "publish only on success" only ever covered the wrapper's push. New `LADDER_NO_PUSH` lets a wrapper that cannot know the outcome until the driver returns own publication. Commits are unchanged — only publication moves.
- The input check (P2): **moved out of `run_lfm25.sh` and into `lfm25_report.py`**, and the shell copy deleted rather than kept in sync. The renderer is the only thing that knows every file it reads, so it is the only place the check can be complete — and this now covers the documented standalone invocation too.
- The arm lists (P2): keyed on successful `(model, prompt_arm)` pairs, since `measure_one` records failures per arm.

**Three rounds, three versions of one sentence, and I think the pattern is worth stating plainly:** it was wrong because it read a field that did not carry the claim (r23), then because it read the configured ladder instead of the rows (r24), then because it read rows at the wrong granularity (r25). Deriving from an artifact is not automatically safer than typing it — it is safer only when the thing derived from means what the sentence says.

One note on the three round-24 tests that asserted this check lived in the shell: I **replaced** them rather than deleting them. The invariant they protected — a refused render must not lose its provenance window — still holds, but now via the driver's generic render-failure branch, so it is asserted there instead. Moving a guard silently changed which code path carries that guarantee.

Gate: 527 passed in `tests/examples/`; `ruff format --check`, `ruff check`, `mypy src` clean; `bash -n` clean on all three drivers; `tmp/probe_drivers.sh` reports DRIVER GUARDS PROVEN with both controls passing; all four generated documents re-render byte-identically; README claim-checker passes with its self-test.

fxd24:
@codex review

Round-26 fixes pushed at 1bc7305. Both warranted — and both are the same shape as the last two rounds (a rule enforced at *some* of the places that need it), so **this round removes the duplication rather than adding the missing copy**.

**The push (P1).** There were **four** implementations of "push this branch, never onto the default one": `run_ladder.sh`'s `push_results`, `run_lfm25.sh`'s final publish, the resume's `publish()`, and — the finding — *none at all* on the abort path. They had already drifted in what they printed and, twice in this series, in whether they ran. They are now one sourced function, `examples/research/publish_lib.sh`, which every driver calls and none reimplements. It never exits: publication is not durability, and a caller whose commits already landed must be free to carry on.

**The inputs (P2).** The guard I added last round enumerated the files I had in mind and skipped the licence text and the prior ladder — both quoted into the output. Every file the module reads is now declared once in `RENDER_INPUTS`, beside the constants, **with a test that fires when a new input is added without being declared**. Enumerating from memory is how the omission happened, so the fix is a single obvious place plus a gate, not remembering better.

Four tests from rounds 24–25 went stale against this refactor — all mine, all **updated rather than deleted**: two asserted the push rule was inline in the resume (they now assert it *uses* the shared one and does **not** re-derive the branch), two patched path constants the guard no longer reads.

Gate: 533 passed in `tests/examples/`; `ruff format --check`, `ruff check`, `mypy src` clean; `bash -n` clean on all three drivers **and the new lib**; `tmp/probe_drivers.sh` reports DRIVER GUARDS PROVEN with both controls passing; all four generated documents re-render byte-identically; README claim-checker passes with its self-test.

fxd24:
Round-27 review addressed at `b34b40c` (2 findings, both warranted, both defects in **round 26's own fixes**).

**P1 — `run_lfm25.sh:143`, the abort push was unconditional.** Correct, and it is my previous fix reversed. Round 26 gave the abort path a push so a closed window could not sit on the local branch while `run_ladder.sh` had already published the rows. But `--finish` exits non-zero exactly when measurement code moved mid-sweep — the same condition that makes `run_ladder.sh` withhold the affected row commits — so `git push HEAD` republished every commit the child had deliberately held back. Publication is now gated on `PROVENANCE_OK`, set to `0` at **both** `--finish` failure sites (the abort helper and the final close). Committing is never gated: withholding the commit would destroy the record of the rejection itself.

**P2 — `lfm25_report.py:50`, `RENDER_INPUTS` omitted the sidecar.** Also correct. `_provenance_section`, `_window_bounds` and `_cross_study_caveat` all quote it, so a standalone render could publish window text derived from bytes no commit holds. It is guarded now, with the explicit allowance you suggested (`LFM25_PROVENANCE_PENDING=1`) claimed by the two drivers that write it moments before rendering and commit both halves in one commit.

The part worth naming: **the test I wrote beside that list asserted the exemption.** `test_the_outputs_are_not_treated_as_inputs` ended with `assert Path(report.PROVENANCE).name not in declared` — a check and a test agreeing with each other about something neither had verified. That is the same shape as the `--porcelain`-blind-to-untracked and the 5-second-review findings this repo already has on record, and it is the fourth time it has appeared in this PR. Rewritten to assert the guard **fires** for a standalone render, that the allowance exempts **only** the sidecar (a dirty rows file still refuses), and that both drivers claim it.

**Both guards proven able to fail** — `tmp/probe_drivers.sh`, arms R27-1 and R27-2, running the same input through `1bc7305` and through the tree:

```
R27-1 -- the abort path PUBLISHED what provenance had rejected  (before: 1bc7305)
  before: rc=4 harness=6 prov=3 | tree: rc=4 harness=6 prov=2
  OK   both arms reached the same rejected-abort site with rc=4
  OK   the before-state ran the push after --finish REJECTED the sweep
  OK   the tree withholds publication and says why
  OK   ...while still COMMITTING the closed window (durability is never gated)

R27-2 -- the renderer quoted an UNCOMMITTED provenance sidecar  (before: 1bc7305)
  OK   it rendered, quoting window text derived from bytes no commit holds
  OK   the tree refuses, naming the sidecar
  OK   a driver that owns both halves of the commit still renders
  OK   and it exempts ONLY the sidecar -- dirty rows still refuse
```

R27-1 needed one correction before it meant anything: `MEM_FLOOR_MIB` is read from the live machine, and on the first run it sent the before-arm to a memory abort (rc 9) while the tree reached the final close (rc 4). The probe would have compared two code paths and called it a difference between two revisions. The arms now assert they landed at the same site with the same exit code *before* comparing publication behaviour.

Gate: 5007 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` clean, `bash -n` on all four shell files. All four generated documents re-render byte-identically, and `tmp/check_readme_claims.py` verifies every claim in the README section against the committed artifact.

@codex review

fxd24:
Round-28 review addressed at `69154cf` (4 findings: 2×P1, 2×P2). All four warranted — each verified against the code and the committed rows before being accepted. **None changes a published number:** all four generated documents re-render byte-identically.

### P1 ×2 — `run_ladder.sh` published what it had just withheld

Both correct, and both the same root cause: `git push origin HEAD` sends the whole **branch**, so withholding a single push withholds nothing.

- **`:211`** — `--verify` fails for one model, its row is committed and not pushed; the tracked `.py` is restored; the next model verifies clean and its push republishes the withheld commit. The endpoint-only `--finish` compares start against end, so it passes too — nothing downstream can see that a row was measured under code the window does not describe.
- **`:773`** — the re-render fails, so the commit deliberately excludes the stale report, and then pushes anyway. The remote ends up holding new rows beside a report generated from the previous ones: exactly the contradiction the exclusion exists to prevent.

**My first fix was wrong in this PR's characteristic way.** I latched a shell variable in `run_ladder.sh` — which latches one *process*. Both wrappers call this driver and then publish for themselves, so a verdict the child reached could never reach them: the same finding one level up, queued for round 29. The condition is a property of the **branch**, so it is now recorded as one — `block_publication` writes a gitignored marker, and `publish_branch` (the single function all three drivers publish through) refuses while it exists. No exit-code plumbing through two levels. It does not self-clear: the withheld commits are still on the branch when the sweep ends, so it fails closed and prints how to clear it.

### P2 — a tie between prompt arms could decide a scientific verdict

Correct, and the data is worse than "possible": **5 of 20 study-A cells and 3 of 20 study-B cells are tied**, including the `random-init-control-350M` control on `walmart_amazon`, whose paired interval decides a published verdict. Its two tied arms carry *different* intervals — `[-0.0275, -0.0072]` and `[-0.0295, -0.0099]`.

A deterministic tie-break alone would not have been enough: deterministic-but-arbitrary is still arbitrary, and a claim that changes with a tie-break rule is not a claim about the benchmark. **Separation now requires every tied-best control arm to clear zero.** Both tied cells agree today and no verdict moves — which is the point: the dependency goes while it is still free, not after a re-measure walks one interval across zero.

While doing it I found I was about to write the fourth copy of `low > 0 or high < 0`; a bound landing *exactly* on 0.0000 is a measured case here (`fodors_zagat`), so which copy a reader hits must not decide it. It is one predicate now.

### P2 — the instruction claim and reproduce command used the wrong cohort

Correct. Both iterated `LADDER` while the document's coverage denominator is `_ladder_specs(rows)`, which row-expands for checkpoints measured via `--models`. A row-backed custom checkpoint appeared in the tables while the prose said no model's own recipe was measured, and the printed `LADDER_ALL_MODELS` would reproduce the document against a smaller denominator than it reports. `_ladder_specs` exists precisely so these cannot disagree — these were its third and fourth expressions.

### Proving the guards can fail

`tmp/probe_drivers.sh` gains R28-1/R28-2, running `33eedc2` and the tree through the same stubbed sweep:

```
R28-1 -- one model's withheld push was republished by the next  (before: 33eedc2)
  before: rc=0 push-attempts=1 | tree: rc=0 push-attempts=0
  OK   both arms had verification fail for exactly one model
  OK   the before-state pushed AFTER the withheld model -- republishing it
  OK   the tree latches: no push for any later model
  OK   ...while still COMMITTING every model's rows

R28-2 -- rows were PUBLISHED without the report they contradict  (before: 33eedc2)
  before: rc=4 push-attempts=1 | tree: rc=4 push-attempts=0
  OK   the before-state published rows whose report it had just excluded
  OK   the tree commits the rows and withholds the push
```

The ladder cohort tests were run against a mutated copy of the module; both fail on the before-state (one of them only after I fixed it — slicing "everything after the reproduce heading" was reading the rest of the report, not the command).

Two of my own probes were wrong first, and both are the same lesson as the findings: the push counter scored a **refusal** as a push attempt, because the refusal line is also a `[ladder]` line; and the marker leaked between arms precisely because it deliberately does not self-clear.

**Boundary, stated rather than implied:** the dynamic probe covers `run_ladder.sh`'s own pushes. The wrapper case rests on all three drivers publishing only through `publish_branch` — asserted by a test, not exercised end to end.

Gate: 5026 passed / 2 skipped, `ruff format --check`, `ruff check`, `mypy src` clean, `bash -n` on all four shell files, `tmp/check_readme_claims.py` green.

@codex review

fxd24:
**Review on this PR is closed at `05e649a`. No further review round is requested.**

This PR reached 100 findings across 22 rounds on a +173k-line diff. That is a runaway loop, not thoroughness: against a reviewer that always finds something, "fix everything and re-request" has no exit condition. Sourcery declining twice for exceeding its 150k diff-char limit was the signal the PR was too big, and it should have been read as one.

**The final bar applied** — two categories only: a statement the write-up makes that the data does not support, and data loss or a silently wrong result.

Both named open false claims were checked against the committed artifact and are already fixed:

- **Headline qualification by prompt arm** — the headline names the specific losing arms (`abt_buy` (instruct), `abt_buy` (none), …), states "these are per-ARM verdicts, not a checkpoint-level one", and lists the untested `documented` arm explicitly, including `abt_buy` **0.9741** vs the baseline's best **0.9703** — above the baseline and untestable, because the baseline has no `documented` arm to pair against.
- **`lfm1.0` under the OSI heading** — the OSI section now says the study's reference is *not* one of the candidates, names `intfloat/e5-base-v2` as the actual `DEFAULT_EMBEDDING_MODEL`, and the "former default" claim is gone. The licence summary also carries §5(c)'s non-profit exemption, so the "everyone above $10M is outside the licence" over-claim is gone too.

No remaining defect can silently misattribute measured rows. The one that could — rows measured under changed code being published under provenance for restored code — is fixed at `69154cf`/`05e649a`: publication is withheld on a branch-scoped marker that `publish_branch` honours, so a later process cannot republish what an earlier one refused. Committing is never gated; only publication is.

**Everything else is deferred to #256** with thread links — publication ordering of the combined write-up, the fail-open marker write, the provenance digest picking up unrelated harnesses' outputs, and tied-best control arms pointing opposite ways. Two of those had fixes already drafted; rather than grow this diff further, the patch is preserved in full in that issue. All four are latent: none changes a published number, and all four documents re-render byte-identically.

### One correction worth stating plainly

An earlier revision of this document said **two of five** benchmarks cannot separate a trained retriever from random weights. **That was measured to be wrong and was retracted.** The control's paired intervals on `walmart_amazon` are `[-0.0275, -0.0072]` and `[-0.0295, -0.0099]` — both exclude zero, so it *does* separate. The supported claim is **one of five**: only `fodors_zagat` fails, with both bounds sitting exactly on 0.0000, which does not exclude zero.

What is true about `walmart_amazon` is the related but weaker statement, and it is in the write-up: the entire trained-vs-random range there is **0.0183**, so a reported margin below roughly that size is a large fraction of the benchmark's whole dynamic range. That is **resolution, not significance** — such a ranking is fragile, not meaningless.

### Gate at `05e649a`

5027 passed / 2 skipped · `ruff format --check`, `ruff check`, `mypy src` clean · all four generated documents re-render byte-identically · `check_readme_claims` verifies every claim in the summary against the committed rows.

### The structural lesson

This PR became unreviewable because harness code, generated data artifacts, and the write-up went into one change. Those are three PRs: the harness reviewed as code, the rows reviewed for provenance and reproducibility, and the write-up reviewed for claims against the data. Recorded in #256.

